PR #30, opened 2025-03-05
API: add `removeMentionsText` util and tests

aacebo:
add utility to remove mentions text from an activity, also add any missing tests

### DIFF
--- a/packages/api/src/activities/utils/remove-mentions-text.ts
+++ b/packages/api/src/activities/utils/remove-mentions-text.ts
@@ -16,6 +16,8 @@ export interface RemoveMentionsTextOptions {
   /**
    * when `true`, the inner text of the tag
    * will not be removed
+   * Eg. input: Hello <at>my-bot</at>! How are you?
+   *     output: Hello my-bot! How are you?
    */
   tagOnly?: boolean;
 }
--- a/packages/api/src/activities/utils/remove-mentions-text.ts
+++ b/packages/api/src/activities/utils/remove-mentions-text.ts
@@ -1,0 +1,51 @@
+import { MessageSendActivity, MessageUpdateActivity } from '../message';
+import { TypingActivity } from '../typing';
+
+/**
+ * any activity type that has a `text` property
+ */
+type TextActivity = MessageSendActivity | MessageUpdateActivity | TypingActivity;
+
+export interface RemoveMentionsTextOptions {
+  /**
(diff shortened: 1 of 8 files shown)


packages/api/src/activities/utils/remove-mentions-text.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export interface RemoveMentionsTextOptions {
+  /**
+   * the account to remove mentions for
+   */
+  accountId?: string;
+
+  /**
+   * when `true`, the inner text of the tag
+   * will not be removed
```
heyitsaamir, line 17:
```suggestion
   * when `true`, the inner text of the tag
   * will not be removed
   * Eg. input: Hello <at>my-bot</at>! How are you?
   *     output: Hello my-bot! How are you?
```

packages/api/src/activities/utils/remove-mentions-text.ts
```diff
@@ -0,0 +1,51 @@
...
+import { TypingActivity } from '../typing';
+
+/**
+ * any activity type that has a `text` property
+ */
+type TextActivity = MessageSendActivity | MessageUpdateActivity | TypingActivity;
+
+export interface RemoveMentionsTextOptions {
+  /**
+   * the account to remove mentions for
```
heyitsaamir, line 11:
```suggestion
   * the account to remove mentions for
   * by default, all at-mentions listed in `entities` are removed.
```

packages/api/src/activities/utils/remove-mentions-text.ts
```diff
@@ -0,0 +1,51 @@
...
+   */
+  accountId?: string;
```
heyitsaamir, line 13: Hm, consider using `entity`. Account isn't verbiage we normally use here right? Looks like activity uses `entity`

heyitsaamir (dismissed):
Minor nits

packages/api/src/activities/utils/remove-mentions-text.ts
aacebo: in BF they use the mentioned accounts id, https://github.com/microsoft/botbuilder-dotnet/blob/main/libraries/Microsoft.Bot.Schema/ActivityExtensions.cs#L64, and entities don't typically have a unique identifier.

heyitsaamir (approved):

PR #37, opened 2025-03-07
Common: Remove logger dependency on process object

Jesperholmbergmsft:
This updates the ConsoleLogger class to 
a) take logger name pattern as an optional parameter to the constructor, 
b) take log level from `process.env` when available, and 
c) be resilient to the process instance being unavailable.
For both logger & pattern options, the environment value is favored over the logger options.


Jesperholmbergmsft:
@microsoft-github-policy-service agree company="Microsoft"

aacebo (approved):

PR #38, opened 2025-03-07
Api/Apps: break `App` class and `IActivityContext` up

aacebo:
the `App` class and `IActivityContext` have gotta too large and need to be organized in a more scalable way. To do this I have broken the `App` class into several implementation files prefixed with `app.` and moved certain things to the context implementation.

I also included util function additions to the `Activity` class implementations to ease of use.
image


packages/config/jest.config.js
```diff
@@ -19,12 +19,12 @@ module.exports = {
...
-      functions: 90,
-      lines: 90,
-      statements: 90,
-    },
+    // global: {
+    //   branches: 90,
+    //   functions: 90,
+    //   lines: 90,
+    //   statements: 90,
+    // },
```
corinagum, line 27: Just checking if this is accidental or left commented out on purpose

aacebo: ya, commented out until we can get coverage to a respectable amount.

PR #94, opened 2025-03-16
Add sign in options

heyitsaamir:
This is a fairly minor (but breaking) interface change. 

1. It changes "name" to "connectionName" which is more clear
2. It adds further options to customize the oauth card
3. It moves options to a partial object for `signin` << this is the breaking change.

I found that these changes made it easier for me to understand what was going on. Happy to remove the breaking change if that's preferred.

### DIFF
--- a/packages/apps/src/contexts/activity.ts
+++ b/packages/apps/src/contexts/activity.ts
@@ -219,7 +219,7 @@ export class ActivityContext<T extends Activity = Activity> implements IActivity
         members: [this.activity.from],
       });
 
-      await this.send({ type: 'message', text });
+      await this.send({ type: 'message', text: oauthCardText });
       convo.conversation = { id: res.id } as ConversationAccount;
     }
 
--- a/packages/apps/src/contexts/activity.ts
+++ b/packages/apps/src/contexts/activity.ts
@@ -9,7 +9,9 @@ import {
   MessageDeleteActivity,
   MessageUpdateActivity,
   toActivityParams,
+  TokenExchangeResource,
   TokenExchangeState,
+  TokenPostResource,
   TypingActivity,
 } from '@microsoft/spark.api';
 import { ILogger } from '@microsoft/spark.common/logging';
@@ -73,7 +75,7 @@ export interface IActivityContextOptions<T extends Activity = Activity> {
   next: (context?: IActivityContext) => (void | InvokeResponse) | Promise<void | InvokeResponse>;
 }
 
-interface ISignInOptions {
+type SignInOptions = {
   /**
    * The name of the auth connection to use
    * @default `graph`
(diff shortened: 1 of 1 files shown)


packages/apps/src/contexts/activity.ts
```diff
@@ -73,6 +73,24 @@ export interface IActivityContextOptions<T extends Activity = Activity> {
   next: (context?: IActivityContext) => (void | InvokeResponse) | Promise<void | InvokeResponse>;
 }
 
+interface ISignInOptions {
```
rajan-chari, line 76: Since this interface only has data members, why make it an interface.  For example, does this make testing easier in typescript?

heyitsaamir: typescript doesn't have a concept of kwargs. So if you have many function args, and you don't want to supply some of them, then you have to do foo(undefined, undefined, "arg").

But using interfaces does:
foo({thirdArg: "arg"})

packages/apps/src/contexts/activity.ts
```diff
@@ -73,6 +73,24 @@ export interface IActivityContextOptions<T extends Activity = Activity> {
...
+  connectionName: string;
+  /**
+   * The text to display on the oauth card
+   * @default `Please Sign In...`
+   */
+  oauthCardText: string;
+  /**
+   * The text to display on the sign in button
+   * @default `Sign In`
+   */
```
aacebo, line 90: can we change this to a `type` instead? After the `I` interface rename I have been trying to use types over interfaces when inheritance isn't being used or when a class isn't intended to implement the interface.

packages/apps/src/contexts/activity.ts
```diff
@@ -106,6 +123,12 @@ export interface IActivityContext<T extends Activity = Activity>
   signout: (name?: string) => Promise<void>;
 }
 
+const DEFAULT_SIGNIN_OPTIONS: ISignInOptions = {
+  connectionName: 'graph',
+  oauthCardText: 'Please Sign In...',
+  signInButtonText: 'Sign In',
```
aacebo, line 129: should we also add the ability for them to add other activity props if they want? Right now there is work being done by Leon where they want in BF to be able to trigger the sign in flow but add extra activity data points so exposing that as well would be helpful.

heyitsaamir: Updated. Take a look. The isGroup clause is still curious though. What's the expected behavior there? User goes to their 1:1 chat, then grants SSO there?

heyitsaamir: Should this be the regular card that we're sending to 1:1 anyway?

aacebo: ya, thats the hack GCP team came up with to support group chat oauth flows.

aacebo (approved):

PR #96, opened 2025-03-17
[cli] Update basic tab template to support dynamic redirect Uri

Jesperholmbergmsft:
This updates the basic tab template so that the app registration SPA redirect URI is updated to match the dynamic tunnel name when running locally. This to set up for supporting NAA next.

This switches the yaml templates from using `botAadApp/create` to `aadApp/create` to create the app registration, just to get the object ID variable really - to allow using `aadApp/update` to update the redirect URI. 

How tested:
 - built the cli locally; created a new app using the template; ran that app in Teams using TTK; verified on the Azure app registration that the tunnel is set correct, and that I can chat with the bot.

### DIFF
--- a/packages/cli/configs/ttk/basic/aad.manifest.json.hbs
+++ b/packages/cli/configs/ttk/basic/aad.manifest.json.hbs
@@ -1,0 +1,112 @@
+{
+    "id": "$\{{AAD_APP_OBJECT_ID}}",
+    "appId": "$\{{BOT_ID}}",
+    "displayName": "{{ toPascalCase name }}$\{{APP_NAME_SUFFIX}}",
+    "identifierUris": [
+        "api://$\{{BOT_DOMAIN}}/$\{{BOT_ID}}"
+    ],
+    "signInAudience": "AzureADMultipleOrgs",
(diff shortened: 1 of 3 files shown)


packages/cli/configs/ttk/basic/aad.manifest.json.hbs
```diff
@@ -0,0 +1,112 @@
+{
+    "id": "$\{{AAD_APP_OBJECT_ID}}",
+    "appId": "$\{{BOT_ID}}",
+    "displayName": "{{ toPascalCase name }}$\{{APP_NAME_SUFFIX}}",
+    "identifierUris": [
```
aacebo, line 5: this manifest is present in the `oauth` ttk configuration, is it need in basic? Basic is not intended to have oauth functionality

Jesperholmbergmsft: Thanks for raising this. As discussed offline, 2nd iteration is updated to undo the basic config changes, and instead adds a separate ttk config for tab project.

aacebo (approved):

PR #105, opened 2025-03-20
Fix system prompt sending, and add comment

heyitsaamir:
- System prompts right now were being `push`ed after the messages. We want them to be at the top of the messages array.
- Remove trailing slashes from API urls if they exist.
- Added a comment that makes it more clear the `instructions` are commonly referred to as system prompts.

### DIFF
--- a/packages/openai/src/chat.ts
+++ b/packages/openai/src/chat.ts
@@ -131,7 +131,7 @@ export class OpenAIChatModel implements IChatModel<ChatCompletionCreateParams> {
       const completion = await this._openai.chat.completions.create({
         ...this.options.requestOptions,
         ...options.request,
-        model: this.options.model,
+        model: 'endpoint' in this.options ? '' : this.options.model,
         stream: !!options.onChunk,
         tools:
           Object.keys(options.functions || {}).length === 0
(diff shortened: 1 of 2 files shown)


packages/openai/src/chat.ts
```diff
@@ -124,14 +124,14 @@ export class OpenAIChatModel implements IChatModel<ChatCompletionCreateParams> {
...
-      messages.push(options.system);
+      messages.unshift(options.system);
     }
 
     try {
       const completion = await this._openai.chat.completions.create({
         ...this.options.requestOptions,
         ...options.request,
-        model: 'endpoint' in this.options ? '' : this.options.model,
+        model: this.options.model,
```
heyitsaamir, line 134: @aacebo do you know why we need this? It wasn't playing well with my AOAI configs

aacebo: that needs to be present for AOAI types, ie if 'endpoints' is in the options they are for AOAI, which you shouldn't pass a `model`

heyitsaamir: Removed it

aacebo (approved):

PR #115, opened 2025-03-27
Add AIPlugins and MCP Client package

heyitsaamir:
- Adds an IAiPlugin (crappy name...) for AI plugins in general. This interface contains some basic life-cycle methods for AI Plugins. See https://github.com/microsoft/spark.js/blob/348028026ee1951965dbfd7549509318e6b48875/packages/ai/src/prompts/plugin.ts
- Minor breaking change in `chat.ts`. It was returning `content: string`, but IChatPrompt returns `content?: string`. 
- Adds an MCPClient external package. This package has the ability to connect to various servers and use them as tools with the ai package
- I couldn't use the @modelcontextprotocol/sdk out of the box because I was getting hit with this error:
```
Error: require() cannot be used on an ESM graph with top-level await. Use import() instead. To see where the top-level await comes from, use --experimental-print-required-tla.
```
It seems to be a known error discussed here: https://github.com/modelcontextprotocol/typescript-sdk/issues/217.
Versioning down to 1.5.0 seems to alleviate the issue for now. I put up a PR (https://github.com/modelcontextprotocol/typescript-sdk/pull/254) which has been merged. Just waiting for their release.

Tests:
```
 PASS  src/prompts/chat.spec.ts
  ChatPrompt
    constructor
      ✓ should initialize with default values (1 ms)
      ✓ should initialize with custom values
    function registration
      ✓ should register a function (1 ms)
      ✓ should register a function with parameters (1 ms)
    plugin system
      ✓ should use plugin
      ✓ should call onBeforeSend hook when sending message (1 ms)
      ✓ should allow onBeforeSend to modify the input
      ✓ should call onAfterSend hook after receiving response (1 ms)
      ✓ should allow onAfterSend to modify the response
      ✓ should chain multiple plugins hooks in order
    send
      ✓ should send message and receive response
      ✓ should handle streaming with onChunk callback
      ✓ should call plugin onBuildFunctions when sending message
      ✓ should send message with functions from both prompt and plugins (1 ms)
      ✓ should send message with custom request options
      ✓ should handle array of content parts
      ✓ should send message with custom messages history
      ✓ should handle complex function parameters
    function calling
      ✓ should call registered function
      ✓ should throw error for non-existent function (7 ms)
    prompt chaining
      ✓ should chain prompts using use() (1 ms)
      ✓ should chain prompts with custom name
```

https://github.com/user-attachments/assets/52306053-62cd-4c83-87f7-acefbbfbc3a9


samples/mcpclient/appPackage/manifest.json
```diff
@@ -0,0 +1,55 @@
+{
+  "$schema": "https://developer.microsoft.com/json-schemas/teams/v1.15/MicrosoftTeams.schema.json",
+  "version": "1.0.0",
+  "manifestVersion": "1.15",
```
aacebo, line 4: is this being checked in for use with TTK? Should the other TTK yml/bicep files be accompanying it?

samples/mcpclient/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "https://json.schemastore.org/tsconfig",
```
aacebo, line 2: is there a reason this can't use the common config files like the other samples?

samples/mcpclient/tsup.config.js
```diff
@@ -0,0 +1,13 @@
+/** @type {import('tsup').Options} */
+module.exports = {
+  dts: true,
+  minify: false,
+  bundle: false,
+  sourcemap: true,
+  treeshake: true,
+  splitting: true,
+  clean: true,
```
aacebo, line 9: since this is a node sample that we know won't be used in the browser, I typically don't use tsup. Wondering what your thoughts are on that or if there are some benefits of it I don't know about outside of bundling

external/mcpclient/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "teams",
+    "msteams"
+  ],
+  "scripts": {
+    "clean": "npx rimraf ./dist",
+    "build": "npx tsup",
+    "test": "npx jest"
+  },
+  "peerDependencies": {
+    "@modelcontextprotocol/sdk": "1.5.0",
```
aacebo, line 30: `@microsoft/spark.common` should also be a peer dep (logging and such)

external/mcp/package.json
```diff
@@ -41,7 +41,7 @@
     "@microsoft/spark.apps": "*",
     "@microsoft/spark.common": "*",
     "@microsoft/spark.dev": "*",
-    "@modelcontextprotocol/sdk": "^1.7.0"
+    "@modelcontextprotocol/sdk": "1.5.0"
```
aacebo, line 44: it has always worked for me, lets sync after standup when your back and try to see what the problem could be.

external/mcpclient/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@microsoft/spark.mcpclient",
```
aacebo, line 2: wondering if there is a way we can combine the client/server into one package?

aacebo:
after reading the error you linked I understand the issue more, it seems their server code is CJS compatible but their client is ESM (which makes perfect sense for browser usage), lets chat about the best path forward when your back!

samples/mcpclient/tsconfig.json
heyitsaamir: No reason, just a mistake

external/mcpclient/package.json
heyitsaamir: It's okay to keep it in dev-dep right? (afaik, peer dep + dev dep is okay to be the same?) thoughts?

external/mcpclient/package.json
heyitsaamir: I think it's possible, but it does conflate the purpose of prompts a bit, so it might get a bit confusing maybe?

packages/ai/src/mcp.ts
```diff
@@ -0,0 +1,17 @@
+import { Function } from './function';
```
aacebo, line 1: I see why you need to do this. To me making changes in the core packages for these kinds of things (external integrations that come and go often) is a smell that the AI package isn't flexible enough and needs more plugin/extension capabilities.

What I mean by that is that with teams AI lots of trendy features came and went, and every time we added to the core package we incurred dept because the features rarely have similar schemas. The same applies to the Apps package which is why I implemented the plugins abstraction. I'd like us to find some way to do something similar here, so we can build integrations as lego pieces on top of the core

aacebo: for example, last year all the hype was around the agent protocol https://agentprotocol.ai/

packages/ai/src/prompts/chat.ts
```diff
@@ -60,12 +61,36 @@ export type ChatPromptSendOptions<TOptions extends Record<string, any> = Record<
...
+  /**
+   * Unique name of the plugin
+   */
+  name: TPluginName;
+
+  /**
+   * The function is set to initialize the plugin
+   * @param args - Arguments to initialize the plugin
+   */
+  usePlugin?: (args: TPluginUseArgs) => void;
```
aacebo, line 74: the name `onInit` might be more clear and match that of the apps package plugins

packages/ai/src/prompts/chat.ts
```diff
@@ -60,12 +61,36 @@ export type ChatPromptSendOptions<TOptions extends Record<string, any> = Record<
...
+
 /**
  * a prompt that can interface with a
  * chat model that provides utility like
  * streaming and function calling
  */
-export interface IChatPrompt<TOptions extends Record<string, any> = Record<string, any>> {
+export interface IChatPrompt<
+  TOptions extends Record<string, any> = Record<string, any>,
+  TChatPromptPlugins extends readonly ChatPromptPlugin<string, any>[] = [],
```
aacebo, line 92: Itd be nice to include the plugins array in the options that exist so the input looks a bit cleaner

packages/ai/src/mcp.ts
heyitsaamir: Added AI Plugins

packages/ai/src/prompts/chat.ts
heyitsaamir: added plugin array, lmk if that aligns to what you were thinking

external/mcpclientplugin/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@microsoft/spark.mcpclientplugin",
```
aacebo, line 2: none of the plugin packages have plugin in the name, I suggest `@microsoft/spark.mcpclient` or `@microsoft/spark.mcp.client`, although I'd still like to see if we can merge the other mcp package with this one someone and use named exports so people can just do `import ... from '@microsoft/spark.mcp/{server | client}'`

packages/ai/src/prompts/__tests__/chat.test.ts
```diff
@@ -0,0 +1,316 @@
+import { ContentPart, Message } from '../../message';
+import { IChatModel } from '../../models';
+import { Schema } from '../../schema';
+import { ChatPrompt, ChatPromptPlugin } from '../chat';
+
```
aacebo, line 5: to match the other projects, tests should use `*.spec.ts` names and be in the same folder location as the source code. This is just to keep this consistent, im open to a discussion of using a tests folder instead but that would require we update the existing packages with tests to match that.

samples/mcpclient/appPackage/manifest.json
heyitsaamir: Nope. Removing.

external/mcp/package.json
heyitsaamir: Fixed now that https://github.com/modelcontextprotocol/typescript-sdk/pull/254 is merged and released

external/mcpclient/package.json
heyitsaamir: Talked offline. Let's keep them separate for now. If a server wants to do both, it can compose it using these packages.

Question: Should I rename `mcp` to `mcpServer`?

packages/ai/src/prompts/chat.ts
heyitsaamir: The way to use it would be
```
chatPrompt.usePlugin('mcp', ...).usePlugin('mcp', ...)
```
Multiple `init`s might be kinda confusing?
```
chatPrompt.onInit('mcp', ...).onInit('mcp', ...)
```
Thoughts?

external/mcpclientplugin/package.json
heyitsaamir: Will update to use mcpClient, but let's keep it separate for simplicity's sake for now, unless you feel strongly

external/mcpclientplugin/src/mcpClientPlugin.ts
```diff
@@ -0,0 +1,205 @@
+import type { ChatPromptPlugin, Function, Schema } from '@microsoft/spark.ai';
+import { Client, ClientOptions } from '@modelcontextprotocol/sdk/client/index.js';
+import { SSEClientTransport } from '@modelcontextprotocol/sdk/client/sse.js';
```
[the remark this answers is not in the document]
heyitsaamir: good catch!

packages/ai/src/prompts/__tests__/chat.test.ts
heyitsaamir: ah shoot. This had to be deleted. I already had a `chat.spec.ts`. removed!

external/mcpclient/package.json
aacebo: peer requires them to explicitly have it installed themselves which all the other samples/packages use currently. It prevents duplicate package installations due to version differences. Its also not a dev dep since it has implementations required at runtime.

external/mcpclient/package.json
aacebo: I don't have strong feelings either way, Socketio does it the way we have it here (naming wise) but I could also see it being the other way.

external/mcpclient/turbo.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": ["//"],
+  "tasks": {
+    "build": {
+      "inputs": ["$TURBO_DEFAULT$", ".env*"],
+      "outputs": [".next/**", "!.next/cache/**"],
+      "cache": false,
+      "dependsOn": ["@microsoft/spark.common#build", "@microsoft/spark.ai#build"]
```
aacebo, line 8: since your not depending on spark.common anymore, it can be removed from this dependsOn Array

packages/ai/src/mcp.ts
```diff
@@ -0,0 +1,17 @@
...
+import { Schema } from './schema';
+
+export type McpClientParams = {
```
aacebo: should these MCP specific types be removed from the core AI package now that we have the plugin changes needed to keep the MCP concepts in their own packages?

aacebo: disregard, somehow was looking at an older commit

aacebo (approved):

Issue #66, opened 2025-03-07
[devtools] Render attachments 'in' the ComposeBox


corinagum:
Requires taking out `<Textarea />` component and re-engineering the features

Issue #61, opened 2025-03-07
[devtools] Performance improvements


corinagum:
Improve React performance (re-renderings)

- [x] Zustand - don't use `ContextProvider`
- [ ] `memo` the components
- [ ] `useCallback` functions
- [ ] `useMemo` more
- [ ] minimize `useEffect`s

Issue #40, opened 2025-03-07
[devtools] Message Update


aacebo:
add support for message updates

PR #116, opened 2025-03-28
[devtools] Rework ComposeBox (ContentEditableArea), messageUpdate, and more

corinagum:
fix: #40 
fix: #61 
fix: #66 

image

image
image
image
image

- Created `ContentEditableArea` to replace FluentUI's `TextArea` - visually, they are identical, but allow for a lot of better UX
- Attachments can render in the ComposeBox or EditComposeBox (just cards for now, still...)
- User can update their messages by clicking the edit button (pencil) in the MessageActionsToolbar
- Better usage of React performance hooks
- You can add card attachments to messages you are editing
- if you type in the ComposeBox, then press "Open card designer", then attach a card, your message that you were typing will be preserved. Same for if you made changes to your messages while editing. Focus will return to the ContentEditableArea you were using before sidetracking to cards.
- Moved location of Toast notifications to top-middle of screen - bottom right was blocking the send button
- Improve markdown handling - preserve line breaks
- Added Adaptive Card target widths to hopefully work if card contains targetWidth


packages/devtools/src/components/ComposeBox/ComposeBoxToolbar/ComposeBoxToolbar.tsx
```diff
@@ -1,186 +1,187 @@
...
+    const [isDialogOpen, setIsDialogOpen] = useState(false);
+    const [menuOpen, setMenuOpen] = useState(false);
+    const [isConfirmCancelOpen, setIsConfirmCancelOpen] = useState(false);
+    const { setDraftMessage, setEditingMessageId } = useCardStore();
 
-  const handleNavigateToCards = () => {
-    navigate('/cards');
-    setMenuOpen(false);
-  };
+    const handleCancelDialogOpen = useCallback(() => {
```
heyitsaamir, line 70: NIT usecallback isn't necessary

heyitsaamir (approved):
Kinda hard to review b/c of size. It'd be good to keep PRs atleast somewhat reviewable, tho i understand it's not always possible

aacebo (approved):

PR #117, opened 2025-03-28
[Client] Use MSAL to generate NAA token and pass in authorization header when making function calls

Jesperholmbergmsft:
This PR updates the Client package to take a dependency on MSAL and teams-js. This in order to generate a bearer token to include in function calls, for the server side to validate. In addition, minor adjustments are done to the TTK configuration and tab sample in order to showcase the feature.

Token validation is not included in this PR, that'll come separately. 

How tested:
 - Building the cli, app and client libraries locally, I used the cli to create a new Embed app from the Tab template
 - In the newly created app, after pointing the app & client dependencies to my local build, simply run in Chrome using TTK
 - The app is built, deployed, starts up. When I install the app in Teams and open the tab, I'm prompted to consent to the user.read scope in order to create a token. After consenting, the app successfully calls the `hello-world` function and prints the response.
 - I see in the network tab that a valid bearer token is included in the function call, and the other custom spark request headers are included as expected.
 -

### DIFF
--- a/packages/apps/src/middleware/with-client-auth.ts
+++ b/packages/apps/src/middleware/with-client-auth.ts
@@ -16,22 +16,24 @@ export type ClientAuthRequest = express.Request & {
 export function withClientAuth(params: WithClientAuthParams) {
   const log = params.logger;
   const clientId = params.clientId;
-  const clientSecret = 'clientSecret' in params ? params.clientSecret : undefined;
   const tenantId = 'tenantId' in params ? params.tenantId : undefined;
 
   return (req: ClientAuthRequest, res: express.Response, next: express.NextFunction) => {
     const appClientId = req.header('X-Spark-App-Client-Id');
-    const appClientSecret = req.header('X-Spark-App-Client-Secret');
     const appTenantId = req.header('X-Spark-App-Tenant-Id');
     const appSessionId = req.header('X-Spark-App-Session-Id');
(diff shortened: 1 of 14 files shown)


packages/apps/src/middleware/with-client-auth.ts
```diff
@@ -16,22 +16,24 @@ export type ClientAuthRequest = express.Request & {
...
-  const clientSecret = 'clientSecret' in params ? params.clientSecret : undefined;
   const tenantId = 'tenantId' in params ? params.tenantId : undefined;
 
   return (req: ClientAuthRequest, res: express.Response, next: express.NextFunction) => {
     const appClientId = req.header('X-Spark-App-Client-Id');
-    const appClientSecret = req.header('X-Spark-App-Client-Secret');
     const appTenantId = req.header('X-Spark-App-Tenant-Id');
     const appSessionId = req.header('X-Spark-App-Session-Id');
     const pageId = req.header('X-Spark-Page-Id');
+    const authorization = req.header('Authorization')?.split(' ');
```
Jesperholmbergmsft, line 26: This is where an upcoming PR will actually validate the token. At that point, I'll remove the `X-Spark-User-Id` and `X-Spark-Tenant-Id` headers as we can get those values from the token itself. I just made minor changes here for the moment not to regress existing validation.

aacebo (approved):

PR #118, opened 2025-03-28
Fix explicit Oauth flow (remove support for multiple connections)

heyitsaamir:
- Currently our oauth flow works well for SSO, but not for explicit Oauth.
- In explicit Oauth, the value for the `tokenExchangeUrl` is set to blank
image
- This means that this is NOT an SSO. When this is the case, the `onTokenExchange` will not be called. And this is the function which sets the storage value to indicate that an Oauth is actually happening. The `onVerifyState` checks if this value is present or not before proceeding.
- In this PR, we are (temporarily) removing the ability to have multiple connections in favor of simplicity. This means that we don't need to store connectionName anywhere and we can use this `defaultConnectionName`.
- I am working on getting stateless oauth prioritized, until this I've raised this issue to track it (https://github.com/microsoft/spark.js/issues/125).

### DIFF
--- a/packages/apps/src/app.ts
+++ b/packages/apps/src/app.ts
@@ -1,36 +1,36 @@
-import { AxiosError } from 'axios';
-import { ILogger, ConsoleLogger } from '@microsoft/spark.common/logging';
-import { LocalStorage, IStorage } from '@microsoft/spark.common/storage';
 import { EventEmitter } from '@microsoft/spark.common/events';
 import * as http from '@microsoft/spark.common/http';
+import { ConsoleLogger, ILogger } from '@microsoft/spark.common/logging';
+import { IStorage, LocalStorage } from '@microsoft/spark.common/storage';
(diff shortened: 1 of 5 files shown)


packages/apps/src/app.ts
```diff
@@ -71,6 +71,12 @@ export type AppOptions = Partial<Credentials> & {
    * Activity Options
    */
   readonly activity?: AppActivityOptions;
+
+  /**
+   * The default connection name to use for the app
+   * @default 'graph'
+   */
+  readonly defaultConnectionName?: string;
```
aacebo, line 79: we should update the existing oauth settings type here https://github.com/microsoft/spark.js/blob/main/packages/apps/src/oauth.ts#L1 with a `default` or something, since we will most likely add auth settings as time goes on and this will keep a nice area for them all to live

aacebo (approved):

PR #127, opened 2025-04-07
[WIP] Add cursor rules

heyitsaamir:
WIP

- Considering if "editor rules" is the right approach so we can do other AI editors too.


heyitsaamir:
Closing this in favor of llmtxt that'll be added to docs + adding this as part of config with the CLI

Issue #114, opened 2025-03-27
[devtools] Adjustments


corinagum:
- [ ] Attachments remove button - visible on hover/focus
- [x] Show 'Edited' next to timestamp as applicable
- [x] attaching card to editcompose doesn't work
- [ ] Render markdown in ContentEditableArea

Issue #39, opened 2025-03-07
[devtools] Feedback Message Support


aacebo:
add support to messages for feedback activities and submission of feedback form via dialog.

PR #128, opened 2025-04-09
[devtools] Add Feedback UI

corinagum:
Fix: #39 
Fix: #114 

- Refactor ChatStore
- Add useOperatingSystem hook (for detecting Mac vs Windows)
- Add 'Edited' next to timestamp after a message has been edited
- Add tooltip for submitting via keyboard for `ComposeBoxToolbar`
- Fix styling of `PasteCardDialog`
- Update feedback docs with new screencaps.
- Add fetching an activity by message id in `ActivityStore`

When feedback is enabled and a user hovers over button:
Hover over Feedback button

Feedback dialog
Feedback dialog

NOT done: custom feedback


packages/devtools/src/components/ChatMessage/ChatMessage.tsx
```diff
@@ -22,38 +24,41 @@ interface ChatMessageProps {
 }
 
 const ChatMessage: FC<ChatMessageProps> = memo(
-  ({ content, streaming = false, feedback = false, sendDirection, value, onMessageAction }) => {
+  ({ content, streaming = false, feedback = true, sendDirection, value, onMessageAction }) => {
```
heyitsaamir, line 27: true by default?

packages/devtools/src/components/ChatMessage/ChatMessage.tsx
```diff
@@ -22,38 +24,41 @@ interface ChatMessageProps {
...
-            if (toolbar) {
-              (toolbar as HTMLButtonElement).focus();
-            }
-          }, 100);
+          if (!isFeedbackDialogOpen) {
+            setOpenedByKeyboard(true);
+            setIsPopoverOpen(true);
+            // Use setTimeout to ensure the Popover is rendered before trying to focus
+            setTimeout(() => {
+              const toolbar = document.querySelector(`[data-message-toolbar="${value.id}"] button`);
```
heyitsaamir, line 46: Why not have the child pick up the focus when it's rendered in a useEffect

heyitsaamir: (kinda a violation of separation of concerns here)

packages/devtools/src/components/Feedback/Feedback.styles.ts
```diff
@@ -0,0 +1,55 @@
+import { makeStyles, tokens } from '@fluentui/react-components';
+
+const useMessageClasses = makeStyles({
+  feedbackContainer: {
+    display: 'flex',
+    alignItems: 'center',
+    gap: tokens.spacingHorizontalM,
+    padding: '0.25rem 0',
+    marginTop: '0.5rem',
```
heyitsaamir, line 9: hmm, any particular reason to use rem instead of pixels for these?

packages/devtools/src/hooks/useOperatingSystem.ts
```diff
@@ -0,0 +1,12 @@
+import { useMemo } from 'react';
+
+const useOperatingSystem = () => {
+  const isMac = useMemo(
+    () => /mac/i.test((navigator as any).userAgentData?.platform || navigator.userAgent),
```
heyitsaamir, line 5: NIT: IMO this can just be a local variable. Not really needed as a hook. A user's OS won't really change right?

heyitsaamir (approved):

packages/devtools/src/components/ChatMessage/ChatMessage.tsx
corinagum: ah, thanks for the catch

packages/devtools/src/components/Feedback/Feedback.styles.ts
corinagum: rem is more accessibility-friendly - if the user zooms in or changes the default sizes of fonts, for example, the spacing/sizes scale. If I zoomed in with a font size of 16px, there's a risk that the font size won't adjust with the zoom.

packages/devtools/src/components/ChatMessage/ChatMessage.tsx
corinagum: This is on keydown (so an action is taken first) so `useEffect` doesn't apply here. Unless I'm misunderstanding your meaning?

packages/devtools/src/hooks/useOperatingSystem.ts
heyitsaamir: Ah okay, FWIW, i meant it can just be a variable. No need to be in react world since it'll get evaluated on script-load

PR #141, opened 2025-04-12
Add logging to MCP Client and ChatPrompt for better debugging

siduppal:
- Introduced detailed debug logs in McpClientPlugin to track tool fetching, function execution, and caching.
- Enhanced ChatPrompt with logging for plugin usage and function calls, improving traceability during interactions.

**🍿 Backstory:**
I built a simple Executive Assistant for myself to manage my Google Calendar and Email, and it really struggled with handling the time-zone properly. I am in NY right now and every time I'd ask it to schedule a meeting, it would schedule it in UTC so 4 hours off. Intuitively, I tried various prompting hints to encourage the LLM to fill out the time-zone in the times for the tool-calls. But it was only after the debug log showed that `start_datetime` DOES NOT want a time-zone it became clear that the LLM was ignoring my requests to fill out the time-zone offset in the ISO-8601 formatted date, and there is a separate `timezone` param that has to be filled out. After that, it was a one liner change to the prompt to remind the LLM to fill out the time-zone and that fixed it.

[DEBUG] @microsoft/spark.ai/prompts/chat     "name": "GOOGLECALENDAR_CREATE_EVENT",
[DEBUG] @microsoft/spark.ai/prompts/chat     "description": "Create a new event in a google calendar.",
**[SNIPPED A BUNCH OF LINES FOR BREVITY]**
[DEBUG] @microsoft/spark.ai/prompts/chat         "timezone": "IANA timezone name (e.g., 'America/New_York'). **Required if datetime is naive. If datetime includes timezone info (Z or offset), this field is optional and defaults to UTC.**",
[DEBUG] @microsoft/spark.ai/prompts/chat         "start_datetime": "Naive date/time (YYYY-MM-DDTHH:MM:SS) with **NO offsets or Z. e.g. '2025-01-16T13:00:00'",**


packages/ai/src/prompts/chat.ts
```diff
@@ -351,8 +382,20 @@ export class ChatPrompt<
...
+    // Log function calls if present
+    if (output.function_calls && output.function_calls.length > 0) {
+      this._log.debug('LLM requested function calls:', output.function_calls.map(call => ({
+        name: call.name,
+        id: call.id,
+        arguments: call.arguments
+      })));
+    }
+
+    this._log.debug('Processing plugins after send');
```
heyitsaamir, line 395: Let's only log this if plugins.length >= 0. Otherwise it'll get noisy for the vast majority of use-cases.

packages/ai/src/prompts/chat.ts
```diff
@@ -279,8 +285,10 @@ export class ChatPrompt<
   }
 
   async send(input: string | ContentPart[], options: ChatPromptSendOptions<TOptions> = {}) {
+    this._log.debug('Processing plugins before send');
```
heyitsaamir, line 288: Let's only log this if plugins.length >= 0. Otherwise it'll get noisy for the vast majority of use-cases.

packages/ai/src/prompts/chat.ts
```diff
@@ -258,11 +261,14 @@ export class ChatPrompt<
   ): this {
     const plugin = this._plugins.find((p) => p.name === name);
     if (!plugin) {
+      this._log.debug(`Plugin "${name}" not found`);
       throw new Error(`Plugin "${name}" not found`);
     }
 
+    this._log.debug(`Using plugin "${name}" with args:`, args);
```
heyitsaamir, line 268: Let's only log this if plugin has an `onUsePlugin`. Otherwise it'll get noisy for the vast majority of use-cases.

external/mcpclient/src/mcp-client-plugin.ts
```diff
@@ -115,6 +115,7 @@ export class McpClientPlugin implements ChatPromptPlugin<'mcpClient', McpClientP
       .map(([url, params]) => {
         const paramsToFetch = params ?? this._cache[url] ?? undefined;
         if (paramsToFetch == null) {
+          console.debug(`[MCP Client] Need to fetch tools for URL: ${url}`);
```
heyitsaamir, line 118: @aacebo - thinking this can benefit from an ILogger too. Create it or have it be passed in?

aacebo: yes, we should allow an optional logger to be passed in and if null, initialize an internal `ConsoleLogger` like we do in the other packages.

SidU: will address, thx

packages/ai/src/prompts/chat.ts
SidU: will address, thanks

packages/ai/src/prompts/chat.ts
SidU: I think it's useful to have a hint in the logs that there weren't any plugins. Will address.

packages/ai/src/prompts/chat.ts
SidU: I think it's useful to have a hint in the logs that there weren't any plugins. Will address.

heyitsaamir (approved):

heyitsaamir (dismissed):

external/mcpclient/src/mcp-client-plugin.ts
```diff
@@ -2,6 +2,10 @@ import { Client, ClientOptions } from '@modelcontextprotocol/sdk/client/index.js
 
 import { ChatPromptPlugin, Function, Schema } from '@microsoft/teams.ai';
 
+import { ILogger } from '@microsoft/teams.common';
+
+import { ConsoleLogger } from '../../../packages/common/dist/logging/console';
```
aacebo, line 7: this is an incorrect import

aacebo (approved):

external/mcpclient/package.json
```diff
@@ -30,7 +30,8 @@
   },
   "peerDependencies": {
     "@modelcontextprotocol/sdk": "^1.9.0",
-    "@microsoft/teams.ai": "*"
+    "@microsoft/teams.ai": "*",
+    "@microsoft/teams.common": "*"
```
aacebo, line 34: this needs to be added to the turbo.json as a build dep as well

aacebo (changes requested):
one minor thing

external/mcpclient/package.json
heyitsaamir: good call!

aacebo (approved):

PR #142, opened 2025-04-14
[Client, Apps] Validate entra app token when remote functions are invoked

Jesperholmbergmsft:
This apps an entra token validator class that uses jwks-rsa and jsonwebtoken to validate the entra auth token when remote functions are invoked.

How tested:
 - extensive unit tests
 - scaffolded a new tab+bot project with TTK and the spark embed config, it just works out of the box.
 - in a test app, played around with a number of token validation scenarios to ensure that they work when expected, and rejected when appropriate: single-tenant app, multi-tenant with and without tenant restrictions, different scopes, expiration, audiences, token purposes, etc.

### DIFF
--- a/packages/apps/src/app.embed.ts
+++ b/packages/apps/src/app.embed.ts
@@ -20,7 +20,7 @@ export function func<TData>(
     `/api/functions/${name}`,
     withClientAuth({
       logger: log,
-      entraTokenValidator: this.entryTokenValidator,
+      entraTokenValidator: this.entraTokenValidator,
       ...this.credentials,
     }),
     async (req: ClientAuthRequest, res) => {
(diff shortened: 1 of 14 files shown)


packages/apps/src/app.embed.ts
```diff
@@ -20,6 +20,7 @@ export function func<TData>(
     `/api/functions/${name}`,
     withClientAuth({
       logger: log,
+      entraTokenValidator: this.entryTokenValidator,
```
heyitsaamir, line 23: `entraTokenValidator`?

heyitsaamir (approved):
Minor comments from me

packages/apps/src/app.embed.ts
Jesperholmbergmsft: Yes - absolutely, thanks for spotting that!

PR #143, opened 2025-04-14
[Fix] Refresh tokens if they are expired

heyitsaamir:
This fixes a bug where the tokens weren't refreshing automatically if the app was just idling.
In this PR, we:
1. Expose expiration time in the `IToken`
2. Refresh token if necessary `onActivity` (checks if token is expired or not). Uses a 5 min buffer (so it'll refresh within 5 mins of expiration)
3. Add tests


packages/apps/src/app.ts
```diff
@@ -366,6 +359,44 @@ export class App {
...
+      }
+    }
+  }
+
+  private async refreshGraphToken(force = false) {
+    if (this.credentials) {
+      // Only do it if the token isn't there, or if it's expired, or if force is true
+      if (
+        !this._tokens.graph ||
+        (this._tokens.graph.expiration != null && this._tokens.graph.expiration < Date.now()) ||
```
heyitsaamir, line 390: Should we add a threshold? Like maybe 20ms?

aacebo: I would go as far as 5 minutes, just to give a ton of buffer for long running requests.

packages/apps/src/app.ts
```diff
@@ -157,10 +157,10 @@ export class App {
   /**
    * the apps auth tokens
    */
-  get tokens() {
+  get tokens(): AppTokens {
     return this._tokens;
   }
-  protected _tokens: AppTokens = {};
+  protected _tokens: Partial<Record<keyof AppTokens, JsonWebToken>> = {};
```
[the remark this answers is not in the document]
aacebo: Can you help me understand this? We should probably be using the `IToken` interface here instead of JsonWebToken, but also since both keys in `AppTokens` are already optional, im not sure what this achieves.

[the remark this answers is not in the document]
heyitsaamir: ummm. Nothing actually. haha it was before I had added expiration to IToken and now we don't need to use this.

packages/apps/src/app.ts
heyitsaamir: Added a 5 min buffer

aacebo (approved):

PR #144, opened 2025-04-14
[a2a] add a2a package/sample

aacebo:
**Note**

This PR does not include an A2A client, that work will be done after.

user-message
agent-message
code
console
streaming


heyitsaamir:
Closing in favor of the current a2a impl. This was really helpful to help me build that so ty!

PR #148, opened 2025-04-14
Fix config response type

heyitsaamir:
From: https://learn.microsoft.com/en-us/microsoftteams/platform/bots/how-to/bot-configuration-experience?tabs=JS1%2Cteams-bot-sdk2%2Cteams-bot-sdk3
```ts
async handleTeamsConfigFetch(_context, _configData) {
    let response = {};
    const adaptiveCard = CardFactory.adaptiveCard(this.adaptiveCardForContinue());
    response = {
       config: {
          value: {
             card: adaptiveCard,
             height: 500,
             width: 600,
             title: 'test card',
          },
          type: 'continue',
       },
    };
    return response;
}
```

Notice how it's not config.task.value?

Tested it locally and it worked:
image
image

### DIFF
--- a/packages/api/src/models/config/config-response.ts
+++ b/packages/api/src/models/config/config-response.ts
@@ -16,7 +16,7 @@ export type ConfigResponse = {
   /**
    * @member {ConfigResponseConfig} [config] The ConfigResponse config of BotConfigAuth or TaskModuleResponse
    */
-  config: ConfigAuth | TaskModuleResponse;
+  config: ConfigAuth | TaskModuleResponse['task'];
 
   /**
    * @member {string} [responseType] The type of response 'config'.


aacebo:
we should try and follow up with SMBA on this, the types in BF don't match this from what I can tell, although what you are pointing out is the expected response of a `task/fetch` , ie a `TaskResponse` https://github.com/microsoft/botbuilder-js/blob/main/libraries/botframework-schema/src/teams/index.ts#L1493

heyitsaamir:
> we should try and follow up with SMBA on this, the types in BF don't match this from what I can tell, although what you are pointing out is the expected response of a `task/fetch` , ie a `TaskResponse` https://github.com/microsoft/botbuilder-js/blob/main/libraries/botframework-schema/src/teams/index.ts#L1493

Check teams [threads](https://teams.microsoft.com/l/message/19:4PLRDRT5qVLpvXpfIHnoNohFElTC0aDzsNwfI0KHDxo1@thread.tacv2/1744671212391?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=1314f851-c930-4caa-b3e0-dbe9b8fe2737&parentMessageId=1744668279360&teamName=Spark%20%F0%9F%94%A5&channelName=Support&createdTime=1744671212391) (contains internal code links)

aacebo (approved):

PR #150, opened 2025-04-15
[Client] Remove custom teamsJs implementation

Jesperholmbergmsft:
This PR removes the custom teamsJs implementation that was added as an exploration before we decided to take a dependency on the actual teamsJs library instead.

How tested:
 - The code builds
 - Verified that the expected Spark client App types are exported


aacebo:
@Jesperholmbergmsft  this has had conflicts with main for a week, waiting for conflicts to be resolved so I can merge

Jesperholmbergmsft:
@aacebo  - thanks for the reminder, should be good to go now

PR #153, opened 2025-04-17
[RFC] Add plugin events

heyitsaamir:
Here, I'm proposing introducing `pluginEvents` as a core `App` concept.

# Why

Plugins are kind of our bread and butter. They make this whole system more extensible and keeps `app` relatively small and focused.
One thing missing from Plugins was the ability for plugins to `emit` an event and have the client application receive it. This concept already exists in `app` for core activity scenarios and other events, but it does not exist in plugins.

# What am I proposing

This change has the following changes:
1. Add a separate emitter for plugins - `app` currently has an event emitter for core events. The app uses them through `this.emit('eventName', eventArgs)`, and the client app can listen to it via `app.event('eventName', (eventArgs) => { ... })`. The even emitter here is type-safe, so `eventName` is a literal, and `evenArgs` corresponds to the associated type for `eventName`. I attempted to reuse this event emitter AND make it typesafe, but it was just too complicated, and after many hours of banging my head against it, I think the simpler and cleaner solution is to create a new `pluginEvent` event emitter in `app`.
Use it like this:
```ts
app.pluginEvent(A2APlugin, 'bar', (event) => {
  event.bar;
});
```
Note that the plugin definition has to be supplied in for typescript to pick up if the event name `bar` is valid for the associated plugin and what the type for `event` is.

2. Plugins now can supply associated EventTypes. This is to make typescript happy. Here is an example:
```ts
export interface Foo {
  bar: {
    bar: string;
  };
  baz: {
    baz: string;
  };
}

/**
 * the A2A plugin
 */
@Plugin({
  name: 'a2a',
  version: pkg.version,
  description: 'an open standard for connecting agents',
})
export class A2APlugin implements IPlugin<Foo> {
  __eventType!: Foo;

  @Event('custom')
  readonly emit!: <Name extends keyof Foo>(name: Name, arg: Foo[Name]) => void;

...
```

Screen Recording 2025-04-19 at 10 53 52 PM

3. Notice that `@Event('custom')`? That basically is a new path for events to say I have custom events to emit. When you have this, you are hooking up to `pluginEvents` vs `events` event emitter.

### DIFF
--- a/packages/apps/src/app.events.ts
+++ b/packages/apps/src/app.events.ts
@@ -1,7 +1,6 @@
 import { EventHandler } from '@microsoft/teams.common';
 
 import { App } from './app';
-import { ISender } from './types';
 import {
   IActivityEvent,
   IActivityResponseEvent,
@@ -9,6 +8,7 @@ import {
   IErrorEvent,
   IEvents,
 } from './events';
+import { IPlugin, IPluginWithEvents, ISender } from './types';
 
 /**
  * subscribe to an event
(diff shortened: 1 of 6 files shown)


packages/apps/src/app.events.ts
```diff
@@ -24,6 +24,30 @@ export function event<Name extends keyof IEvents>(
   return this;
 }
 
+type PluginWithEvents<TPlugin extends IPlugin> =
+  TPlugin extends IPluginWithEvents<infer TEvents> ? TEvents : never;
+
+type PluginConstructor = { new (...args: any[]): IPlugin & PluginWithEvents<any> };
+type PluginInstance = IPlugin & PluginWithEvents<any>;
+
+export function pluginEvent<
```
aacebo, line 33: the `app.event` that exists is where plugin events get emitted today, we should keep using that method to subscribe to plugin events. I think by implementing a generic for event types we can make this more implicit so that when plugins are imported the app automatically infers the events needed for the `app.event` method intellisense.

heyitsaamir: Let me try once more to have these two types of events be merged :)

heyitsaamir:
Closing in favor of https://github.com/microsoft/teams.ts/pull/205

PR #156, opened 2025-04-18
minor auth fix

kavins14:
* `User.Read` permission is too restrictive. `User.ReadBasic.All` allows for more access. Even restrictive tenants that require admin consent for most graph scopes are lenient with this one.
* `isSignedIn` was set to false when this event is emitted - when infact it shouldn't.

### DIFF
--- a/packages/apps/src/app.oauth.ts
+++ b/packages/apps/src/app.oauth.ts
@@ -36,7 +36,7 @@ export async function onTokenExchange(
       })
     );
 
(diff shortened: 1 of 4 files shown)


kavins14:
_Sidenote:_ If the `body` property is not set in the `{status: 200}` invoke response, this will trigger a err status 500 response back to the Teams client from Bot API (APX). There's no known side effects though. It's fixed by setting it to `body: {}`.

aacebo (approved):

PR #159, opened 2025-04-18
Allow binding to correct port

### LINKED ISSUE #158: [BUG] TTK configs don't make it easy to hit breakpoints
Run an agent after adding config ttk basic using Debug Edge or Debug Chrome. Notice that breakpoints are not bounded in your app.

heyitsaamir:
closes #158

Currently our ttk configs do all the setup, but the breakpoint binding wasn't happening. This requires an inspect port to be opened up when running the application. This PR adds that.

### DIFF
--- a/packages/cli/src/project/attributes/ttk.ts
+++ b/packages/cli/src/project/attributes/ttk.ts
@@ -1,9 +1,9 @@
+import fs from 'node:fs';
 import path from 'node:path';
 import url from 'node:url';
-import fs from 'node:fs';
 
-import { IProjectAttribute } from '../project-attribute';
 import { Compound, Copy, FileJsonSet, FileYamlSet, If } from '../operations';
+import { IProjectAttribute } from '../project-attribute';
 
 export class TeamsToolkitAttribute implements IProjectAttribute {
   readonly id: string;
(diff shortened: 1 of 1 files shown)


packages/cli/src/project/attributes/ttk.ts
```diff
@@ -27,13 +27,13 @@ export class TeamsToolkitAttribute implements IProjectAttribute {
         targetDir,
         'package.json',
         'scripts.dev:teamsfx',
-        'npx env-cmd --silent -f .env npm run dev'
+        "NODE_OPTIONS='--inspect=9239' npx env-cmd --silent -f .env npm run dev"
```
corinagum, line 30: nitpick: do we want to consider taking out `--silent`? The other day when I was having ttk troubles it was difficult for me to pin down what the errors I was getting were because there was no feedback at all. Can be for discussion later and not in this PR.

corinagum (dismissed):

packages/cli/src/project/attributes/ttk.ts
heyitsaamir: @aacebo any thoughts? I feel like we're tryna make it as close as possible to what exists, but I'm happy to be more opinionated here.

rajan-chari: Where does this log show up?  in vscode?

heyitsaamir: Terminal, i assume?

aacebo: seems fine to me to remove the silent option

heyitsaamir: Removed --silent

corinagum (approved):

PR #166, opened 2025-04-22
Add ability to add custom headers to MCP clients

heyitsaamir:
Previously, if you wanted to add custom headers to MCP clients, you'd need to recreate the transport. But adding custom headers is very common and we should support it out of the box.

In this PR we do just that.

### DIFF
--- a/external/mcpclient/src/mcp-client-plugin.ts
+++ b/external/mcpclient/src/mcp-client-plugin.ts
@@ -3,7 +3,7 @@ import { Client, ClientOptions } from '@modelcontextprotocol/sdk/client/index.js
 import { SSEClientTransport } from '@modelcontextprotocol/sdk/client/sse.js';
 import { Transport } from '@modelcontextprotocol/sdk/shared/transport.js';
 
-type McpClientServerDetails = {
+type McpClientToolDetails = {
   name: string;
   description: string;
   schema: Schema;
@@ -12,7 +12,8 @@ type McpClientServerDetails = {
 type PromiseOrValue<T> = T | Promise<T>;
 type ValueOrFactory<T> = T | (() => PromiseOrValue<T>);
 
-type McpClientPluginArgs = {
+type McpClientPluginParams = {
+  availableTools?: McpClientToolDetails[];
   /**
    * optional headers to pass in per request
    */
@@ -40,26 +41,10 @@ type McpClientPluginArgs = {
   onError?: (error: any) => PromiseOrValue<any>;
 };
 
-type AllOrNothing<T extends Record<string, any>> = T | Partial<Record<keyof T, never>>;
-
-export type McpClientPluginParams = McpClientPluginArgs & AllOrNothing<McpClientServerDetails>;
-
-const getServerDetails = (params: McpClientPluginParams): McpClientServerDetails | null => {
-  if (params.schema != null) {
-    return {
-      name: params.name,
-      description: params.description,
-      schema: params.schema,
-    };
-  }
-
-  return null;
-};
-
 /**
  * A map of Mcp client params keyed off of their corresponding urls
  */
-export type McpClientPluginParamsCache = Record<string, McpClientPluginParams[]>;
+export type McpClientPluginParamsCache = Record<string, McpClientPluginParams>;
 
 /**
  * A function that creates a transport for the Mcp client
@@ -104,7 +89,7 @@ export interface McpClientPluginUseParams {
    * If not provided, the client will fetch the params from the server
    * or use the cached params if provided
    */
-  params?: McpClientPluginParams[];
+  params?: McpClientPluginParams;
 }
 
 export class McpClientPlugin implements ChatPromptPlugin<'mcpClient', McpClientPluginUseParams> {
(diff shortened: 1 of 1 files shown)


external/mcpclient/src/mcp-client-plugin.ts
```diff
@@ -3,16 +3,48 @@ import { Client, ClientOptions } from '@modelcontextprotocol/sdk/client/index.js
 import { SSEClientTransport } from '@modelcontextprotocol/sdk/client/sse.js';
 import { Transport } from '@modelcontextprotocol/sdk/shared/transport.js';
 
-export type McpClientPluginParams = {
+type McpClientToolDetails = {
```
aacebo, line 6: any types publicly visible via class methods/constructors/properties should be exported publicly as well.

external/mcpclient/src/mcp-client-plugin.ts
```diff
@@ -203,3 +261,31 @@ export class McpClientPlugin implements ChatPromptPlugin<'mcpClient', McpClientP
     return [client, transport] as const;
   }
 }
+
+const buildSSEClientTransport = async (
```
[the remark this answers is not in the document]
heyitsaamir: Good idea

external/mcpclient/src/mcp-client-types.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  /**
+   * The params to use for the Mcp server
+   * If not provided, the client will fetch the params from the server
+   * or use the cached params if provided
+   */
+  params?: McpClientPluginParams;
+}
+
+export { McpClientPluginParams };
```
aacebo, line 73: why is this export down here instead of on line 14?

external/mcpclient/src/mcp-client-plugin.ts
```diff
@@ -189,8 +160,16 @@ export class McpClientPlugin implements ChatPromptPlugin<'mcpClient', McpClientP
     }
   }
 
-  private makeMcpClientPlugin(serverUrl: string) {
-    const transport = this.createTransport(serverUrl);
+  private async makeMcpClientPlugin(
+    serverUrl: string,
+    headers: ValueOrFactory<Record<string, string>> | undefined
```
aacebo, line 165: nit: prefer `headers?: ValueOrFactory<Record<string, string>>`

external/mcpclient/src/mcp-client-plugin.ts
```diff
@@ -189,8 +160,16 @@ export class McpClientPlugin implements ChatPromptPlugin<'mcpClient', McpClientP
...
+  ) {
+    let transport;
```
aacebo, line 167: nit: line can be shortened to a ternary like
```typescript
const transport = this.createTransport
    ? this.createTransport(serverUrl)
    : await buildSSEClientTransport(serviceUrl, headers)
```

external/mcpclient/src/mcp-client-plugin.ts
```diff
@@ -189,8 +160,16 @@ export class McpClientPlugin implements ChatPromptPlugin<'mcpClient', McpClientP
...
+    if (this.createTransport != null) {
```
aacebo, line 168: nit: should typically use !== for Comparisions. Not a huge deal tho.

external/mcpclient/src/mcp-client-plugin.ts
```diff
@@ -1,64 +1,15 @@
...
-}
+import {
+  CreateTransport,
+  McpClientPluginOptions,
+  McpClientPluginParams,
+  McpClientPluginParamsCache,
+  McpClientPluginUseParams,
+  McpClientToolDetails,
+  ValueOrFactory,
+} from './mcp-client-types.js';
```
aacebo, line 11: why are these imports for `.js` file extensions?

external/mcpclient/src/mcp-client-types.ts
heyitsaamir: brain fart

external/mcpclient/src/mcp-client-plugin.ts
heyitsaamir: for null-checks it's ok (that's the only one where it makes sense. it checks for both undefined, or null)

external/mcpclient/src/mcp-client-plugin.ts
heyitsaamir: Hm, I think i prefer this one, because then it's a clear decision that `headers` are not required (I actually forgot in a few places so I changed it to this). Lmk if you disagree!

external/mcpclient/src/mcp-client-plugin.ts
```diff
@@ -189,8 +239,16 @@ export class McpClientPlugin implements ChatPromptPlugin<'mcpClient', McpClientP
     }
   }
 
-  private makeMcpClientPlugin(serverUrl: string) {
-    const transport = this.createTransport(serverUrl);
+  private async makeMcpClientPlugin(
```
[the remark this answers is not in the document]
heyitsaamir: It's okay. If the custom transport is provided, they can create headers based on the serverUrl

external/mcpclient/src/mcp-client-plugin.ts
aacebo: if thats the case then I would consider using `null` like you did on line 41 for consistency. Non blocking tho.

PR #172, opened 2025-04-23
Add snippet for sending cards to cards-sample

heyitsaamir:
Docs were missing _how_ to send adaptive cards. This adds it in the sample, and provides it in the docs too by providing a snip.

### DIFF
--- a/samples/cards/src/index.ts
+++ b/samples/cards/src/index.ts
@@ -131,39 +131,6 @@ function createJsonCard() {
   return rawCard;
 }
 
-// :snippet-start: task-form-card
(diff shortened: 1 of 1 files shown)


samples/cards/src/index.ts
```diff
@@ -131,39 +131,6 @@ function createJsonCard() {
   return rawCard;
 }
 
-// :snippet-start: task-form-card
```
heyitsaamir, line 134: Replacing in favor of the e2e example

corinagum (approved):

PR #173, opened 2025-04-23
Fix submit action data and task-fetch types

heyitsaamir:
Not entirely sure if this is the right way to fix this, but it seems to be working.

1. I found out that for SubmitActions, if I want to be able to supply data values without it being in `msteams`, I currently cannot do that. So I updated `withData` in `SubmitAction` to  expect data that includes `msteams` in it, but also other values. This allows me to add data to SubmitAction that is not necessarily inside `msteams`:
```json
{
  "msteams": {
    "type": "foobar"
  },
  "other": "values"
}
```

2. Task-fetch - The current issue is that if you want to send extra data with task fetch, you need to do it in the `data` property, not in the `data.msteams.data` property, like it's currently doing. With this change, the following becomes possible. Without it, we get no values back in `value.data` other than `type: "task/fetch"`.

```ts
app.on('message', async ({ send }) => {

  // Create the launcher adaptive card
  const card: ICard = new Card()
    .withBody({
      type: 'TextBlock',
      text: 'Select the examples you want to see!',
      size: 'large',
      weight: 'bolder',
    })
    .addActions(
      // raw action
      {
        type: 'Action.Submit',
        title: 'Simple form test',
        data: {
          msteams: {
            type: 'task/fetch',
          },
          dialogtype: 'simple_form',
        },
      },
      // Special type of action to open a dialog
      new TaskFetchAction({})
        .withTitle('Multi-step Form')
        // This data will be passed back in an event so we can
        // handle what to show in the dialog
        .withValue(new TaskFetchData({ dialogtype: 'multi_step_form' })),
      new TaskFetchAction({})
        .withTitle('Webpage Dialog')
        .withValue(new TaskFetchData({ dialogtype: 'webpage_dialog' })),
      new TaskFetchAction({})
        .withTitle('Mixed Example')
        .withValue(new TaskFetchData({ dialogtype: 'mixed_example' }))
    );

  console.log(JSON.stringify(card, null, 2));

  // Send the card as an attachment
  await send(new MessageActivity('Enter this form').addCard('adaptive', card));
});
```

Clicking the first button gives the following in `dialog.submit`:
```json
{
...
  value: {
    data: { dialogtype: 'simple_form', type: 'task/fetch' },
    context: { theme: 'default' }
  },
...
```

Also in my testing, I found that supplying in `type` in `data` would just not send the invokes back. So I'm omitting it from types. I wonder if this is the case for any data, or just this type.


packages/cards/src/actions/submit/task-fetch.ts
```diff
@@ -2,6 +2,13 @@ import { ISubmitAction, MSTeamsData, SubmitAction, SubmitActionOptions } from '.
 
 export type TaskFetchActionOptions = SubmitActionOptions & { data: MSTeamsData<ITaskFetchData> };
 
+export type ValidTaskFetchDataValues = {
```
heyitsaamir, line 5: This is an opportunistic fix that I'm happy to revert for now. I found that `type` being supplied to data was causing it to not send invokes back. Happy to do more testing before checking this in.

aacebo: no this seems fine, only nit is that unless there is an `InvalidTaskFetchDataValues` I think we can just call this type `TaskFetchDataValues`, since it can be implied as valid data if it makes our types

packages/cards/src/actions/submit/collab-stage.ts
```diff
@@ -46,7 +46,7 @@ export class CollabStageAction extends SubmitAction implements ICollabStageActio
   }
 
   withData(value: ICollabStageData) {
-    this.data.msteams = value;
```
heyitsaamir, line 49: (technically i don't need to do this, and i can keep it as is if preferred)

aacebo (approved):

PR #174, opened 2025-04-24
[tests] Rename samples to tests workspace

corinagum:
This is a straight Find All -> Replace All change.

@samples to @tests


turbo/generators/templates/sample-package.json.hbs
```diff
@@ -1,5 +1,4 @@
-{
```
heyitsaamir, line 1: all the /samples-* files will need to be renamed

heyitsaamir: and the config.ts file as well

corinagum: 🤦🏻‍♀️ Thanks for catching. Updated!

heyitsaamir (approved):

Issue #185, opened 2025-04-26
[ai] citations don't show up


heyitsaamir:
The types for citations and sensitivity labels are wrong.

They need to be inside Message, under the citation key or usageInfo. But they're all set in root-level message entity right now.

Look at the interface in v1: https://github.com/microsoft/teams-ai/blob/6473c84d4bb9fa73895e86ec58589cd3b112fcb5/js/packages/teams-ai/src/types/AIEntity.ts#L12

Without this I get:
```
[ERROR] @teams/app Request failed with status code 400
console.ts:68
[ERROR] @teams/app /amer/72f988bf-86f1-41af-91ab-2d7cd011db47/v3/conversations/a:1dXpqANu3xMYb5EWLxNKW3wIODq2Q_xCHlnZC5yhXs59-3gI_vkDW-P2o_YdXa0dh2FVPxzKNm7pjJ41ilafsgAeFHtPzvGEUWuJUY4Ij5InMsuZYjmkceuZg47DBAYf3/activities
console.ts:68
[ERROR] @teams/app {
console.ts:68
[ERROR] @teams/app   "error": {
console.ts:68
[ERROR] @teams/app     "code": "BadSyntax",
console.ts:68
[ERROR] @teams/app     "message": "Multiple root message entities found in the bot activity"
console.ts:68
[ERROR] @teams/app   }
console.ts:68
[ERROR] @teams/app }
```

PR #186, opened 2025-04-26
Fix AI Entity bug (citations / sensitivity usage)

heyitsaamir:
Fixes https://github.com/microsoft/teams.ts/issues/185

The backend expects 1 single rootlevel "message" entity. Running `addCitation` would result in many of these root-level message entities. See BE logic [here](https://skype.visualstudio.com/SCC/_git/async_messaging_botapiservice?path=/Library/Utils/MessageEncoderUtils.cs&version=GBmaster&line=73&lineEnd=74&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

In this PR:
1. We fix up types to be more correct (validated with BE and v1)
2. Fix up `activity` to ensure a single root-level message entity


packages/api/src/activities/activity.ts
```diff
@@ -333,16 +336,42 @@ export class Activity<T extends string = string> implements IActivity<T> {
     return this;
   }
 
+  /**
+   * Get or create the base message entity.
+   */
+  private ensureSingleRootLevelMessageEntity(): MessageEntity {
```
aacebo, line 342: nit: privates should always come after public methods

aacebo: this is interesting logic, but I think its validating in the client something we should either be validating in the server (not sure that it is) or throw a hard exception. If this is invalid here it seems like the logic says it would just silently do nothing.

heyitsaamir: So basically if i do something like
`addAIGenerated().addCitation(...).addCitation(...)`, the logic about it being a root-entity etc etc, is all abstracted away from the developer. The library is saying it's managing all that for the developer. So I think we need to make sure we only keep one of these.

The backend is already throwing a hard exception when this happens

aacebo: oh I see what you are saying now, that makes sense

aacebo (approved):

PR #187, opened 2025-04-26
OAuth config & docs improvements 

kavins14:
* fixed minor bug where `isSignedIn` was false when it shouldn't be.
* update oauth ttk readme
* update scope to be `User.ReadBasic.All` which has a broader scope
* added bluehawk snippets


CONTRIBUTING.MD
```diff
@@ -24,6 +24,8 @@ source ~/.zshrc # or ~/.bashrc
 
 - Run `cargo install mdbook` to install the mdbook CLI globally.
 - Run `cargo install mdbook-theme` to install the theme CLI.
+- Run `cargo install mdbook-mermaid` to install the mermaid preprocessor.
+- Run `cargo install mdbook-alerts` to install the alerts preprocessor.
```
heyitsaamir, line 28: We can get rid of all these. These apply to teams-ai now.

tests/auth/src/index.ts
```diff
@@ -1,62 +1,36 @@
...
 
   const me = await api.user.me.get();
   log.info(`user "${me.displayName}" already signed in!`);
 });
 
-app.event('signin', async ({ send, api }) => {
+app.event('signin', async ({ send, api, token }) => {
   const me = await api.user.me.get();
-
-  await send(
```
heyitsaamir, line 31: I kinda liked this result vs. the new one. Can we include the token in an optional showcard?

heyitsaamir (approved):
Minor comments

tests/auth/src/index.ts
kavins14: I removed it because it would clutter the snippets. I could create a simpler card though.

heyitsaamir: Fair! Atleast throw it in a console.log after.

heyitsaamir (approved):

heyitsaamir (approved):

PR #189, opened 2025-04-28
[apps] fix streaming retry logic

aacebo:
found an issue where chunks had retry logic but not the final message, also fixed some minor other things.

### DIFF
--- a/packages/apps/src/app.process.ts
+++ b/packages/apps/src/app.process.ts
@@ -154,11 +154,12 @@ export async function $process(this: App, sender: ISender, event: IActivityEvent
   try {
     let res = await next();
 
+    await context.stream.close();
+
     if (!res || !isInvokeResponse(res)) {
       res = { status: 200, body: res };
     }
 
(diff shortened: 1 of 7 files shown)


packages/apps/src/app.process.ts
```diff
@@ -154,11 +154,12 @@ export async function $process(this: App, sender: ISender, event: IActivityEvent
   try {
     let res = await next();
 
+    await context.stream.close();
+
     if (!res || !isInvokeResponse(res)) {
       res = { status: 200, body: res };
     }
```
heyitsaamir, line 161: NIT: does this matter?

heyitsaamir (approved):

packages/apps/src/app.process.ts
aacebo: just thought it would be easier to read if we close the stream, ie finish sending everything and hit any errors that may occur, before doing response stuff. Won't affect code logically.

PR #191, opened 2025-04-28
Add mcp server sample

heyitsaamir:
1. Makes some changes to the mcp server plugin
  - Makes tool, resource, and prompt, return `this` values so that they can be used like other builder patterns.
2. Updates the samples so that
  - We use tool directly
  - We don't use ChatPrompt (ChatPrompt was a bit confusing to explain in docs - why are we using ChatPrompt when there's no generation happening here at all?)
  - We show how we can use an MCP server to send a message to a user on Teams

Question:
* Should we use chat-prompt? ChatPrompt requires `model` as a parameter, whereas an mcp server doesn't have that at all. Also, ChatPrompt is used for generation. The handler piece in the chat-prompt is useful, can we just use that part? Without a dependency on IChatPrompt?


tests/mcp/.vscode/launch.json
```diff
@@ -0,0 +1,95 @@
+{
```
heyitsaamir, line 1: Remove

tests/mcp/.vscode/tasks.json
```diff
@@ -0,0 +1,204 @@
+// This file is automatically generated by Teams Toolkit.
```
heyitsaamir, line 1: Remove

tests/mcp/env/.env.local.user
```diff
@@ -0,0 +1,2 @@
+SECRET_BOT_PASSWORD=crypto_4de3da77e60fdf48374789d40dd40b138b3d85756d42bc8df9fbc66985eac03a1a8facc5c059889b15a0084b456d6217b653bf4fefd12e759fd1c10f212518f7b43cf38a1463b4531ee0a52ee23e855895be1cb5105e8c8141990d93d06dd99cde16b57721e9c34168abfa8dc28a71e420f86c92a9e10713c3d7fbba100d711cf7e05d5d12be52b9
```
corinagum, line 1: Accidentally added file

tests/mcp/env/.env.testtool
```diff
@@ -0,0 +1,5 @@
+TEAMSFX_ENV=testtool
```
corinagum, line 1: Do we want to check in testtool files? Seems unnecessary because of DevTools

corinagum (commented):
For the purpose of testing, I don't see a need to keep this separate from another test app like the original echo bot. Do you think it's important to have this isolated from echobot? 

also, maybe we can bring it up in discussion, but I don't think it's necessary to include the `testtool` related TTK files :)

tests/mcp/src/index.ts
```diff
@@ -1,50 +1,116 @@
...
+);
+// :snippet-end:
+
+// :snippet-start: mcp-server-message-handler-store-conversation-id
 app.on('message', async ({ send, activity }) => {
   await send({ type: 'typing' });
   await send(`you said "${activity.text}"`);
+  if (activity.from.aadObjectId && !userToConversationId.has(activity.from.aadObjectId)) {
+    userToConversationId.set(activity.from.aadObjectId, activity.conversation.id);
+    console.log(
```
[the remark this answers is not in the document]
heyitsaamir: Good call

tests/mcp/env/.env.local.user
heyitsaamir: aah

aacebo (approved):

PR #195, opened 2025-05-01
Make mcp client more resilient

heyitsaamir:
1. Default to MCP clients not throwing if there's a server error (this can happen if the server is temporarily down)
2. Adds a way to refetch tools after some timeout - without refetching, the server may have updated, but the client will only use the cached version
3. Add tests


tests/ai/src/index.ts
```diff
@@ -1,23 +1,23 @@
...
+import { DevtoolsPlugin } from "@microsoft/teams.dev";
 // :snippet-start: ai-imports
-import { ChatPrompt } from '@microsoft/teams.ai';
-import { OpenAIChatModel } from '@microsoft/teams.openai';
+import { ChatPrompt } from "@microsoft/teams.ai";
+import { OpenAIChatModel } from "@microsoft/teams.openai";
 // :snippet-end:
-import { MessageActivity } from '@microsoft/teams.api';
-import { ConsoleLogger } from '@microsoft/teams.common';
+import { MessageActivity } from "@microsoft/teams.api";
```
corinagum, line 7: These should be single quotes. Could you run `fmt`?

heyitsaamir: Adding linting for tests/samples in a diff PR as per out of band discussion

PR #197, opened 2025-05-01
[RFC][WIP] Differentiate send and reply

heyitsaamir:
This is an initial crack at differentiating `send` and `reply`. 

Currently, `send` sends the activity to the same `conversation` as the one that came in. No changes.
`reply` adds a `replyToId` to the activity before sending it.

Problems:
1. `send` - if the conversation id contains the message id (eg. `19:ff4030b352c34ba4aecb654aa25c6a27@thread.tacv2;messageid=1746121818956`), the backend will treat it as a reply and add it to the thread automatically.
2. `reply` - currently doesn't differ at all from `send`. I think the expected behavior was for reply to do a "quote-reply" like we do. But that's actually not possible.

In this PR, I am proposing two changes:
1. Send should send the message to the main channel / group chat.
2. Reply should include the reply in the thread in question - in 1:1 chat, this will be no different than `send` b/c it doesn't support threads.

```ts
app.on('message', async ({ send, reply, activity }) => {
   await send(`You said ${activity.text}`); // Sends to the main channel
   await reply(`You said ${activity.text}`); // Sends as a reply to the thread
}); 
```

Remaining work:
[x] - Test if this works
[ ] - Validate from backend team if the logic to get original channel is fine
[ ] - The plugin hooks only work for send, not replies. We should fix that right?
[ ] - This is kind of a breaking change, so we should get this in earlier, and fix all the docs to use `reply`.

### DIFF
--- a/packages/apps/src/contexts/activity.ts
+++ b/packages/apps/src/contexts/activity.ts
@@ -178,13 +178,19 @@ export class ActivityContext<T extends Activity = Activity> implements IActivity
   }
 
   async send(activity: ActivityLike) {
-    return await this._plugin.send(toActivityParams(activity), this.ref);
+    const ref = {
+      ...this.ref,
+      conversation: {
+        ...this.ref.conversation,
+        id: this.stripOutThreadId(this.ref.conversation.id),
+      },
+    };
+    return await this._plugin.send(toActivityParams(activity), ref);
   }
 
   async reply(activity: ActivityLike) {
-    activity = toActivityParams(activity);
-    activity.replyToId = this.activity.id;
-    return this.send(activity);
+    console.log('replying to activity', toActivityParams(activity), this.ref);
+    return await this._plugin.send(toActivityParams(activity), this.ref);
   }
 
   async signin(options?: Partial<SignInOptions>) {
@@ -284,5 +290,13 @@ export class ActivityContext<T extends Activity = Activity> implements IActivity
       signin: this.signin.bind(this),
       signout: this.signout.bind(this),
     };
+  }
+
+  private stripOutThreadId(conversationId: string): string {
+    const parts = conversationId.split(';');
+    if (parts.length > 1) {
+      return parts[0];
+    }
+    return conversationId;
   }
 }


ryanbliss (approved):
LGTM

packages/apps/src/contexts/activity.ts
```diff
@@ -285,4 +291,12 @@ export class ActivityContext<T extends Activity = Activity> implements IActivity
       signout: this.signout.bind(this),
     };
   }
+
+  private stripOutThreadId(conversationId: string): string {
```
aacebo, line 295: very minor thing, but this could be a useful util to but on the `Activity` class in the `api` package if we think devs may need to do it ever. If not disregard.

aacebo:
want to add that yes we should make the plugin hooks work with this ideally

packages/apps/src/contexts/activity.ts
```diff
@@ -178,13 +178,19 @@ export class ActivityContext<T extends Activity = Activity> implements IActivity
   }
 
   async send(activity: ActivityLike) {
-    return await this._plugin.send(toActivityParams(activity), this.ref);
+    const ref = {
+      ...this.ref,
+      conversation: {
```
aacebo, line 183: we may want to move this logic to the http plugin, there is already some manipulation happening there and it would be nice to have it all in one place. This logic is also highly teams specific (how the id is being structured), this logic may break compatibility with webchat/sharepoint and leadership has hinted that we may want to support those in the long term.

packages/apps/src/contexts/activity.ts
```diff
@@ -178,13 +178,19 @@ export class ActivityContext<T extends Activity = Activity> implements IActivity
...
+      },
+    };
+    return await this._plugin.send(toActivityParams(activity), ref);
   }
 
   async reply(activity: ActivityLike) {
-    activity = toActivityParams(activity);
-    activity.replyToId = this.activity.id;
-    return this.send(activity);
+    console.log('replying to activity', toActivityParams(activity), this.ref);
```
aacebo, line 192: this should use the `ILogger`. We should ideally never be using `console.*`.

aacebo (changes requested):
a few comments with minor changes. Overall great find!

packages/apps/src/contexts/activity.ts
```diff
@@ -285,4 +291,12 @@ export class ActivityContext<T extends Activity = Activity> implements IActivity
       signout: this.signout.bind(this),
     };
   }
+
+  private stripOutThreadId(conversationId: string): string {
+    const parts = conversationId.split(';');
```
aacebo, line 296: no need to pass in conversationid, can just use `this.ref.conversation.id`

aacebo:
> Could possibly add some markdown in a quoteblock to mimic the reply to behavior in teams chats, though might be too opinionated. image

this is a good idea, we might even long term want two methods (not now cause threads aren't public), one for thread reply and one for quote reply. Here is an example attachment payload for a quote reply
```html
<blockquote itemscope="" itemtype="http://schema.skype.com/Reply" itemid="1746383365913">
    <strong itemprop="mri" itemid="28:931280bc-4abd-45b8-a3c6-455411603679">
        oauth-test-local
    </strong>
    <span itemprop="time" itemid="1746383365913"></span>
    <p itemprop="preview">hi!</p>
</blockquote>
<p>test</p>
```

PR #198, opened 2025-05-01
Proactive notification example

heyitsaamir:
Adding proactive notification example (and snippet) to echo bot.

Happy to move it to its own sample, but I thought it would be a simple addition to echo bot.

image


tests/echo/src/mockReminderService.ts
```diff
@@ -0,0 +1,15 @@
+export class MockReminderService<T> {
```
aacebo, line 1: filename should be kabab-case

PR #199, opened 2025-05-01
[Client] Update app options

Jesperholmbergmsft:
This updates the client app options to improve usability. Summary of the changes:

- Removing tenantId from the app options, as it's no longer used.
- Adding msalInstance to the msalOptions, to make it easier to pick up this library if you're already using MSAL. 
- Moving baseUrl to a new remoteApiOptions member, to make it easier to understand its purpose.
- Adding remoteAppResource to the new remoteApiOptions member, to make it easier to call remote functions in a different AAD app.

### DIFF
--- a/packages/client/src/app.ts
+++ b/packages/client/src/app.ts
@@ -23,18 +23,16 @@ export type MsalOptions = (
        * This is useful if you want to use a custom MSAL instance or if you want to share the
        * same MSAL instance across multiple apps.
        */
-      readonly msalInstance: msal.IPublicClientApplication;
+      readonly msalInstance?: msal.IPublicClientApplication;
       readonly configuration?: never;
     }
   | {
       readonly msalInstance?: never;
-
       /**
        * MSAL configuration to use when constructing an MSAL instance used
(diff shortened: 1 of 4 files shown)


ryanbliss (approved):
LGTM

packages/client/src/app.ts
```diff
@@ -13,12 +13,29 @@ import {
...
+    }
+  | {
+      readonly msalInstance?: never;
+
+      /**
+       * MSAL configuration to use when constructing an MSAL instance used
+       * to make authenticated function calls to remote endpoints. */
+      readonly configuration: msal.Configuration;
+    }
+  | { readonly msalInstance?: never; readonly configuration?: never }
```
heyitsaamir, line 37: Nit: do we need the never, never combo?

heyitsaamir (approved):

packages/client/src/app.ts
Jesperholmbergmsft: Good point, thanks for calling that out!

aacebo (approved):

PR #202, opened 2025-05-03
Split up `ApiClient` to ApiClient, AppGraphClient, UserGraphClient

heyitsaamir:
> [!IMPORTANT]
> This is a breaking change. Started docs here - https://github.com/microsoft/teams-ai/pull/2430/files

The `api` variable that sits in the activity context, or the `api` variable that sits in the `App` is confusing.
This is what the api value looks like:
image

Here, it's confusing to know what's what. user? users? app? http? etc. This leads to the `api` being a very confusing variable on both the activity context and the `App`. 

So before our preview launch, we should split it up like we're doing for dot-net. 

So here we have:
1. In `App`, we split `api` into `api` (which is now an `ApiClient`) and `graph` (which is a `GraphClient`).
image
image

3. In `IActivityContext`, we split `api` into `api`, `userGraph` (which is now `GraphClient`) and `appGraph`.
image


aacebo (changes requested):
always appreciate the pictures and level of detail in your PR descriptions!

There is at least one cli template using graph that most likely needs to be updated as well.

heyitsaamir:
> always appreciate the pictures and level of detail in your PR descriptions!
> 
> There is at least one cli template using graph that most likely needs to be updated as well.

Updated the CLI template, and tested:
image

heyitsaamir:
> always appreciate the pictures and level of detail in your PR descriptions!
> 
> There is at least one cli template using graph that most likely needs to be updated as well.

https://github.com/microsoft/teams.ts/pull/203 Here's a PR that makes sure both are edited

tests/auth/package.json
```diff
@@ -13,7 +13,10 @@
     "clean": "npx rimraf ./dist",
     "build": "npx tsc",
     "start": "node .",
-    "dev": "npx nodemon -w \"./src/**\" -e ts --exec \"node -r ts-node/register -r dotenv/config ./src/index.ts\""
+    "dev": "npx nodemon -w \"./src/**\" -e ts --exec \"node -r ts-node/register -r dotenv/config ./src/index.ts\"",
+    "dev:teamsfx": "NODE_OPTIONS='--inspect=9239' npx env-cmd -f .env npm run dev",
```
aacebo, line 17: not a huge deal since these are tests, but remember this method of setting env vars doesn't work in windows.

aacebo (approved):

tests/auth/package.json
heyitsaamir: You're right. I used the cli and I guess it wasn't updated. Will fix

PR #203, opened 2025-05-03
GIthub Action to sync up templates and tests

heyitsaamir:
If there are any changes in `tests/*`, and there is a corresponding template in `cli/templates/typescript/*`, then it checks if both have been modified. And vice-versa.
Does NOT validate the modification
Can be skipped if `skip-test-verification` in PR description
image


singhk97:
A few points:
1. How do you determine "corresponding template in `cli/templates/typescript/*`"?
2. Do we want the templates and tests to always be in sync? For example, the message-extensions tests will be comprehensive, but having all those features in a template might not be necessary since devs can look at the docs snippets to fill in the gaps.

heyitsaamir:
> A few points:
> 
>     1. How do you determine "corresponding template in `cli/templates/typescript/*`"?
> 
>     2. Do we want the templates and tests to always be in sync? For example, the message-extensions tests will be comprehensive, but having all those features in a template might not be necessary since devs can look at the docs snippets to fill in the gaps.

1. Using the directory name
2. We don't have to, and it's your call if you think a change should be made in the corresponding directory. You can use `skip-test-verification` to skip that verification if needed.

corinagum:
If a person skips it when they add a template, won't that mean all future PRs will need the skip? I'm not sure we want to enforce having 1:1 on test apps v.s. templates. For example, maybe we'll eventually add templates from other sources like (formerly) TTK?

heyitsaamir:
> If a person skips it when they add a template, won't that mean all future PRs will need the skip? I'm not sure we want to enforce having 1:1 on test apps v.s. templates. For example, maybe we'll eventually add templates from other sources like (formerly) TTK?

This triggers per-PR, and IFF there is a directory that's named the same in tests as templates. So if you have a template named echo, but no test, this doesn't trigger. And vice-versa.

heyitsaamir:
> If a person skips it when they add a template, won't that mean all future PRs will need the skip? I'm not sure we want to enforce having 1:1 on test apps v.s. templates. For example, maybe we'll eventually add templates from other sources like (formerly) TTK?

(though generally good practice to have a test for each template, IMO, but this PR doesn't enforce it)

corinagum (dismissed):

corinagum (approved):

PR #205, opened 2025-05-04
Update plugins to hook into `app.event` hook

heyitsaamir:
This is an evolution of #153 . In that PR, we were proposing creating a completely new `PluginEvent` emitter which also meant that apps would need to do `app.pluginEvent` instead of `app.event`. This adds coginitive overhead, that we can now avoid:

In this PR, we are allowing plugins to be able to hook into `app.event` hooks by providing their own Event map of possible events that they can emit. Here is a demo:

https://github.com/user-attachments/assets/a691b646-2cf4-40b1-8866-062bc54d436d

What this means:
1. Plugins may be able to also `emit` events if they want, by using the `emit` function. This allows applications to hook into _plugin_ events via `app.event` in a type-safe way.
To do this, they may simply use the plugin decorator `@Event('custom')`.
3. When events are merged, the priority is given to IEvents - If a plugin tries to override an event that is already defined in IEvents, the app will throw. This is the behavior i'm proposing, but I'm happy to just ignore it and log a warning.

<details>

<summary>Check out the code below</summary>

```ts
interface MyFancyRoutes {
  foo: {
    bar: string;
  },
  baz: {
    qux: string;
  };
}
class FancyPlugin implements IPlugin<MyFancyRoutes> {
  __eventType!: MyFancyRoutes;

  @Event('custom')
  emit!: <Name extends keyof MyFancyRoutes>(name: Name, arg: MyFancyRoutes[Name]) => void;

  onInit(): void | Promise<void> {
    // Start a timer for 5 seconds, then emit an event
    setTimeout(() => {
      this.emit('foo', { bar: 'baz' });
    }
      , 5000);
  }
} 

interface MyFancyRoutes2 {
  bat: {
    bar: string;
  },
}
class FancyPlugin2 implements IPlugin<MyFancyRoutes2> {
  __eventType!: MyFancyRoutes2;

  @Event('custom')
  emit!: <Name extends keyof MyFancyRoutes2>(name: Name, arg: MyFancyRoutes2[Name]) => void;

  onInit(): void | Promise<void> {
    // Start a timer for 5 seconds, then emit an event
    setTimeout(() => {
      this.emit('bat', { bar: 'baz' });
    }
      , 5000);
  }
}

const app = new App({
  logger: new ConsoleLogger('@tests/echo', { level: 'debug' }),
  plugins: [new FancyPlugin(), new FancyPlugin2(), new DevtoolsPlugin()],
});

app.on('message', async ({ send, activity }) => {
  await send(`you said "${activity.text}"`);
});

app.event('baz', event => {
  event.qux = 'bat'; // described above
})

app.event('start', event => {
  // event is of type ILogger from IEvents
})
```

</details>


.gitattributes
```diff
@@ -1 +1,2 @@
-* text eol=lf
```
heyitsaamir, line 1: do this in a diff PR

packages/apps/src/app.events.ts
```diff
@@ -8,23 +8,51 @@ import {
...
+
+// Prioritizes the keys of IEvents, and merges all the Events for the Plugins
+export type AppEvents<TPlugins> = {
+  [K in keyof IEvents | keyof MergePluginEventMaps<TPlugins>]:
+  K extends keyof IEvents
+  ? IEvents[K]
+  : K extends keyof MergePluginEventMaps<TPlugins>
+  ? MergePluginEventMaps<TPlugins>[K]
+  : never;
+};
```
heyitsaamir, line 32: move all this to a diff file?

packages/apps/src/app.plugins.ts
```diff
@@ -29,17 +30,26 @@ export function plugin(this: App, plugin: IPlugin) {
...
+  this: App<TPlugin>,
+  name: PluginName
+): IPlugin | undefined {
   return this.plugins.find((plugin) => {
     const metadata = getMetadata(plugin);
     return metadata.name === name;
   });
 }
 
+type TupleUnion<U extends string, R extends any[] = []> = {
```
heyitsaamir, line 43: we need a type-helper file

packages/apps/src/app.ts
```diff
@@ -243,15 +250,15 @@ export class App {
...
+    const plugins: Array<TPlugin> = this.options.plugins || [];
     let httpPlugin = plugins.find((p) => {
       const meta = getMetadata(p);
       return meta.name === 'http';
     }) as HttpPlugin | undefined;
 
     if (!httpPlugin) {
       httpPlugin = new HttpPlugin();
-      plugins.unshift(httpPlugin);
+      plugins.unshift(httpPlugin as any);
```
singhk97, line 261: is the `as any` avoidable?

singhk97:
How is the intellisense showing custom plugin event names when using `.event` when - in the code example - `App` is initialized without explicitly defining the `TPlugin` type?

packages/apps/src/app.ts
heyitsaamir: It might be, with some level of annoying-effort. I'll need to see...

heyitsaamir:
> What if `MyFancyRoutes` and `MyFancyRoutes2` had overlapping properties, which one would take precedence? Or would it throw a type error?

Right now it just merges them:
So if you have 
```
interface Foo1 {
   foo: {
      bar: string;
   }
}

interface Foo2 {
   foo: {
      baz: number;
   }
}
```

The final result is:
```
{
   foo: {
      bar: string;
   } & {
      baz: number
   }
}
```
Probably not the behavior we want.

We can:
1. Force plugins to have a prefix ("a2a:foo", "a2a:bar")
2. Last one wins

Will play around with this.

heyitsaamir:
> How is the intellisense showing custom plugin event names when using `.event` when - in the code example - `App` is initialized without explicitly defining the `TPlugin` type?

The same way it knows:
```ts
const myFn: <T>(input: T) => T = (input) => {
  return input
}

const x = myFn(1);
// x: 1 literal
const y = myFn("two")
// y: "two" litera

const str: string = "str"
const myStr = myFn(str);
// myStr: string
```

https://www.typescriptlang.org/docs/handbook/2/generics.html#hello-world-of-generics

> Notice that we didn’t have to explicitly pass the type in the angle brackets (<>); the compiler just looked at the value "myString", and set Type to its type. While type argument inference can be a helpful tool to keep code shorter and more readable, you may need to explicitly pass in the type arguments as we did in the previous example when the compiler fails to infer the type, as may happen in more complex examples.

packages/apps/src/app.ts
singhk97: if it's too much of a hassle then i think it's fine. just don't want to make this a common practice.

singhk97 (dismissed):

packages/apps/src/app.events.ts
aacebo: These are great, I would suggest we add some more verbose comments to the `AppEvents` since it gets exported.

packages/apps/src/app.plugins.ts
aacebo: if you could move all these "utility types" to the `./types` directory like you see with this one https://github.com/microsoft/teams.ts/blob/main/packages/apps/src/types/prefixed.ts that would be great and keep the repo tidy.

packages/apps/src/app.ts
aacebo: it might be worth throwing an warning if two plugins that support the same events are imported to an app, what do you guys think?

packages/apps/src/app.ts
```diff
@@ -36,7 +42,7 @@ import { IPlugin } from './types';
 /**
  * App initialization options
  */
-export type AppOptions = Partial<Credentials> & {
+export type AppOptions<TPlugin extends IPlugin> = Partial<Credentials> & {
```
aacebo, line 45: this should probably have the same default as the `App` class, `IPlugin`

packages/apps/src/types/plugin/sender.ts
```diff
@@ -7,7 +7,7 @@ import { IPlugin } from './plugin';
 /**
  * a plugin that can send activities
  */
-export interface ISender extends IPlugin {
+export type ISender<TCustomEvents extends {} | undefined = undefined> = IPlugin<TCustomEvents> & {
```
aacebo, line 10: is there a way to keep this an interface? Ideally the rule of thumb for the repo has been if it gets extended it should be an interface, otherwise use types.

packages/apps/src/app.plugins.ts
```diff
@@ -29,17 +30,26 @@ export function plugin(this: App, plugin: IPlugin) {
 /**
  * get a plugin
  */
-export function getPlugin(this: App, name: PluginName): IPlugin | undefined {
+export function getPlugin<TPlugin extends IPlugin>(
```
aacebo, line 33: a default for the generic of `IPlugin` would help for those that doesn't have a specific type in mind.

packages/apps/src/app.routing.ts
```diff
@@ -24,8 +24,8 @@ export function on<Name extends keyof IRoutes>(
  * @param pattern pattern to match against message text
  * @param cb callback to invoke
  */
-export function message(
-  this: App,
+export function message<TPlugin extends IPlugin>(
```
aacebo, line 27: function generics shouldn't be necessary if we have a class level generic on `App`

packages/apps/src/app.oauth.ts
```diff
@@ -99,3 +100,5 @@ export async function onVerifyState(
     return { status: 412 };
   }
 }
+export type OnTokenExchange<TPlugin extends IPlugin> = typeof onTokenExchange<TPlugin>;
```
aacebo, line 103: it would be nice to organize types a bit more consciously so that in the long term we don't have dangling types like this all over the place.

aacebo (commented):
added some feedback, I think this is headed in the right direction overall. I think my biggest thing other than the nit code organization stuff is that Id like to avoid having to specify the generic everywhere the `App` class goes, so providing a default and leaning into that where possible would be ideal.

aacebo:
note:

packages in need of updates if this is approved:
- `@microsoft/teams.botbuilder`
- `@microsoft/teams.dev`
- `@microsoft/teams.mcp`

packages/apps/src/types/plugin/plugin.ts
```diff
@@ -86,3 +82,26 @@ export interface IPlugin {
    */
   createStream?(ref: ConversationReference): IStreamer;
 }
+
+export type IPluginWithEvents<TEvents extends {}> = IBasePlugin & {
```
aacebo, line 86: same interface comment. Also is this a private type or just to be used internally byt the `App`? I'm not sure a plugin would ever want a public `emit` method since it should be emitting its events internally not allowing external acters to manually emit its events.

packages/apps/src/types/plugin/plugin.ts
```diff
@@ -86,3 +82,26 @@ export interface IPlugin {
...
+  /**
+   * The event types that this plugin can emit. This is just a type, but we need it
+   * for the type system to pick it up.
+   */
+  __eventType: TEvents;
```
aacebo, line 91: this should be invisible and added by the decorators via metadata, not explicitly required by the interface.

packages/apps/src/types/plugin/plugin.ts
aacebo: also instead of having a bunch of different types for this it would be nice for it to all be in `IPlugin` as the core type for a generic plugin. From there via inheritance we expose more targeted plugin types like `ISender`.

packages/apps/src/app.plugins.ts
```diff
@@ -81,6 +91,13 @@ export function inject(this: App, plugin: IPlugin) {
       handler = (event: IPluginActivityEvent) => {
         this.onActivity(plugin as ISender, event);
       };
+    } else if (name === 'custom') {
```
aacebo, line 94: what would be really cool is if instead of using a static key like `custom` we allowed any string literal and merged the types into app holistically, so I could do something like `app.event('mcp.task`)` (also to avoid name collisions I thinking namespacing the custom events with the app name would be ideal and save us a ton of headaches down the road.

packages/apps/src/app.ts
heyitsaamir: So right now, the events are a type-level check. The only place we do a runtime check is in apps.plugins. What it does there there is just ensures that emissions for core-level events are not take-over by a plugin. I'll warn there.

heyitsaamir: I did some work to see if any was avoidable, but it's a stupid amount of indirection to do it, so I'm just leaning toward keeping it like it is for now. Lmk if y'all disagree

packages/apps/src/types/plugin/sender.ts
heyitsaamir: This is a good rule of thumb, but with generics, this gives me problems:
image

Supposedly it's because you can't extend interfaces with types that are unions (which makes sense i think). I can add a comment here?

heyitsaamir: https://www.totaltypescript.com/concepts/an-interface-can-only-extend-an-object-type

packages/apps/src/app.ts
heyitsaamir: https://chatgpt.com/share/681d9003-b300-8010-9a3f-79f9fb1d3693 - here is a pretty reasonable explanation chatgpt gave me

packages/apps/src/app.plugins.ts
```diff
@@ -13,7 +14,7 @@ import { PLUGIN_METADATA_KEY, PluginOptions } from './types/plugin/decorators/pl
  * add a plugin
  * @param plugin plugin to add
  */
-export function plugin(this: App, plugin: IPlugin) {
+export function plugin<TPlugin extends IPlugin>(this: App<TPlugin>, plugin: TPlugin) {
```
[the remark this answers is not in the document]
heyitsaamir: Check out https://github.com/microsoft/teams.ts/pull/205#discussion_r2080939096.

packages/apps/src/app.plugins.ts
heyitsaamir: Check out https://github.com/microsoft/teams.ts/pull/205#discussion_r2080939096

packages/apps/src/app.routing.ts
heyitsaamir: Check out https://github.com/microsoft/teams.ts/pull/205#discussion_r2080939096 (i know i'm a broken record, but i did try lol to avoid this....)

packages/apps/src/app.ts
heyitsaamir: One important thing to note is that this doesn't leak out to clients. So you can still do 
```
new App()
```

without any generics

packages/apps/src/app.oauth.ts
heyitsaamir: Yeah that's totally fair. let me see if i can clean it up.

heyitsaamir: Okay I had to do this to avoid the explicit type creation:
```ts
this.on('signin.token-exchange', (...args) => this.onTokenExchange(...args));
this.on('signin.verify-state', (...args) => this.onVerifyState(...args));
```
Not a huge deal imo, but lmk if you disagree. Basically typescript doesn't know what the value of `this` is when i'm binding it unless I explicitly do 
```ts
onTokenExchange: OnTokenExchange<TPlugin> = onTokenExchange
```

packages/apps/src/types/plugin/plugin.ts
heyitsaamir: > same interface comment. Also is this a private type or just to be used internally byt the `App`? I'm not sure a plugin would ever want a public `emit` method since it should be emitting its events internally not allowing external acters to manually emit its events.

Responded to your earlier comment about interfaces not being possible when extending union types.

EDIT: Okay I went back and moved `__eventType` to be an optional property inside `IPlugin`. It just won't be super obvious that `__eventType` needs to be present to pick up events, it does follow convention a bit more. I also updated `ISender`

packages/apps/src/types/plugin/sender.ts
heyitsaamir: https://github.com/microsoft/teams.ts/pull/205#discussion_r2080985506

Decided to fix this via ^

packages/apps/src/types/plugin/plugin.ts
heyitsaamir: Without this __eventType type (it's just a type), the type-system doesn't pick up that this is a `PluginWithEvents`. It needs a concrete required type to distinguish it from IPlugin without events. 

TypeScript only infers generic arguments when they are actually used in the structure of the object/type — not just as a generic parameter. Without __eventType, we're not actually using `__eventType` from the generic anywhere. So typescript doesn't know how to infer its value when we merge it

packages/apps/src/app.plugins.ts
heyitsaamir: That's exactly what this PR is doing:
app.event('mcp.task') is possible right now. 

The namespace idea is something i was thinking about too.... I wonder if we could do it with template strings. WIll think about it more. Too sleepy now

aacebo (approved):
as long as we are sure none of the existing plugins need updating, I think this is good to go.

PR #206, opened 2025-05-04
Differentiate send and reply using quoted replies

heyitsaamir:
This is a different direction than what was proposed in https://github.com/microsoft/teams.ts/pull/197. @aacebo discovered that it is, in fact, possible to send quoted replies using the API.

In this PR we:
1. Update `reply` to add a blockquote. This blockquote is truncated to 120 characters for now.
2. Added some tests
3. Tested it
image


.vscode/settings.json
```diff
@@ -1,4 +1,8 @@
 {
   "typescript.preferences.quoteStyle": "single",
-  "css.lint.unknownAtRules": "ignore"
+  "css.lint.unknownAtRules": "ignore",
+  "editor.formatOnSave": false,
```
heyitsaamir, line 4: https://eslint.style/guide/faq#how-to-auto-format-on-save

aacebo: can we get rid of the on save one? I prefer my editor not do that personally.

heyitsaamir: Oh yeah, totally.

heyitsaamir: Happy to not do this at all.

heyitsaamir: reverted this

packages/apps/src/contexts/activity.ts
```diff
@@ -279,6 +292,26 @@ export class ActivityContext<T extends Activity = Activity> implements IActivity
...
+<p itemprop="preview">${truncatedText}</p>
+</blockquote>`;
+    } else {
+      this.log.debug('Skipping building blockquote for activity type:', this.activity.type);
+    }
+
+    return null;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/member-ordering
```
aacebo, line 314: why are you ignoring this rule instead of putting the private method below?

heyitsaamir: Hmm, I thought we wanted these to be at the bottom, stylistically, but I guess that was just a coincidence.

singhk97 (approved):
How would the feature `reply` work with the `replies` feature enabled in the teams client? 

The one where you have a separate panel for replies. It would make sending the block quotes kinda redundant.

aacebo (approved):

aacebo:
> How would the feature `reply` work with the `replies` feature enabled in the teams client?
> 
> The one where you have a separate panel for replies. It would make sending the block quotes kinda redundant.

we talked about this a bit, once that feature is out of preview we will either add an options param to the reply method that allows you to customize how the reply is sent (ie thread vs block quote vs custom block quote), or expose a new method.

PR #207, opened 2025-05-05
[cli] Add dotnet `echo` template (first draft) for initial testing

aacebo:
image


packages/cli/templates/csharp/echo/Echo.sln
```diff
@@ -0,0 +1,34 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31903.59
```
heyitsaamir, line 4: NIT: Can we put all these magic strings somewhere common?

OTOH, I can also see this being simple copies

heyitsaamir (approved):

packages/cli/src/commands/new/csharp.ts
```diff
@@ -18,7 +19,7 @@ export function CSharp(_: IContext): CommandModule<{}, Args> {
   return {
     command: 'csharp <name>',
     aliases: ['c#', 'dotnet', '.net'],
-    describe: 'create a new csharp app project',
+    describe: '!!BETA!! create a new csharp app project',
```
heyitsaamir, line 22: NIT: ⚠️BETA⚠️ ?

packages/cli/src/commands/new/csharp.ts
```diff
@@ -68,7 +69,7 @@ export function CSharp(_: IContext): CommandModule<{}, Args> {
           return true;
         });
     },
-    handler: async ({ name, template, ttk }) => {
+    handler: async ({ name, template, start, ttk }) => {
```
heyitsaamir, line 72: NIT: Not related to this line, but this file is called "csharp". Consider consistency with naming (dotnet, vs csharp etc)

packages/cli/templates/csharp/echo/Nuget.Config
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+    <packageSources>
+        <add key="github" value="https://nuget.pkg.github.com/microsoft/index.json" />
+    </packageSources>
+    <packageSourceCredentials>
+        <github>
```
heyitsaamir, line 7: Can you add comments to these files? Consider adding one?

heyitsaamir (approved):

packages/cli/templates/csharp/echo/Echo.sln
aacebo: those are auto gen from the dotnet cli, not sure of a way to make them common but I can look into it a bit.

packages/cli/src/commands/new/csharp.ts
aacebo: @heyitsaamir  can you elaborate? The template folder and command name are both `csharp`.

heyitsaamir: I meant, consider consistency for naming when it comes to dotnet. For eg, we aren't using nodejs and  typescript both. We're just using typescript. So similarly, should err just use one mane when it comes to this?

aacebo: ok I'm assuming you mean the `alias`, updated to just be `cs`.

PR #210, opened 2025-05-06
`tests/auth` improvements

kavins14:
* fix token display
* add oauth setup
* add comments to code snippets


tests/auth/src/index.ts
```diff
@@ -2,10 +2,17 @@ import { App } from '@microsoft/teams.apps';
...
-  plugins: [new DevtoolsPlugin()],
+// :snippet-start: auth-config
+const app = new App({ 
+  oauth: { // oauth configurations
+    // the name of the auth connection to use.
+    // it should be the same as the OAuth connection name defined in the Azure Bot configuration.
+    defaultConnectionName: 'graph' 
+  },
+  logger: new ConsoleLogger('@tests/auth', { level: 'debug' }), // optional
+  plugins: [new DevtoolsPlugin()], // optional
```
corinagum, line 13: I don't think optional is necessary here. Could you also change the comments in lines 8-9 to /* */?

heyitsaamir, line 13: IMO let's remove it? In Oauth flows this doesn't add much value does it?

heyitsaamir (dismissed):

tests/auth/src/index.ts
kavins14: updated

kavins14: fair, removed it

heyitsaamir (approved):

PR #214, opened 2025-05-07
[Client] Ensure exec() contextual headers contain current values

Jesperholmbergmsft:
The client `app.exec()` method includes contextual information such as current team, channel, chat, meeting, etc. To better support cached apps where a single app instance can be reused in different scenarios, this PR moves the `teamsJs.getContext()` call from the `start()` method where it was captured & cached to the `exec()` function. This ensures that the values are correct if a cached app is used in multiple channels, for instance.

The context is cheap to get. So cheap that teamsJs doesn't even bother caching it. If for whatever reason we find that it's expensive to fetch it during exec, we'll work with the teamsJs owners to cache it there so that all apps benefit.


packages/client/src/app.ts
```diff
@@ -159,17 +162,19 @@ export class App {
     this._state = { phase: 'starting' };
 
     await teamsJs.app.initialize();
-    const context = await teamsJs.app.getContext();
```
Jesperholmbergmsft, line 162: 1/2 the fix -- don't fetch context during start()

packages/client/src/app.ts
```diff
@@ -191,11 +196,17 @@ export class App {
...
    */
-  async exec<T = unknown>(name: string, data?: unknown, options?: ExecOptions): Promise<T> {
-    const { msalInstance, context } = this.appStateGuard();
+  async exec<T = unknown>(
+    name: string,
+    data?: unknown,
+    options?: ExecOptions
+  ): Promise<T> {
+    const { msalInstance } = this.appStateGuard();
+    const context = await teamsJs.app.getContext();
```
Jesperholmbergmsft, line 205: 1/2 the fix -- fetch context during exec()

corinagum (approved):

packages/client/src/app.ts
```diff
@@ -210,9 +221,9 @@ export class App {
         'x-teams-channel-id': context.channel?.id,
         'x-teams-chat-id': context.chat?.id,
         'x-teams-meeting-id': context.meeting?.id,
-        'x-teams-message-id': context.app.parentMessageId,
-        'x-teams-page-id': context.page.id,
-        'x-teams-sub-page-id': context.page.subPageId,
+        'x-teams-message-id': context.app.parentMessageId || undefined,
```
Jesperholmbergmsft, line 224: Testing the change, I noticed that we pass these headers always even though they tend to have just an empty string. No huge deal, but not really helpful either - let's normalize empty strings to undefined, so that the whole header is omitted when it doesn't apply.

packages/client/src/app.ts
```diff
@@ -159,17 +162,19 @@ export class App {
...
     if (!msalInstance) {
       const msalConfig =
-        this.options.msalOptions?.configuration ?? buildMsalConfig(this.clientId, this._log);
-      msalInstance = await msal.createNestablePublicClientApplication(msalConfig);
+        this.options.msalOptions?.configuration ??
+        buildMsalConfig(this.clientId, this._log);
+      msalInstance = await msal.createNestablePublicClientApplication(
+        msalConfig
+      );
       await msalInstance.initialize();
```
ydogandjiev, line 174: Out of curiousity, what does this initialize() call do? I don't see it mentioned in any of our NAA docs or used in any of our samples.

ydogandjiev: I also noticed that we are not getting/setting the active account in this library? In my previous usage of NAA, I found that to be needed before acquiring tokens successfully. Have you not seen any issues because of this missing logic? Here is how we do it in our official NAA sample:
https://github.com/OfficeDev/Microsoft-Teams-Samples/blob/1246ef54b97e94a298f76d967d88f91c51386bb3/samples/tab-nested-auth/nodejs/src/components/Msal.js#L40

ydogandjiev (dismissed):

packages/client/src/app.ts
Jesperholmbergmsft: You're right -- I don't need to call initialize() when using `createNestablePublicClientApplication()`, as it [initializes the PCA for us](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/36a9d5f1fe7bacbce393283483189b8fb0029a5d/lib/msal-browser/src/app/PublicClientApplication.ts#L454C9-L454C40). I'll clean that up!

As for the account, I don't know that we need to set explicitly when using `createNestablePublicClientApplication`. The account is managed in the [NestedAppAuthController](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/36a9d5f1fe7bacbce393283483189b8fb0029a5d/lib/msal-browser/src/controllers/NestedAppAuthController.ts#L142) used by the PCA and there's [documentation](https://learn.microsoft.com/en-us/office/dev/add-ins/develop/enable-nested-app-authentication-in-your-add-in#msaljs-apis-supported-by-naa) that `setActiveAccount` is unsupported in NAA. I haven't seen account related issues since adopting `createNestablePublicClientApplication` instead of constructing a PCA by hand.

packages/client/src/app.spec.ts
```diff
@@ -60,9 +60,10 @@ jest.mock('@microsoft/teams-js', () => {
 jest.mock('@azure/msal-browser', () => {
   return {
     ...jest.requireActual('@azure/msal-browser'),
-    createNestablePublicClientApplication: msalCreateNPCAppMock.mockResolvedValue({
-      initialize: msalInitializeMock,
-    }),
+    createNestablePublicClientApplication:
```
Jesperholmbergmsft, line 63: Most changes are from `npm run lint:fix`; the rest are from removing the unnecessar call to `msalInstance.initialize()`

singhk97:
Quick question - is user auth setup required for server-side exec calls from the client app? or just client id & secret?

Jesperholmbergmsft:
> Quick question - is user auth setup required for server-side exec calls from the client app? or just client id & secret?

@singhk97  - Yes, this package is intended for SPA Teams apps using NAA for authentication. When integrating this to an existing app, the [Configure NAA](https://learn.microsoft.com/en-us/microsoftteams/platform/concepts/authentication/nested-authentication#configure-naa) page is helpful to make sure the app is configured correctly.

corinagum (approved):

ydogandjiev (approved):

PR #216, opened 2025-05-09
A2A Plugin (Server)

heyitsaamir:
> [!NOTE]
> This is an initial implementation of the A2A Server plugin. It was inspired by #144 and I took a bunch of code from there.

A2A spec is currently pretty loose. They have an [initial spec](https://github.com/google/A2A/blob/main/specification/json/a2a.json) and some [sample code](https://github.com/google/A2A/tree/main/samples/js). The sample code implements some basic functionality, but definitely not everything.

> [!CAUTION]
> Given the above, we should treat A2A as experimental, and subject to heavy change.

In this PR:
1. It sets up a new A2APlugin. This plugin hooks into the app's HTTPlugin (so great that this is easily exposed!)
1. We use #205 to emit events to the app when A2A events happen
2. Currently only taskGet, and taskSend are supported (I'll follow up with taskSubscribe. I know #144 has a great first implementation of it, but I wanted to start simple)
3. Uses IStorage to manage tasks
4. Tested it locally


external/a2a/README.md
```diff
@@ -0,0 +1,27 @@
...
+    </a>
+    <a href="https://www.npmjs.com/package/@microsoft/teams.a2a" target="_blank">
+        <img src="https://img.shields.io/npm/dw/@microsoft/teams.a2a" />
+    </a>
+    <a href="https://microsoft.github.io/teams.ts" target="_blank">
+        <img src="https://img.shields.io/badge/📖 docs-open-blue" />
+    </a>
+</p>
+
+A plugin enabling a2a on a Teams agent application
```
aacebo, line 21: might be nice to link to the official a2a repo/docs from here.

external/a2a/package.json
```diff
@@ -0,0 +1,44 @@
...
+  ],
+  "scripts": {
+    "clean": "npx rimraf ./dist",
+    "build": "npx tsc",
+    "test": "npx jest",
+    "lint": "npx eslint",
+    "lint:fix": "npx eslint --fix"
+  },
+  "peerDependencies": {
+    "@microsoft/teams.common": "*",
```
aacebo, line 32: these will need to be set to the latest package version since dependencies need to be explicit when using prerelease versions.

external/a2a/turbo.json
```diff
@@ -0,0 +1,15 @@
...
+  "extends": ["//"],
+  "tasks": {
+    "build": {
+      "inputs": ["$TURBO_DEFAULT$", ".env*"],
+      "outputs": [".next/**", "!.next/cache/**"],
+      "cache": false,
+      "dependsOn": [
+        "@microsoft/teams.common#build",
+        "@microsoft/teams.apps#build",
+        "@microsoft/teams.ai#build"
```
aacebo, line 11: I don't think this is needed for ai since you aren't using it as a dependency in your package.json.

external/a2a/README.md
heyitsaamir: Yeah, actually good idea to update this readme with how this thing works

external/a2a/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@microsoft/teams.a2a",
+  "version": "0.1.0",
```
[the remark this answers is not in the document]
heyitsaamir: Fixed this + fixed the turbo generator to use the version in the root package.json

external/a2a/package.json
heyitsaamir: ah TIL. thanks!

external/a2a/turbo.json
heyitsaamir: Added dep in package.json

external/a2a/README.md
heyitsaamir: Updated readme

aacebo (dismissed):

PR #219, opened 2025-05-12
[MCPClient] Update test to use fake timers

Jesperholmbergmsft:
I found that a couple of the MCP client tests can be flaky, as they're dependent on system clock timing. To help stabilize them, this PR updates the affected tests to use fake timers instead of real ones.


external/mcpclient/src/mcp-client-plugin.spec.ts
```diff
@@ -225,7 +240,7 @@ describe('McpClientPlugin', () => {
       });
 
       // Wait for 2 ms for the timeout to happen
-      await new Promise((resolve) => setTimeout(resolve, 2));
+      jest.advanceTimersByTime(2);
```
heyitsaamir, line 243: oh neat-o

heyitsaamir (approved):

PR #233, opened 2025-05-16
For TSUP, we also type-check with no-emit

heyitsaamir:
TS obscures type errors, so we shouldn't rely on it solely for type-checking.

https://github.com/egoist/tsup/issues/609


kavins14 (approved):

heyitsaamir:
> What is the decision for when to use tsup and tsc?

TSUP when it might also get used in client, tsc for anything server only. 

I don't know if there's very many things that require tsup tbh. @aacebo i can only think of like 2 libraries that would use it right? ACs, and Teamsjs

PR #235, opened 2025-05-19
Fixes cli bugs

kavins14:
Fixes

**Before:**
`teams new quote-agent` doesn't throw an error since the language is not set.
**After:**
 Gracefully requires language to be set.
___
**Before:** 
`teams config add --help` in a repo with no `package.json` or `.sln` file throws an error. 

**After:** 
Shows the help info of the command.


packages/cli/src/commands/config/add.ts
```diff
@@ -35,6 +35,12 @@ export function Add(_: IContext): CommandModule<{}, Args> {
               .map((type) => `${name}.${type}`);
           })
           .flat(),
+      }).check(() => {
+        if (!Project.detectLanguage()) {
+          throw new Error('Invalid project. Directory should contain a package.json (typescript) or .sln (csharp) file.');
```
heyitsaamir, line 40: Instead of "Invalid project" what are your thoughts on asking a question? Like "Are you sure you are in a valid directory?"

heyitsaamir (dismissed):

packages/cli/src/commands/config/add.ts
singhk97: fixed it to something along those lines. it's a warmer error message for sure.

PR #240, opened 2025-05-22
Raw messages as part of the ChatMessageOutput

heyitsaamir:
This PR adds "raw" payloads as part of the final MessageOutput. Users may want this for something (like structured output).

skip-test-verification - I fixed the turbo.json for one of the tests.

#### PR Dependency Tree

* **PR #240** 👈
  * **PR #241**

### DIFF
--- a/packages/ai/src/message.ts
+++ b/packages/ai/src/message.ts
@@ -8,12 +8,13 @@ export type UserMessage = {
   content: string | ContentPart[];
 };
 
-export type ModelMessage = {
+export type ModelMessage<TRaw extends Record<string, any> = Record<string, any>> = {
   role: 'model';
   audio?: MessageAudio;
   content?: string;
   context?: MessageContext;
   function_calls?: FunctionCall[];
+  raw?: TRaw;
 };
 
 export type SystemMessage = {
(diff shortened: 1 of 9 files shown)


packages/ai/src/prompts/chat-types.ts
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     * the conversation history
+     */
+    readonly messages?: Message[] | IMemory;
+
+    /**
+     * Logger instance to use for logging
+     * If not provided, a ConsoleLogger will be used
+     */
+    logger?: ILogger;
```
aacebo, line 51: nit: readonly

packages/ai/src/message.ts
```diff
@@ -8,12 +8,13 @@ export type UserMessage = {
   content: string | ContentPart[];
 };
 
-export type ModelMessage = {
+export type ModelMessage<TRaw extends Record<string, any> = Record<string, any>> = {
```
aacebo, line 11: this seems odd, should we just not have our own message types in favor of using the original types? A message in a message with duplicate data seems bad for storage and confusing.

PR #241, opened 2025-05-22
Disable automatic function calling, and show structured output

heyitsaamir:
This PR allows for the disabling of automatic function calling. With this, we enable users to be able to leverage ChatPrompt to do structured output. Added a test for that too.

skip-test-verification - We really should make an ai test that's as close as possible to the template. Right now it's very different. I'll do that in a different PR.

image

#### PR Dependency Tree

* **PR #241** 👈


packages/openai/src/chat.ts
```diff
@@ -279,10 +279,14 @@ export class OpenAIChatModel implements IChatModel<ChatCompletionCreateParams, O
       };
 
       if (message.tool_calls && message.tool_calls.length > 0) {
-        return this.send(modelMessage, {
-          ...options,
-          messages: memory,
-        });
+        if (!options.disableAutomaticFunctionCalling) {
```
lilyydu: personal pref but would it be less confusing if called it enableAutoFunctionCalling - the double negation took me a bit to digest

tests/ai/src/index.ts
```diff
@@ -25,11 +26,16 @@ const app = new App({
   plugins: [new DevtoolsPlugin()],
 });
 
-const model = new OpenAIChatModel({
+console.log({
```
lilyydu: did we mean to console log here?

packages/openai/src/chat.ts
corinagum: Plus 1 to this! :)

tests/ai/src/index.ts
```diff
@@ -25,11 +26,16 @@ const app = new App({
...
   apiKey: process.env.AZURE_OPENAI_API_KEY || process.env.OPENAI_API_KEY,
   endpoint: process.env.AZURE_OPENAI_ENDPOINT,
   apiVersion: process.env.AZURE_OPENAI_API_VERSION,
-  model: process.env.AZURE_OPENAI_MODEL_DEPLOYMENT_NAME!,
+  model: process.env.AZURE_OPENAI_MODEL_DEPLOYMENT_NAME,
+});
+
+const model = new OpenAIChatModel({
+  apiKey: process.env.OPENAI_API_KEY,
+  model: 'gpt-4o',
```
corinagum: I don't think we should be hardcoding the model right? Ah I see it looks like this is just a test variable related to the console log that snuck in.

packages/openai/src/chat.ts
corinagum: nit: might be simpler to refactor like: 
```typescript
if (message.tool_calls && message.tool_calls.length > 0) {
  // Skip recursive function calling if explicitly disabled
  if (!options.enableFunctionCalling) {
    this._log.debug(`Function calling is disabled; skipping function call (total calls: ${message.tool_calls.length})`);
    return modelMessage;
  }
  
  return this.send(modelMessage, {
    ...options,
    messages: memory,
  });
```

packages/openai/src/chat.ts
```diff
@@ -279,10 +279,14 @@ export class OpenAIChatModel implements IChatModel<ChatCompletionCreateParams, O
...
+          return this.send(modelMessage, {
+            ...options,
+            messages: memory,
+          });
+        } else {
+          this._log.debug(`Automatic function calling is disabled, skipping function call (total calls: ${message.tool_calls.length})`);
```
corinagum:
```suggestion
          this._log.debug(`Automatic function calling is disabled; skipping recursive call(s) (total calls: ${message.tool_calls.length})`);
```
What is the intent when listing total calls? Is there a case where the developer would need this number when debugging structured output?

corinagum (changes requested):
Thanks for the PR @heyitsaamir! I made a couple nit suggestions / asked a question, but looks good to me!

packages/openai/src/chat.ts
heyitsaamir: Cool with this -- I usually agree with this, but in places where the default is _true_ AND the input is optional, i tend to prefer disable.

Corina your code will trigger if enableFunctionCall == null though.

heyitsaamir: updated to `options.enableFunctionCalling === false`

tests/ai/src/index.ts
heyitsaamir: It's just a test, so doesn't matter, right?

tests/ai/src/index.ts
heyitsaamir: Hm, we seem to be doing this in a number of places. Let's tackle this separately

packages/openai/src/chat.ts
heyitsaamir: yeah they'd need to execute those calls manually if they want to re-use the prompt

Jesperholmbergmsft (dismissed):

tests/ai/src/index.ts
heyitsaamir: Nvm, removing.

packages/ai/src/models/chat.ts
```diff
@@ -30,6 +30,12 @@ export type ChatSendOptions<TOptions = Record<string, any>> = {
    * stream chunk
    */
   readonly onChunk?: TextChunkHandler;
+
+  /**
+   * disable automatic function calling
+   * @default true
+   */
+  readonly enableAutomaticFunctionCalling?: boolean;
```
aacebo: nit: `autoFunctionCalling` possibly to shorten this var name, but I realize this is VERY subjective so non blocking feel free to ignore

tests/ai/src/index.ts
```diff
@@ -26,10 +27,8 @@ const app = new App({
 });
 
 const model = new OpenAIChatModel({
-  apiKey: process.env.AZURE_OPENAI_API_KEY || process.env.OPENAI_API_KEY,
-  endpoint: process.env.AZURE_OPENAI_ENDPOINT,
-  apiVersion: process.env.AZURE_OPENAI_API_VERSION,
-  model: process.env.AZURE_OPENAI_MODEL_DEPLOYMENT_NAME!,
+  apiKey: process.env.OPENAI_API_KEY,
```
aacebo: I think our samples/tests have to use azure openai by default right?

aacebo (approved):

tests/ai/src/index.ts
heyitsaamir: oh word. you're right. fixing.

PR #251, opened 2025-06-06
[repo] Fix image pointers and improve release process

corinagum:
- Fixed issue with images tracked by Git LFS in `.gitattributes`; the correct behavior is NOT to have lfs track images
- Fixed images
- Wrote a script to add a 'root' release tag (e.g. `v2.0.0-preview.X`) when publishing a release
- Wrote a pre-release script to make the release process take up less brain energy 
    - I tested this locally but the true test will be when we do a new release.
- Updated `RELEASE.md` to reflect script changes
- Added `eslint` to root package because terminal complained about it.

#minor
skip-test-verification


aacebo:
closing as stale

Issue #243, opened 2025-05-23
[devtools] Update Chat rendering to support blockquotes


corinagum:
`replyToId` functionality was updated to include the blockquote of the text when a reply is sent, but we need to render the blockquote instead of sending the html as a message. 
This is what is seen right now when a reply is sent:

![Image](https://github.com/user-attachments/assets/98e1ed11-0230-466e-87f7-9b5c88e6efa4)

PR #254, opened 2025-06-11
[devtools] Update Chat rendering to support blockquote

Jesperholmbergmsft:
# Context
This PR addresses https://github.com/microsoft/teams.ts/issues/243, where the dev tools aren't able to render quoted messages. The change here adds support for those and other common message formatting. There's still plenty ways to go if we want full parity with the Teams client - adaptive cards, deep link chiclets, code snippet colorization etc - but this moves us in the right direction. 

Before this change, the message list would simply render the message payload as text:
image

After this change, quoted messages render similar to the standard Teams client, with timestamp and all:
image

# How tested
 - ran ` npm run dev --workspace=@tests/echo` and sent myself a message. The quoted reply now renders
 - In storybook, tested other common formatting in both light & dark mode; for both sent & received messages. Looks pretty good.

# Storybook screen shots
Here's to illustrate how we support common message formatting:

Dark mode, received message
image

Light mode, sent message
image

### DIFF
--- a/packages/devtools/src/components/HtmlMessageContent/HtmlMessageContent.tsx
+++ b/packages/devtools/src/components/HtmlMessageContent/HtmlMessageContent.tsx
@@ -54,7 +54,7 @@ const getEffectiveAttributes = ({name, attribs } : {name: string, attribs: Recor
 }
 
 const renderNode = (node: any, key: number = 0): React.ReactNode => {
-  if (node.type === 'text' && !!node.data.trim())  {
+  if (node.type === 'text')  {
     return <span key={key}>{node.data}</span>;
   }
 
@@ -75,8 +75,10 @@ const renderNode = (node: any, key: number = 0): React.ReactNode => {
       case 'ol':
       case 'p':
       case 'pre':
+      case 's':
       case 'span':
       case 'strong':
+      case 'u':
       case 'ul':
         {
           const Component = node.name;


packages/devtools/package.json
```diff
@@ -28,6 +28,7 @@
     "codemirror": "^6.0.1",
     "date-fns": "^4.1.0",
     "highlight.js": "^11.11.1",
+    "htmlparser2": "^10.0.0",
```
Jesperholmbergmsft, line 31: This is what TMP uses to parse chat messages, and bot builder relies on it too.

packages/devtools/src/components/HtmlMessageContent/HtmlMessageContent.tsx
```diff
@@ -0,0 +1,106 @@
+import React from 'react';
```
Jesperholmbergmsft, line 1: This component simply parses the HTML message payload, iterates through the element tree and renders the same components out again - with some small tweaks to attributes and content. If we want to go further and support richer content - adaptive cards, deep link chiclets, @mentions etc - we can build up the renderNode() function to render richer components as and when needed. 
I kinda started by doing that - added a Link and a Code and a Quote - and then realized that I can solve the immediate problem mostly through styling.

heyitsaamir (approved):

PR #256, opened 2025-06-11
[apps] fix streaming retry logic

aacebo:
promises cannot be reused after they have rejected, this PR fixes the streaming rate limit retry logic to account for this.

### DIFF
--- a/packages/apps/src/plugins/http/stream.ts
+++ b/packages/apps/src/plugins/http/stream.ts
@@ -10,7 +10,7 @@ import {
   SentActivity,
   TypingActivity,
 } from '@microsoft/teams.api';
-import { EventEmitter } from '@microsoft/teams.common';
+import { ConsoleLogger, EventEmitter, ILogger } from '@microsoft/teams.common';
 
 import { IStreamer, IStreamerEvents } from '../../types';
 import { promises } from '../../utils';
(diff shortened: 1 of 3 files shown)


packages/apps/src/plugins/http/stream.ts
```diff
@@ -88,10 +101,9 @@ export class HttpStream implements IStreamer {
       this._timeout = undefined;
     }
 
-    const size = Math.round(this.queue.length / 10);
     let i = 0;
 
-    while (this.queue.length && i <= size) {
+    while (this.queue.length && i < 10) {
```
heyitsaamir, line 106: why limit this to 10?

packages/apps/src/utils/promises/retry.ts
```diff
@@ -7,23 +9,35 @@ export type RetryOptions = {
...
+  /**
+   * the logger to use
+   */
+  readonly logger?: ILogger;
 };
 
-export async function retry<T = any>(promise: Promise<T>, options?: RetryOptions) {
-  const max = options?.max || 3;
-  const delay = options?.delay || 200;
+export async function retry<T = any>(promise: () => Promise<T>, options?: RetryOptions) {
```
heyitsaamir: NIT: rename to promiseFactory?

heyitsaamir (dismissed):

packages/apps/src/plugins/http/stream.ts
aacebo: you need some kind of artificial limit thats smaller so that it gives a more "smooth"ish chunk rendering experience in teams, otherwise it has a tendency to show up in large chunks due to the rate limiting. This logic matches what we have in C# now.

PR #259, opened 2025-06-19
Add StreamableHttp support to MCP Client

heyitsaamir:
Looks like SSE was deprecated and StreamableHttp was introduced to MCP Client.

This change makes SSE optional. Tested using https://github.com/MicrosoftDocs/mcp?tab=readme-ov-file#-installation--getting-started (which fails using SSE, but works using Http).

image

We can deprecate SSE out soon, but for now I provided an escape hatch by allowing devs to pass in `transport: "sse"` as a parameter for a given URL.

skip-test-verification - mcp test is changed, but the CLI doesn't use "tool" so it's unaffected.

### DIFF
--- a/external/mcpclient/src/mcp-client-plugin.ts
+++ b/external/mcpclient/src/mcp-client-plugin.ts
@@ -11,9 +11,10 @@ import {
   McpClientPluginParams,
   McpClientPluginUseParams,
   McpClientToolDetails,
+  McpClientTransportType,
   ValueOrFactory,
 } from './mcp-client-types';
-import { buildSSEClientTransport } from './mcp-transport';
+import { buildSSEClientTransport, buildStreamableHttpClientTransport } from './mcp-transport';
 
 export class McpClientPlugin implements ChatPromptPlugin<'mcpClient', McpClientPluginUseParams> {
   readonly name = 'mcpClient';
@@ -88,6 +89,7 @@ export class McpClientPlugin implements ChatPromptPlugin<'mcpClient', McpClientP
     if (args.params?.availableTools && args.params.availableTools.length > 0) {
       this._cache[args.url] = {
         ...this._cache[args.url],
+        transport: args.params?.transport,
         // If the tools are being supplied, we assume they are up to date
         lastAttemptedFetch: Date.now(),
         availableTools: args.params.availableTools,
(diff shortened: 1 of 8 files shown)


external/mcpclient/src/mcp-client-plugin.ts
```diff
@@ -199,10 +201,11 @@ export class McpClientPlugin implements ChatPromptPlugin<'mcpClient', McpClientP
 
   private async fetchTools(
     url: string,
-    headers?: ValueOrFactory<Record<string, string>>,
-    skipIfUnavailable?: boolean
+    headers: ValueOrFactory<Record<string, string>> | undefined,
```
singhk97, line 204: what's the difference between "`| undefined`" and "`headers?`"?

singhk97 (dismissed):
LGTM

external/mcpclient/src/mcp-client-plugin.ts
aacebo: the need to explicitly set the property to undefined vs it being optional to set.

external/mcpclient/src/mcp-client-plugin.ts
```diff
@@ -226,11 +229,22 @@ export class McpClientPlugin implements ChatPromptPlugin<'mcpClient', McpClientP
...
-      ? this.createTransport(serverUrl)
-      : await buildSSEClientTransport(serverUrl, headers);
+    const buildTransport = () => {
+      if (this.createTransport) {
+        return this.createTransport(serverUrl);
+      }
+      switch (transportType) {
+        case 'sse':
+          return buildSSEClientTransport(serverUrl, headers);
+        case 'streamable-http':
```
aacebo, line 242: line 242 can be removed.

aacebo (dismissed):

external/mcpclient/src/mcp-client-plugin.ts
heyitsaamir: value optional vs key optional

external/mcpclient/src/mcp-client-plugin.ts
heyitsaamir: True, I kinda prefer this though, so it's clear that both an explicit streamable-http or default will create streamablehttp. Don't think it's a biggie, so i'm gonna leave this be. Hope that's okay!

PR #261, opened 2025-06-23
[App] Add send() method to app function handler props

Jesperholmbergmsft:
## What
This adds a `send` method to the app function handler props to let the function handler post an activity to the conversation where the tab app that invoked the function is active - just as we already have for the `app.on('message')` handler. Also exposes a `getCurrentConversationId()` method that the function handler can use if they need to conversation ID for other purposes - e.g. to call some graph API or what have you.

### Setup
I set up a simple react tab app that invokes a function like so:
```
export default function App() {
  const [app, setApp] = React.useState<client.App | null>(null);
  React.useEffect(() => {
    (async () => {
      const app = new client.App(clientId, {
        logger: new ConsoleLogger('@tests/tab', { level: 'debug' }),
      });

      await app.start();
      setApp(app);
    })();
  }, []);

  const callFunction = React.useCallback(async () => {
    await app?.exec<string>('send-function-test');
  }, [app]);

  return (
    <div className="App">
        <button disabled={!app} onClick={callFunction}>Click me</button>
    </div>
  );
}
```

On the bot side, the function handler is:
```
app.function('send-function-test', async ({send, getCurrentConversationId}) => {
  await send(`message sent to ${await getCurrentConversationId()}`);
});
```

### Tests
Installed the app and tested the feature in 
 - group chat
 - meeting chat
 - channel chat
 - channel meeting chat
 - private scope

### DIFF
--- a/packages/apps/src/contexts/client.ts
+++ b/packages/apps/src/contexts/client.ts
@@ -10,14 +10,19 @@ export interface IClientContext {
   readonly appSessionId: string;
 
   /**
-   * The Microsoft Entra tenant ID of the current user.
+   * The Microsoft Entra tenant ID of the current user, extracted from request auth  token.
    */
-  readonly tenantId?: string;
+  readonly tenantId: string;
 
   /**
-   * The Microsoft Entra object id of the current user.
+   * The Microsoft Entra object id of the current user, extracted from the request auth token.
    */
-  readonly userId?: string;
+  readonly userId: string;
+
+  /**
+   * The name of the current user, extracted from the request auth token.
+   */
+  readonly userName: string;
 
   /**
    * The Microsoft Teams ID for the team with which the content is associated.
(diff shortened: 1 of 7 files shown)


packages/apps/src/contexts/client.ts
```diff
@@ -10,14 +10,19 @@ export interface IClientContext {
   readonly appSessionId: string;
 
   /**
-   * The Microsoft Entra tenant ID of the current user.
+   * The Microsoft Entra tenant ID of the current user, extracted from request auth  token.
    */
-  readonly tenantId?: string;
+  readonly tenantId: string;
```
Jesperholmbergmsft, line 15: These don't need to be optional -- if we can't get these values from the bearer token, the request is rejected and the function handler is not invoked.

heyitsaamir (approved):
nice! it might be worth it to include tests like the one you have in the description in the tests folder.

Jesperholmbergmsft:
> nice! it might be worth it to include tests like the one you have in the description in the tests folder.

@heyitsaamir  - totally agree, I'll do that in a separate change

PR #263, opened 2025-06-23
Create auth package for bot token validation

heyitsaamir:
This package implements the authentication validation protocol for bot framework requests to your server as outlined in https://learn.microsoft.com/en-us/azure/bot-service/rest-api/bot-framework-rest-connector-authentication.

Tested in the branch above with real Teams requests, and added unit tests.


packages/apps/src/plugins/http/plugin.ts
```diff
@@ -217,15 +233,32 @@ export class HttpPlugin implements ISender {
...
+      if (!authorization) {
+        res.status(401).send('unauthorized no authorization header');
+        return;
+      }
+
+      try {
+        await this.botTokenValidator.validateToken(authorization, activity.serviceUrl || '');
+        token = new JsonWebToken(authorization);
+      } catch (error: any) {
+        if (error instanceof TokenAuthenticationError || error instanceof TokenClaimsError || error instanceof TokenFormatError || error instanceof TokenInfrastructureError || error instanceof TokenValidationError) {
```
heyitsaamir: Put this in the package? or we throw 401 always and never 500?

packages/apps/src/plugins/http/plugin.ts
```diff
@@ -217,15 +233,32 @@ export class HttpPlugin implements ISender {
...
-      : {
-          appId: '',
-          from: 'azure',
-          fromId: '',
-          serviceUrl: activity.serviceUrl || '',
-          isExpired: () => false,
-        };
+    let token: IToken;
+    if (this.botTokenValidator) {
+      if (!authorization) {
```
singhk97: this check is already implemented in line 230.

singhk97 (approved):
This should also be added to the .NET library.

singhk97:
How is this different from this: https://github.com/microsoft/teams.ts/blob/main/packages/apps/src/middleware/entra-token-validator.ts?

aacebo:
why was this implemented as a new package? Since this isn't a modular or optional piece of logic that could be used elsewhere why not keep it in the `apps` package where it will be used? Today our auth logic primarily lives in the `api` package so I could also see it being there.

packages/apps/src/plugins/http/plugin.ts
heyitsaamir: Technically only accurate if node_env !== "local". TS complains otherwise

packages/apps/src/middleware/entra-token-validator.ts
```diff
@@ -67,50 +70,46 @@ export class EntraTokenValidator {
...
-    return validatedToken;
+    // Return the token in the format expected by the existing interface
+    return verifyResult;
   }
 
   getTokenPayload(token: jwt.Jwt): JwtPayload | null {
     return token.payload instanceof Object ? token.payload : null;
   }
 
-  /**
```
heyitsaamir: Oops didn't mean to get rid of the comment here

Jesperholmbergmsft (dismissed):

packages/apps/src/middleware/cache-manager.ts
```diff
@@ -0,0 +1,43 @@
+interface ICacheEntry<T> {
+  data: T;
+  expiry: number;
+}
+
+export class CacheManager<T> {
```
Jesperholmbergmsft: We're already indirectly dependent on lru-cache -- maybe we can just use that directly and not have to build our own cache manager?

heyitsaamir: Oh do we? didn't know that

heyitsaamir: Ah, looks like we do, but it's an indirect dependency. Let's use this super simple cache for now. LRU cache looks kinda heavy for what we need

packages/apps/src/middleware/jwt-utils.ts
```diff
@@ -0,0 +1,104 @@
+import {
```
aacebo: this should probably either go in the `/utils` folder or a top level `jwt` or auth folder for cleaner organization. ideally the middleware folder should just be middleware to keep it consistent and easy to navigate. I would recommend putting everything but the middleware into a top level `auth` folder or something similar.

aacebo (commented):
@heyitsaamir just curious, have you looked into this dependency/implementation in agents SDK? https://github.com/microsoft/Agents-for-js/blob/main/packages/agents-hosting/src/auth/jwt-middleware.ts? Any insights into why we are doing this our way over how they approached it?

heyitsaamir:
> @heyitsaamir just curious, have you looked into this dependency/implementation in agents SDK? https://github.com/microsoft/Agents-for-js/blob/main/packages/agents-hosting/src/auth/jwt-middleware.ts? Any insights into why we are doing this our way over how they approached it?

@aacebo  updated it to reflect this approach. I wanted this class to also be used by entra-token middleware so it's ready for that, but i want jesper to validate it before i switch it over.

heyitsaamir:
> How is this different from this: https://github.com/microsoft/teams.ts/blob/main/packages/apps/src/middleware/entra-token-validator.ts?

Similar! And now with this class we can consolidate logic.

packages/apps/src/utils/jwt-validator.ts
```diff
@@ -0,0 +1,286 @@
+import jwt, { type JwtHeader, type JwtPayload, type SignCallback } from 'jsonwebtoken';
+import jwksRsa, { JwksClient, SigningKey } from 'jwks-rsa';
+
+import { ILogger } from '@microsoft/teams.common';
+
+import { asserts } from '.';
```
aacebo: should import from the file not the barrel

packages/apps/src/plugins/http/plugin.ts
```diff
@@ -217,15 +240,33 @@ export class HttpPlugin implements ISender {
...
-        };
+    let token: IToken;
+    if (this.serviceTokenValidator) {
+      if (!authorization) {
+        res.status(401).send('unauthorized no authorization header');
+        return;
+      }
+
+      // Use cached validator with per-request service URL validation
+      const validationResult = await this.serviceTokenValidator.validateAccessToken(authorization, activity.serviceUrl ? {
```
aacebo: I think we should stick to making this an optional configuration to avoid customers that don't want it not being able to disable it.

packages/apps/src/utils/jwt-validator.ts
```diff
@@ -0,0 +1,286 @@
...
+  } | {
+    /** For multi-tenant apps, restrict to specific tenant IDs */
+    allowedTenantIds?: string[];
+  };
+
+  /** Optional: Clock tolerance in seconds (default: 300) */
+  clockTolerance?: number;
+}
+
+export class JwtValidator {
```
aacebo: I think this should be implemented as an http middleware in the middleware folder instead, utils folder was made to house pure functional utility methods. Also if we make this a middleware it can easily with 1 line be added to any other plugin as well.

packages/apps/src/utils/jwt-validator.ts
```diff
@@ -0,0 +1,286 @@
...
+
+  constructor(options: IJwtValidationOptions, logger?: ILogger) {
+    this.options = options;
+    this.logger = logger;
+  }
+
+  /**
+   * Validates a JWT token using the configured options
+   */
+  validateAccessToken = async (
```
aacebo: nit: prefer function members over functions as variables

packages/apps/src/utils/jwt-validator.ts
```diff
@@ -0,0 +1,286 @@
...
+}
+
+export class JwtValidator {
+  public readonly options: IJwtValidationOptions;
+  private readonly logger?: ILogger;
+  private readonly jwksCache: Map<string, JwksClient> = new Map();
+
+  constructor(options: IJwtValidationOptions, logger?: ILogger) {
+    this.options = options;
+    this.logger = logger;
```
aacebo: if no logger is provided a default should be initialized

packages/apps/src/utils/jwt-validator.ts
```diff
@@ -0,0 +1,286 @@
...
+        }
+      });
+    });
+  };
+
+  private getJwksClient = () => {
+    switch (this.options.jwksUriOptions.type) {
+      case 'tenantId':
+        {
+          const cachedClient = this.jwksCache.get(`${this.options.tenantId}`);
```
aacebo: prefer throwing when null instead of unsafely coercing the tenantId that may be null

packages/apps/src/utils/jwt-validator.ts
```diff
@@ -0,0 +1,286 @@
...
+  };
+
+  /** Optional: Clock tolerance in seconds (default: 300) */
+  clockTolerance?: number;
+}
+
+export class JwtValidator {
+  public readonly options: IJwtValidationOptions;
+  private readonly logger?: ILogger;
+  private readonly jwksCache: Map<string, JwksClient> = new Map();
```
aacebo: why cache the clients? It seems easier to cache a `Map<string, SigningKey>` where the key is the header id. Also may want to consider an LFU or LRU cache combined with a `ttl`, so that you limit the amount of member that the cache can use and force evictions.

aacebo (changes requested):

packages/apps/src/utils/jwt-validator.ts
heyitsaamir: Stylistic decision? I have to bind it if it's a function, which is easy to forget.

packages/apps/src/utils/jwt-validator.ts
heyitsaamir: Can't be null, because `this.options.jwksUriOptions.type === 'tenantId'`. It's using type discriminator here.

packages/apps/src/utils/jwt-validator.ts
heyitsaamir: `JwksClient` itself is doing the caching for the key here. I'm just making it so that we store the JwksClient so it can maintain its cache internally. If we recreate this client, we can't cache the keys in memory.
`JwksClient` has a cache implementation internally, with a TTL etc.

packages/apps/src/utils/jwt-validator.ts
heyitsaamir: Sounds good

packages/apps/src/plugins/http/plugin.ts
heyitsaamir: sure, but why would a customer want to disable this?

heyitsaamir: I'll add a flag to disable it, but not sure what the use case might be

aacebo: for example, what if this request was coming from webchat or some other service other than teams? There is no guarantee all providers of the activity protocol will use this form of auth.

Also, in the case of things like devtools/testing or even "I have protocol A and ill send it as an activity payload to you", you would likely want to disable this functionality. At the very least its an escape hatch and I typically modularize these kinds of things to give devs the power of choice, although I could see us having this enabled by default.

packages/apps/src/utils/jwt-validator.ts
aacebo: ya its a nit so not blocking, but ideally I'd like our syntax to be consistent throughout the project unless we have a reason to diverge.

packages/apps/src/plugins/http/plugin.ts
heyitsaamir: Added `skip-auth` which optionally allows you to disable it. Also verified that the stuff works with web-chat, so i think this is safe to enable for all channels. But in case someone complains, they can pass in `skip-auth`.

packages/apps/src/utils/jwt-validator.ts
heyitsaamir: Updated to be middleware

packages/apps/src/utils/jwt-validator.ts
heyitsaamir: updated!

packages/apps/src/middleware/auth/jwt-validator.ts
```diff
@@ -0,0 +1,286 @@
...
+}
+
+export class JwtValidator {
+  public readonly options: IJwtValidationOptions;
+  private readonly logger?: ILogger;
+  private readonly jwksCache: Map<string, JwksClient> = new Map();
+
+  constructor(options: IJwtValidationOptions, logger?: ILogger) {
+    this.options = options;
+    this.logger = logger ?? new ConsoleLogger('jwt-validator');
```
aacebo: this isn't namespacing the logger properly.

packages/apps/src/middleware/jwt-validation-middleware.ts
```diff
@@ -0,0 +1,68 @@
...
+  credentials?: Credentials;
+  logger: ILogger;
+};
+
+export type JwtValidatedRequest = express.Request & {
+  validatedToken?: IToken;
+};
+
+export function withJwtValidation(params: JwtValidationParams) {
+  const { credentials, logger = new ConsoleLogger('jwt-validation-middleware') } = params;
```
aacebo: this isn't namespacing the logger properly

aacebo (approved):

packages/apps/src/middleware/auth/jwt-validator.ts
heyitsaamir: Ah good call. Fixed

packages/apps/src/middleware/jwt-validation-middleware.ts
heyitsaamir: Good call! fixed

PR #269, opened 2025-07-07
Fix oauth overriding

heyitsaamir:
Without this, we were ALWAYS forcing users to use `graph` as the oauth name.


singhk97 (approved):
LGTM!

PR #273, opened 2025-07-15
Added support for new shared channel events

Stevenic:
- Updated activity schema with new data structures and fields.
- Added new channel events that can be subscribed to.


packages/api/src/models/membership-source.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+   * The group ID of the team associated with this membership source.
+   */
+  teamGroupId: string
+
+  /**
+   * Optional. The tenant ID for the user.
+   */
+  tenantId?: string
```
arocc: Hi @Stevenic, can you make this required please? We'll always send it for membership sources

arocc: (just FYI, same request for Agents SDK: https://github.com/microsoft/Agents-for-js/pull/438/files#r2216912093)

aacebo (approved):

PR #275, opened 2025-07-18
Separate user token fetching and bot token fetching. And a minor refactor

heyitsaamir:
We normally refresh user tokens every time we're processing the activity. But user and app tokens were coupled here. So if user-token fetching fails, so does graph token fetching. This change decouples them so that we can still fetch app token even if user token fetching fails.

#### PR Dependency Tree

* **PR #275** 👈

### DIFF
--- a/packages/apps/src/app.ts
+++ b/packages/apps/src/app.ts
@@ -518,7 +518,7 @@ export class App<TPlugin extends IPlugin = IPlugin> {
   protected async getUserToken(
     channelId: ChannelID,
     userId: string
-  ): Promise<string | undefined> {
+  ) {
     const res = await this.api.users.token.get({
       channelId,
       userId,


packages/apps/src/app.ts
```diff
@@ -513,4 +514,33 @@ export class App<TPlugin extends IPlugin = IPlugin> {
     const graphResponse = await this.api.bots.token.getGraph(this.credentials);
     this._tokens.graph = new JsonWebToken(graphResponse.access_token);
   }
+
+  protected async getUserToken(
+    channelId: ChannelID,
+    userId: string
+  ): Promise<string | undefined> {
```
aacebo: nit: why does one method have an explicit return type but not the other?

aacebo (approved):

packages/apps/src/app.ts
heyitsaamir: good point. let me fix

heyitsaamir: removed to match others

PR #280, opened 2025-07-27
Add next to default oauth handlers

heyitsaamir:
Without these, we don't propagate the activity to any other handlers

### DIFF
--- a/packages/apps/src/app.oauth.ts
+++ b/packages/apps/src/app.oauth.ts
@@ -16,50 +16,50 @@ export async function onTokenExchange<TPlugin extends IPlugin>(
   ctx: contexts.IActivityContext<ISignInTokenExchangeInvokeActivity>
 ) {
   const { api, activity, log, next } = ctx;
+  if (this.oauth.defaultConnectionName !== activity.value.connectionName) {
+    log.warn(
+      `default connection name "${this.oauth.defaultConnectionName}" does not match activity connection name "${activity.value.connectionName}"`
+    );
+  }
+
   try {
-    if (this.oauth.defaultConnectionName !== activity.value.connectionName) {
-      log.warn(
-        `default connection name "${this.oauth.defaultConnectionName}" does not match activity connection name "${activity.value.connectionName}"`
-      );
+    const token = await api.users.token.exchange({
+      channelId: activity.channelId,
+      userId: activity.from.id,
+      connectionName: activity.value.connectionName,
+      exchangeRequest: {
+        token: activity.value.token,
+      },
+    });
+
+    ctx.userGraph = new graph.Client(
+      this.client.clone({
+        token: token.token,
+      })
+    );
+
+    this.events.emit('signin', { ...ctx, token, isSignedIn: true });
+    const nextResult = await next();
+    if (nextResult && nextResult.status !== 200) {
+      log.warn('Expecting token exchange to return 200, but got:', nextResult.status);
     }
-
-    try {
-      const token = await api.users.token.exchange({
-        channelId: activity.channelId,
-        userId: activity.from.id,
-        connectionName: activity.value.connectionName,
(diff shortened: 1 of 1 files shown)


heyitsaamir:
> is the use case that you want to implement custom handlers for the auth invoke activities? since this would only be an issue with verify state or exchange token im wondering what handlers you would need to come after them? I do think we need to have a way to disable the default auth invoke handlers to allow for custom implementations

Ya just for these. 

> since this would only be an issue with verify state or exchange token im wondering what handlers you would need to come after them

It's just unexpected that my handlers just don't work if i have them in app.on('signin.verifystate') etc. I could just wanna log it? Isn't it just good practice to propagate it? We can do it on success instead of always if that's the concern?

singhk97:
> > is the use case that you want to implement custom handlers for the auth invoke activities? since this would only be an issue with verify state or exchange token im wondering what handlers you would need to come after them? I do think we need to have a way to disable the default auth invoke handlers to allow for custom implementations
> 
> Ya just for these.
> 
> > since this would only be an issue with verify state or exchange token im wondering what handlers you would need to come after them
> 
> It's just unexpected that my handlers just don't work if i have them in app.on('signin.verifystate') etc. I could just wanna log it? Isn't it just good practice to propagate it? We can do it on success instead of always if that's the concern?

I believe the convention is that when there's an error in the middleware it should terminate processing. Middlewares can determine when and if the next handler should be called. 

I agree that it's confusing if the user defined `singin.verifyState` handler is not triggered in a non-error scenario. It's worth adding a comment in the method definiton labelling the auth handlers as having default implementation. And these handlers could also have an "override" param if that makes sense w/ the existing architecture.

heyitsaamir:
@aacebo  @singhk97 updated, take a look :)

aacebo:
> > is the use case that you want to implement custom handlers for the auth invoke activities? since this would only be an issue with verify state or exchange token im wondering what handlers you would need to come after them? I do think we need to have a way to disable the default auth invoke handlers to allow for custom implementations
> 
> Ya just for these.
> 
> > since this would only be an issue with verify state or exchange token im wondering what handlers you would need to come after them
> 
> It's just unexpected that my handlers just don't work if i have them in app.on('signin.verifystate') etc. I could just wanna log it? Isn't it just good practice to propagate it? We can do it on success instead of always if that's the concern?

oh I see, on that case I would personally expect that if you do `app.on('signin.verify-state', ...)` our default handler implementation should be replaced entirely, I think we should add an optional `replace = false` to the `app.on` method that when true replaces all existing routes of that selector with the provided new route, this would allow devs to override our defaults safely.

heyitsaamir:
> > > is the use case that you want to implement custom handlers for the auth invoke activities? since this would only be an issue with verify state or exchange token im wondering what handlers you would need to come after them? I do think we need to have a way to disable the default auth invoke handlers to allow for custom implementations
> > 
> > 
> > Ya just for these.
> > > since this would only be an issue with verify state or exchange token im wondering what handlers you would need to come after them
> > 
> > 
> > It's just unexpected that my handlers just don't work if i have them in app.on('signin.verifystate') etc. I could just wanna log it? Isn't it just good practice to propagate it? We can do it on success instead of always if that's the concern?
> 
> oh I see, on that case I would personally expect that if you do `app.on('signin.verify-state', ...)` our default handler implementation should be replaced entirely, I think we should add an optional `replace = false` to the `app.on` method that when true replaces all existing routes of that selector with the provided new route, this would allow devs to override our defaults safely.

Ah that's an interesting idea. Though, then the devs would need to be aware that those routes _require_ replacing right? Or are you saying for some routes (like the ones we provide default handlers for), we always replace the default handler? We could always flag "isInternal" for a given handler if we write it internally, and if the application overrides it, we replace it?

heyitsaamir:
Don't need this after we added #317

PR #281, opened 2025-07-27
Send group Oauth requests to personal scope

heyitsaamir:
When an oauth request comes, we create a 1:1 chat to send it to them. But we weren't actually sending the message there.

This was accidentally fine because we added support for Group SSO.

But i'm not sure if that's GA'd so creating this PR to not forget and check with the team. If it is released, we can remove the code to create that 1:1 chat.


aacebo (approved):

singhk97 (approved):
Good catch - afaik the current state of group sso won't work in channel scope so this code would most likely stay in one form of the other for the forseeable future.

PR #285, opened 2025-08-01
Fixed streaming bug in stream.ts

teddyam:
Added locking to flush function in stream.ts. Prevents multiple stream activities from being created if flush requires a retry.

### DIFF
--- a/packages/apps/src/plugins/http/stream.ts
+++ b/packages/apps/src/plugins/http/stream.ts
@@ -31,6 +31,7 @@ export class HttpStream implements IStreamer {
   private _result?: SentActivity;
   private _timeout?: NodeJS.Timeout;
   private _logger: ILogger;
+  private _flushing: boolean = false;
 
   constructor(client: Client, ref: ConversationReference, logger?: ILogger) {
     this.client = client;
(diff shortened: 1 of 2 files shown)


packages/apps/src/plugins/http/stream.ts
```diff
@@ -95,50 +96,58 @@ export class HttpStream implements IStreamer {
...
+        .withText(this.text)
+        .addStreamUpdate(this.index + 1);
 
-    const res = await promises.retry(() => this.send(activity), {
-      logger: this._logger
-    });
+      res = await promises.retry(() => this.send(activity), {
+        logger: this._logger
+      });
+    } finally {
```
heyitsaamir: Why finally here and not below?

teddyam: The line causing multiple activities to get created is 145, retrying the send. I chose to unlock right after this bc the code after can run async (afaik). Can move the finally to the end but it'd lock the function longer than necessary imo

heyitsaamir: probably okay for JS since it's single threaded. and all the code after is synchronous. I would say it's clearer if it's wrapping the whole thing though because it clearly shows intent on what you're trying to do.

> but it'd lock the function longer than necessary imo

JS is single threaded, so when the execution isn't suspended, it's only executing in 1 single thread. so effectively, the other threads aren't really working anyway, so your "lock" isn't doing anything after the await's done anyway.

teddyam: oic, will move to the end for clarity

teddyam:
> Btw do you have evidence that it was trying to flush mulitple times and that was causing the issue? Did you add a print line or something to verify

only verified by tracing the logic in code. can get some concrete evidence rn

teddyam:
`if (activity.entities?.[0].streamSequence === 1) {
      this._logger.debug('creating new stream sequence');
    }`

by adding this at the end of `send`, I was able to see that two streaming sequences could begin for one single message. this shows the race condition of multiple calls entering flush at one time creating multiple activties.
image

heyitsaamir (approved):

PR #288, opened 2025-08-04
Add pluggable activity context

heyitsaamir:
This PR introduces the ability for plugins to add _typed_ values to the activity context. It does this by adding a dedicated `buildActivityContext` function that gets called in `app.process`. Then via a bunch of generics plumbing we merge the plugin's context values with the default ones.

We can use this to separate graph out into other packages, but keep the contextual objects for the activity-handlers unchanged (aka. `appGraph`, and `userGraph` will still be available). However, `graph` object on `app` would most likely be part of the the `GraphPlugin` - (maybe we make that accessible via a new `app.getPlugin('graph')` method?)

Here's a mini demo:

https://github.com/user-attachments/assets/328517e4-99b2-459d-ad05-80291c7b7e99

#### PR Dependency Tree

* **PR #288** 👈

### DIFF
--- a/packages/apps/src/types/app-routing.ts
+++ b/packages/apps/src/types/app-routing.ts
@@ -1,9 +1,17 @@
 import { IPlugin } from './plugin';
+import { UnionToIntersection } from './union-to-intersection';
+
+import type { App } from '../app';
+import type { IRoutes } from '../routes';
 
 /**
  * Extracts the events from a plugin if it extends PluginWithEvents
  */
-export type PluginAdditionalContext<T> = T extends IPlugin<infer Ctx, any>
-  ? Ctx
-  : {};
+export type PluginAdditionalContext<T> = UnionToIntersection<
+  T extends IPlugin<infer Ctx, any> ? Ctx : {}
+> extends infer R ? R extends Record<string, any> ? R : {} : {};
 
+
+type AppPlugin<TApp extends App> = TApp extends App<infer TPlugin> ? TPlugin : never;
+
+export type AppRoutingHandler<Name extends keyof IRoutes, TApp extends App<any>> = Exclude<IRoutes<PluginAdditionalContext<AppPlugin<TApp>>>[Name], undefined>;
--- a/packages/apps/src/app.routing.ts
+++ b/packages/apps/src/app.routing.ts
@@ -6,6 +6,10 @@ import { IRoutes } from './routes';
 import { IPlugin, RouteHandler } from './types';
 import { PluginAdditionalContext } from './types/app-routing';
 
+type AppPlugin<TApp extends App> = TApp extends App<infer TPlugin> ? TPlugin : never;
+
+export type AppRoutingHandler<Name extends keyof IRoutes, TApp extends App<any>> = Exclude<IRoutes<PluginAdditionalContext<AppPlugin<TApp>>>[Name], undefined>;
+
 /**
  * subscribe to an event
  * @param name event to subscribe to


packages/apps/src/types/plugin/plugin.ts
```diff
@@ -27,7 +27,7 @@ export type OnErrorPluginEvent = (event: IErrorEvent) => void;
  */
 export type OnActivityPluginEvent = (event: IActivityEvent) => void;
 
-export interface IPlugin<TEvents extends {} = {}> {
+export interface IPlugin<TCtxAddons extends {} = {}, TEvents extends {} = {}> {
```
aacebo: suggestion: might be clearer/simpler to just call it `TContext`

packages/apps/src/types/plugin/plugin.ts
```diff
@@ -66,6 +66,8 @@ export interface IPlugin<TEvents extends {} = {}> {
    */
   onActivity?(event: IPluginActivityEvent): void | Promise<void>;
 
+  buildActivityContext?(activity: Activity): TCtxAddons;
```
aacebo: would prefer if we kept to the `onXYZ` naming convention to keep things consistent, but we can also maybe reuse the `onActivity` for this and change the return type to `void | TContext | Promise<void | TContext>`

packages/apps/src/types/app-routing.ts
```diff
@@ -0,0 +1,9 @@
+import { IPlugin } from './plugin';
+
+/**
+ * Extracts the events from a plugin if it extends PluginWithEvents
+ */
+export type PluginAdditionalCtx<T> = T extends IPlugin<infer Ctx, any>
```
aacebo: nit: prefer `Context` over `Ctx` since it is what we have used throughout the project thus far.

packages/apps/src/routes/index.ts
```diff
@@ -9,22 +11,23 @@ import { InvokeActivityRoutes } from './invoke';
...
-    InstallActivityRoutes,
-    ConversationUpdateActivityRoutes,
-    MessageUpdateActivityRoutes,
-    MessageDeleteActivityRoutes,
-    EventActivityRoutes {
-  mention?: RouteHandler<IActivityContext>;
-  activity?: RouteHandler<IActivityContext>;
+export interface IRoutes<TExtraCtx extends Record<string, any> = Record<string, any>>
+  extends ActivityRoutes<TExtraCtx>,
+  InvokeActivityRoutes<TExtraCtx>,
```
aacebo, line 16: this is why I had originally avoided extra generics on the context, not against it but would want to see what impacts this has to creating a handler function where type inference isn't an option, like when your creating handler functions in separate files and then registering them on the app in your index.ts.

Example:
```typescript
// index.ts
import * as handlers from './handlers';

const app = new App(...);
app.on('message', handlers.onMessageActivity);
```

```typescript
// handlers/index.ts
export async function onMessageActivity({ message, send }: IActivityContext<MessageActivity>) {
  send('hi!');
}
```

packages/apps/src/contexts/activity.ts
```diff
@@ -84,10 +84,12 @@ export interface IActivityContextOptions<T extends Activity = Activity> {
    * call the next event/middleware handler
    */
   next: (
-    context?: IActivityContext
+    context?: IActivityContext & TExtraCtx
```
aacebo, line 92: It would be interesting to know, if two plugins are imported that have 1 or more overlapping context additions:

case 1: the types do not match
- does it union them or give a type error?
- does it not allow two plugins to be imported that have overlapping context extra types?
- how do you decide which plugin gets to set the overlapping property?

the overlapping property issue could be resolved by namespacing the context extra keys, either by prefixing the keys with the plugin name, or by putting the keys in an object of the plugins name.

aacebo (commented):
left questions, this is a good first draft let me know when its ready for another review!

packages/apps/src/routes/index.ts
heyitsaamir: Good point - let me test this scenario out. I think IActivity with no generics passed in should still work but let me be sure

packages/apps/src/contexts/activity.ts
heyitsaamir: Good question - 

image


TS makes it a union, but in practice, i just log.warn to say the key is being overridden. Might be good enough for a first pass?

packages/apps/src/routes/index.ts
heyitsaamir: image

image


Seems happy!

packages/apps/src/types/plugin/plugin.ts
heyitsaamir: Good idea. Updated it ^ lmk what you think

heyitsaamir:
> left questions, this is a good first draft let me know when its ready for another review!

Updated it with your suggestions @aacebo . Take a look!

packages/apps/src/contexts/activity.ts
aacebo: I think we might want to use an intersection instead of a union in this case, so that it gives a type error when this kind of overlap occurs, since there is a high probability of the plugin not working correctly if this happens. The alternative is to namespace in some way so that its not possible for plugins to have context overlap at all.

packages/apps/src/routes/index.ts
aacebo: but in this case you are not getting the `foo` plugin context right?

heyitsaamir: No i'm not, which is expected i think right? If i want the `foo` plugin context, i'd have to explicitly type `IActivityContext<Activity, typeof FooPugin>`. That's the tradeoff i guess?
Or the other way could be we provide a type IActivityContextHandler<typeof app>, and this gives you something like:
```ts
const handler: IActivityContextHandler<'message', typeof app> = (ctx) => { ctx.foo }
```

packages/apps/src/contexts/activity.ts
heyitsaamir: Fair. I updated it to use intersection, where will become "never"

https://github.com/user-attachments/assets/3255c76d-cf17-490a-a4d5-e92d3b83bb65

Maybe a good stepping stone for now, and we basically encourage plugins to namespace their keys to their plugins, and we can make certain exceptions (like "userGraph"). Thoughts?

packages/apps/src/routes/index.ts
heyitsaamir: Added a helper

https://github.com/user-attachments/assets/16f7103a-9000-44dd-93e9-cb67cb4c3338

packages/apps/src/contexts/activity.ts
Jesperholmbergmsft: The video doesn't load for me, but this puts a type gard to ensure that they types don't overlap? Ideally we'd prevent overlap altogether, even when the types are compatible. If a developer finds they need to use two plugins that give overlapping contexts, they can wrap one of them in a custom plugin that simply rewrites the returned type and avoid overlap that way.

heyitsaamir: @Jesperholmbergmsft 
image

It just flips it to "never". I'm trying to not over-engineer this unless necessary. If we just say "convention is to do <plugin_name.property_name> for properties", that might be good enough for plugins which are mainly built by us anyway.

Jesperholmbergmsft: Typing as `never` seems fine to me. Different plugins shouldn't have to know about each other to avoid naming conflicts, but the app developer can always handle it if they need to mix'n'match conflicting plugins.

aacebo (approved):

joshkaplinsky:
Does this behavior still work? I'm unable to pass the additional context from `onActivity` in a plugin down to the activity handler via `context` parameter (missing the typing as well). Appears that potentially the value isn't passed due to the call to `.toInterface()` here when the `ctx` is not passed through the `next` middleware chain: https://github.com/microsoft/teams.ts/blob/a8986a4753ae427c4c861383cb887c8516f2d35a/packages/apps/src/app.process.ts#L97

PR #289, opened 2025-08-04
Graph client endpoint separation

Jesperholmbergmsft:
This PR does the following:
a) Adds two new packages, graph-endpoints and graph-endpoints-beta to hold small, standalone functions that can build the endpoint-specific requests and provide strong typing for both inputs and outputs. The packages are built to allow for aggressive tree shaking. N.b. I'm only including a small set of endpoints in this PR, enough to illustrate the results without bloating the PR crazy.
b) Adds a common set of scripts to auto-generate those endpoint functions.
c) Adds a common script to auto-generate filtered-down types for the endpoints.
d) Adds a new Call() method to the graph.Client class. The method takes endpoint functions and arguments to construct and issue the appropriate request and return results.

The calling pattern changes from:
```
const graphClient = app.graph;
const presence = await graphClient.me.presence.get();

await graphClient.me.presence.update({ availability: 'Away', activity: 'Away' });
```

To:
```

const graphClient = app.graph;
const presence = await graphClient.call(endpoints.me.presence.get);

await graphClient.call(endpoints.me.presence.update, { availability: 'Away', activity: 'Away' });
```

### DIFF
--- a/packages/graph-tools/package.json
+++ b/packages/graph-tools/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@microsoft/teams.graph-tools",
   "version": "2.0.0-preview.9",
+  "private": true,
   "license": "MIT",
   "description": "Tools for generating TypeScript Graph API endpoints from OpenAPI specifications",
   "main": "./dist/index.js",


packages/graph-endpoints-beta/package.json
```diff
@@ -0,0 +1,143 @@
...
+    "gen:endpoints": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx ts-node ../../scripts/graph/endpoints.ts ../graph/openapi-beta.yaml ./src beta"
+  },
+  "dependencies": {
+  },
+  "devDependencies": {
+    "rimraf": "^6.0.1",
+    "tsup": "^8.4.0",
+    "typescript": "^5.4.5"
+  },
+  "exports": {
```
Jesperholmbergmsft, line 59: All these exports are to enable intellisense, should the developer import from a deep path like
`import { get as getApprovers } from '@microsoft/teams.graph-endpoints/privacy/subjectRightsRequests/approvers';`
There's no tooling support for glob pattern here, but they do support wildcard for each level - and both v1.0 and beta go ten levels deep.

packages/graph-endpoints-beta/package.json
```diff
@@ -0,0 +1,143 @@
...
+    "dist",
+    "README.md"
+  ],
+  "engines": {
+    "node": ">=20"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/microsoft/teams.ts.git",
+    "directory": "packages/graph"
```
Jesperholmbergmsft: auto-nit - fix this for both packages

packages/graph-endpoints-beta/src/common.ts
```diff
@@ -0,0 +1,105 @@
+// DO NOT EDIT THIS FILE
+import * as http from '@microsoft/teams.common/http';
+
+import type { paths } from './types.ts';
+
+export type EndpointRequest<TResponse> = {
```
Jesperholmbergmsft: This is the return type for all the endpoint functions, giving enough details to issue the request and return a typed response. 'ver' will be used to figure out the right baseUrl - v1.0 vs beta.

packages/graph-endpoints-beta/src/index.ts
```diff
@@ -0,0 +1 @@
+export * as privacy from './privacy';
```
Jesperholmbergmsft, line 1: I'm just generating endpoints for the smallest top-level folder in this PR, enough to illustrate the results. After removing the endpoint wildcard in the code generator, this will grow dramatically.

packages/graph-endpoints-beta/tsup.config.js
```diff
@@ -0,0 +1,23 @@
...
+/** @type {import('tsup').Options} */
+module.exports = {
+  ...require('@microsoft/teams.config/tsup.config'),
+  minify: true,
+  bundle: false, // Keep bundle false to preserve directory structure
+  sourcemap: false,
+  target: 'node18',
+  splitting: true,
+  treeshake: true,
+  dts: false, // .d.ts file generation is done with tsc --emitDeclarationOnly
```
Jesperholmbergmsft, line 13: tsup isn't able to generate types for the whole entire graph. It takes a half hour and then it fails with out of memory exception. tsc has no problem though, it completes in just a moment.

packages/graph-endpoints-beta/tsup.config.js
```diff
@@ -0,0 +1,23 @@
+const pkg = require('./package.json');
+
+
+/** @type {import('tsup').Options} */
+module.exports = {
+  ...require('@microsoft/teams.config/tsup.config'),
+  minify: true,
+  bundle: false, // Keep bundle false to preserve directory structure
```
Jesperholmbergmsft, line 8: bundle+preserveModules is key to make the barrel files tree-shakable. this leaves the full folder structure in the dist folder, and apps using ES Modules + a modern bundler get good tree shaking even when importing the root barrel file.

scripts/graph/allow-list.ts
```diff
@@ -0,0 +1,2 @@
+// the smallest API surface of the whole graph
```
Jesperholmbergmsft: This will go away when we want to support the whole graph.

scripts/graph/client.ts.hbs
```diff
@@ -0,0 +1,95 @@
+
```
Jesperholmbergmsft: Based on /graph/scripts/client.ts.template, but updated to generate the new format you see under the graph-endpoints/src folder

scripts/graph/endpoints.ts
```diff
@@ -0,0 +1,330 @@
+import camelcase from 'camelcase';
```
Jesperholmbergmsft: Based on the /graph/scripts/endpoints.ts script, but updated to support the new format and to take arguments for inputs & outputs. I put it here as a step towards adding a CLI command to invoke this too, to let developers generate their own custom-filtered set of endpoints.

scripts/graph/types.ts
```diff
@@ -0,0 +1,573 @@
+import fs from 'fs';
```
Jesperholmbergmsft: This script generates the type definitions for the endpoints. The existing /graph/scripts/types.ts script reads the openapi.yaml file; parses it into a typescript AST, and saves to a file. This one does the same, but filters out paths, operations and schema elements that aren't used based on the allowlist; and removes some type crud that isn't actually exposed to developers and only serve to bloat the file. The result is a much smaller types.ts file than we have presently, which helps reduce the NPM package size as well as the burden for intellisense.

scripts/utils/fix-esm-import-paths.ts
```diff
@@ -0,0 +1,58 @@
+// This script rewrites extensionless relative imports in .mjs files to include .mjs
```
Jesperholmbergmsft: The graph-endpoints and graph-endpoints-beta opt out of bundling and preserve the output structure, and this is key to allowing tree-shaking even when importing from barrel files. However, it means I have to go and update the ESM import statements in the .mjs files. By default, the built files contain import statements like `import*as subjectRightsRequests from'./subjectRightsRequests';`, but ESM doesn't allow importing files without a file extension. So this script is hooked up to the graph packages' build scripts to rewrite into `import*as subjectRightsRequests from'./subjectRightsRequests/index.mjs'`

More details:
ESM file extension import -
https://nodejs.org/api/esm.html#import-specifiers
There are three types of specifiers:
Relative specifiers like './startup.js' or '../config.mjs'. They refer to a path relative to the location of the importing file. The file extension is always necessary for these.

https://nodejs.org/api/esm.html#mandatory-file-extensions
Mandatory file extensions#
A file extension must be provided when using the import keyword to resolve relative or absolute specifiers. Directory indexes (e.g. './startup/index.js') must also be fully specified.
This behavior matches how import behaves in browser environments, assuming a typically configured server.

packages/graph-endpoints-beta/package.json
```diff
@@ -0,0 +1,143 @@
...
+    "clean": "npx rimraf ./dist",
+    "build:esm": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx tsup --format esm",
+    "build:esm:no-clean": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx tsup --format esm --no-clean",
+    "build:cjs": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx tsup --format cjs",
+    "build:cjs:no-clean": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx tsup --format cjs --no-clean",
+    "build:types": "npx tsc --emitDeclarationOnly",
+    "build": "npm run clean && npm run build:esm:no-clean && npm run build:cjs:no-clean && npm run build:types && npm run fix:esm-imports && npx rimraf ./dist/index.d.mts",
+    "fix:esm-imports": "node ../../scripts/utils/fix-esm-import-paths.js",
+    "gen": "npm run gen:types && npm run gen:endpoints",
+    "gen:types": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx ts-node ../../scripts/graph/types.ts ../graph/openapi-beta.yaml ./src beta",
```
Jesperholmbergmsft: The script here is running against ../graph/openapi-beta.yaml, but I'm not checking that in. it's too big, it doesn't belong in the repo. Instead, in an upcoming change I'll make the generate script find & fetch the latest on demand and dump it somewhere locally.

scripts/graph/client.ts.hbs
aacebo: I know this is pushing it lol, but leadership has been asking for us to document this stuff and I am wondering how hard it would be to also generate simple markdown docs for each controller/endpoint we support that get generated as the code is generated.

Jesperholmbergmsft: The yaml doesn't have enough content that we can rival the official graph documentation (e.g. the yaml doesn't include permissions requirements), but I we can absolutely auto-document which endpoints we support and how to call them. I think this would be worthwhile to do, for sure.
I actually experimented with it earlier this summer (https://github.com/microsoft/teams-ai/pull/2498/discussion), and realized the hardest part is making the content readable at all!

packages/graph-endpoints-beta/tsup.config.js
heyitsaamir: We shouldn't really be using tsup anyway, tsup is more for client code

packages/graph-endpoints-beta/src/common.ts
```diff
@@ -0,0 +1,105 @@
...
+};
+
+/**
+ * @license (MIT OR CC0-1.0)
+ * @source https://github.com/sindresorhus/type-fest/blob/570e27f8fdaee37ef5d5e0fbf241e0212ff8fc1a/source/simplify.d.ts
+ */
+export type Simplify<T> = { [KeyType in keyof T]: T[KeyType] } & {};
+
+// https://stackoverflow.com/a/50375286/206879
+export type UnionToIntersection<U> = (U extends any ? (k: U) => void : never) extends (
```
heyitsaamir: we have this in a few places :P

packages/graph-endpoints-beta/src/common.ts
```diff
@@ -0,0 +1,105 @@
+// DO NOT EDIT THIS FILE
```
heyitsaamir: NIT: feels like this should be in a types folder

packages/graph/scripts/client.ts.template
```diff
@@ -1,4 +1,7 @@
-{{#if (notEmpty endpoints)}}
+{{#if (eq url "/")}}
```
heyitsaamir, line 1: NIT: should we use handlebars?

heyitsaamir (approved):
I think this makes sense and looks good. Let's add a few tests in `graph` tests.

packages/graph-endpoints-beta/src/common.ts
Jesperholmbergmsft: That's a good idea, I'll make it so.

packages/graph-endpoints-beta/src/common.ts
Jesperholmbergmsft: By the time this effort is done, we'll maintain it in only two places - one in packages\apps\src\types\union-to-intersection.ts that we're using within the library, and then one in scripts\graph\common.ts.hbs used to auto-generate the endpoints. It'd be nice to have just one copy, but two is probably ok as it allows the endpoint code to be generated with no external dependencies at all, making it easier for folks to auto-generate a custom set of endpoints with the CLI and integrate those in their existing code base.

packages/graph/scripts/client.ts.template
Jesperholmbergmsft: This file will go away when the effort is complete and we no longer auto-generate endpoints in the graph package. I just wanted to add the call() function here temporarily though, so it doesn't get lost in case someone updates the endpoints right quick before the whole switch is done.

packages/graph/jest.config.js
```diff
@@ -1 +1,16 @@
-module.exports = require('@microsoft/teams.config/jest.config');
+const baseConfig = require('@microsoft/teams.config/jest.config');
+
+module.exports = {
+  ...baseConfig,
+  collectCoverageFrom: [
+    'src/index.ts', // Only collect coverage for this file
```
Jesperholmbergmsft, line 6: src/index.ts is the only file with unit tests presently, to test the new call() method. There's no reason to collect code coverage on all the untested auto-generated endpoints, as doing so takes a minute or two for no benefit.

Jesperholmbergmsft:
> I think this makes sense and looks good. Let's add a few tests in `graph` tests.

I added tests for the client.call() method, as that's the only thing that'll stay once the new model is in place.

packages/graph/src/index.ts
```diff
@@ -54,6 +56,59 @@ export class Client {
...
+   *
+   * // With HTTP configuration
+   * const user = await client.call(endpoints.users.get, 'user-id', {
+   *   requestConfig: {
+   *     timeout: 5000
+   *   }
+   * });
+   * ```
+   */
+  async call<
```
aacebo, line 79: since this is a breaking change, will merging this break the apps package since it relies on this client? Will merging this break our samples/cli templates that use the existing graph client?

package.json
```diff
@@ -43,6 +43,7 @@
     "@changesets/cli": "latest",
     "@turbo/gen": "^2.4.4",
     "bluehawk": "^1.6.0",
+    "prettier": "^3.2.5",
```
aacebo: scripts should have their own independent package.json and turbo.json files and be private packages that can be utilized as part of the workspace, this way they can be added as dev dependencies and invoked via `npx` instead of `../../` and isolates dependencies to where they are actually used.

packages/graph-endpoints/package.json
```diff
@@ -0,0 +1,143 @@
...
+    "clean": "npx rimraf ./dist",
+    "build:esm": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx tsup --format esm",
+    "build:esm:no-clean": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx tsup --format esm --no-clean",
+    "build:cjs": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx tsup --format cjs",
+    "build:cjs:no-clean": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx tsup --format cjs --no-clean",
+    "build:types": "npx tsc --emitDeclarationOnly",
+    "build": "npm run clean && npm run build:esm:no-clean && npm run build:cjs:no-clean && npm run build:types && npm run fix:esm-imports && npx rimraf ./dist/index.d.mts",
+    "fix:esm-imports": "node ../../scripts/utils/fix-esm-import-paths.js",
+    "gen": "npm run gen:types && npm run gen:endpoints",
+    "gen:types": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx ts-node ../../scripts/graph/types.ts ../graph/openapi.yaml ./src v1.0",
```
aacebo: see comment about scripts being their own private packages

packages/graph-endpoints-beta/package.json
```diff
@@ -0,0 +1,143 @@
...
+  ],
+  "scripts": {
+    "clean": "npx rimraf ./dist",
+    "build:esm": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx tsup --format esm",
+    "build:esm:no-clean": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx tsup --format esm --no-clean",
+    "build:cjs": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx tsup --format cjs",
+    "build:cjs:no-clean": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx tsup --format cjs --no-clean",
+    "build:types": "npx tsc --emitDeclarationOnly",
+    "build": "npm run clean && npm run build:esm:no-clean && npm run build:cjs:no-clean && npm run build:types && npm run fix:esm-imports && npx rimraf ./dist/index.d.mts",
+    "fix:esm-imports": "node ../../scripts/utils/fix-esm-import-paths.js",
```
aacebo: see comment about scripts being their own private packages

scripts/utils/fix-esm-import-paths.js
```diff
@@ -0,0 +1,68 @@
+// This script rewrites extensionless relative imports in .mjs files to include .mjs
+
+const fs = require("fs");
+const path = require("path");
+
+// Get the directory where this script is being run from (should be the package directory)
+const distFolder = process.argv[2] || path.join(process.cwd(), "dist");
```
aacebo: why make this js instead of ts?

packages/graph/src/index.ts
Jesperholmbergmsft: This change is only additive for the moment so it won't break the samples & templates. I'll make sure to update those before making any breaking changes.

packages/graph-endpoints-beta/package.json
```diff
@@ -0,0 +1,146 @@
...
+  ],
+  "scripts": {
+    "clean": "npx rimraf ./dist",
+    "build:esm": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx tsup --format esm",
+    "build:esm:no-clean": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx tsup --format esm --no-clean",
+    "build:cjs": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx tsup --format cjs",
+    "build:cjs:no-clean": "npx cross-env NODE_OPTIONS='--max-old-space-size=16384' npx tsup --format cjs --no-clean",
+    "build:types": "npx tsc --emitDeclarationOnly",
+    "build": "npm run clean && npm run build:esm:no-clean && npm run build:cjs:no-clean && npm run build:types && npm run fix:esm-imports && npx rimraf ./dist/index.d.mts",
+    "fix:esm-imports": "node ../graph-tools/dist/fix-esm-import-paths.js ./dist",
```
Jesperholmbergmsft, line 48: Ideally this would use `npx @microsoft/teams.graph-tools somethingseomething`. I haven't yet found a way to make the build output from graph-tools available to npx during a clean build in a way that works across all four node versions. I don't want to block the feature work on that though, I think it's ok to do this for the moment to get the endpoints packages in place, and then solve this bother in a separate effort.

package.json
Jesperholmbergmsft: Fair point - done.

packages/graph-tools/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "@microsoft/teams.graph-tools",
```
aacebo, line 2: is this intended to be published? If not it should be marked as private

aacebo (approved):

packages/graph-endpoints/tsup.config.js
```diff
@@ -0,0 +1,23 @@
+const pkg = require('./package.json');
```
heyitsaamir, line 1: Do we need TSUP here? or can we just use tsc?

packages/graph-tools/package.json
Jesperholmbergmsft: Oh good catch!

packages/graph-endpoints/tsup.config.js
Jesperholmbergmsft: I might be able use just tsc. I'm already doing two passes - one for cjs, one for esm - so it should be doable. Let me fiddle with it and see if I can make it so in a separate change.

PR #291, opened 2025-08-04
Add optional conversation ref parameter to `send` and fix tenant app graph token during app.process

heyitsaamir:
1. Just tested out `send` for the group oauth scenario, and we need to plumb conversation ref through to `send` for this to work. 
2. Also tested out the app-graph scenario, and it was defaulting to using `this._tokens.graph?.toString()` which wasn't working in calling graph successfully.

Tested both scenarios now, and both are working correctly now.

### DIFF
--- a/packages/apps/src/app.process.ts
+++ b/packages/apps/src/app.process.ts
@@ -124,8 +124,8 @@ export async function $process<TPlugin extends IPlugin>(
   }
 
   const send = context.send.bind(context);
-  context.send = async (activity: ActivityLike) => {
-    const res = await send(activity);
+  context.send = async (activity: ActivityLike, conversationRef?: ConversationReference) => {
+    const res = await send(activity, conversationRef);
 
     this.onActivitySent(sender, {
       ...ref,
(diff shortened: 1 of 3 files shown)


packages/apps/src/app.process.ts
```diff
@@ -124,8 +124,8 @@ export async function $process<TPlugin extends IPlugin>(
   }
 
   const send = context.send.bind(context);
-  context.send = async (activity: ActivityLike) => {
-    const res = await send(activity);
+  context.send = async (activity: ActivityLike, conversationRef?: ConversationReference) => {
```
aacebo, line 127: why is this necessary when the activity context already has the conversation reference on it? https://github.com/microsoft/teams.ts/blob/3725bffd404756610c29cb969adfde66567f3bd1/packages/apps/src/contexts/activity.ts#L161

heyitsaamir: For oauth, to be able to send to 1:1, the ref is the group chat. We need to send to the 1:1 chat. Hence this, and https://github.com/microsoft/teams.ts/blob/3725bffd404756610c29cb969adfde66567f3bd1/packages/apps/src/contexts/activity.ts#L290

aacebo (approved):

PR #297, opened 2025-08-07
Send undefined instead of null

heyitsaamir:
In #296 we made it so that we set empty object for empty responses. Turns out the problem was with the body that we're sending. We coalesce to null for undefined objects, but we should really just send undefined. The backend does a check for Content-Length == 0, which is not true for Null, but is true for undefined.

### DIFF
--- a/packages/apps/src/app.process.ts
+++ b/packages/apps/src/app.process.ts
@@ -158,7 +158,7 @@ export async function $process<TPlugin extends IPlugin>(
     await context.stream.close();
 
     if (!res || !isInvokeResponse(res)) {
-      res = { status: 200, body: res ?? {} };
+      res = { status: 200, body: res };
     }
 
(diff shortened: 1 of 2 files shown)


packages/apps/src/app.process.ts
```diff
@@ -158,7 +158,7 @@ export async function $process<TPlugin extends IPlugin>(
     await context.stream.close();
 
     if (!res || !isInvokeResponse(res)) {
-      res = { status: 200, body: res ?? {} };
+      res = { status: 200, body: res };
```
lilyydu, line 161: `body: res` - res can still be null in this case right?

heyitsaamir: yeah if the application decides to send back null, it can be null. But if the application decides to send back undefined, we don't set it to null

PR #300, opened 2025-08-13
Enable sending updates to streams before stream starts

MehakBindra:
To make typescript consistent with python and .net, 

- __IStreamer__ (`packages/apps/src/types/streamer.ts`): Interface definition for the update method

`__HttpStream__ (`packages/apps/src/plugins/http/stream.ts`)`: Added `update()` method implementation for sending informative updates
`__HttpStream__ (`packages/apps/src/plugins/http/stream.ts`)`: Changed flush logic to handle informative updates separately and send them immediately with withChannelData({ streamType: 'informative' })
- __TypingActivity__ (`packages/api/src/activities/typing.ts`): Changed logic to only assign `streamType' etc  if not set

https://github.com/user-attachments/assets/338b5b72-a564-4321-a58f-4b138f13e248

### DIFF
--- a/packages/apps/src/plugins/http/stream.ts
+++ b/packages/apps/src/plugins/http/stream.ts
@@ -6,6 +6,7 @@ import {
   ConversationReference,
   Entity,
   IMessageActivity,
+  ITypingActivity,
   MessageActivity,
   SentActivity,
   TypingActivity,
@@ -26,7 +27,7 @@ export class HttpStream implements IStreamer {
   protected attachments: Attachment[] = [];
   protected channelData: ChannelData = {};
   protected entities: Entity[] = [];
-  protected queue: Array<Partial<IMessageActivity>> = [];
+  protected queue: Array<Partial<IMessageActivity | ITypingActivity>> = [];
 
   private _result?: SentActivity;
   private _timeout?: NodeJS.Timeout;
@@ -39,7 +40,7 @@ export class HttpStream implements IStreamer {
     this._logger = logger?.child('stream') || new ConsoleLogger('@teams/http/stream');
   }
 
-  emit(activity: Partial<IMessageActivity> | string) {
+  emit(activity: Partial<IMessageActivity | ITypingActivity> | string) {
     if (this._timeout) {
       clearTimeout(this._timeout);
       this._timeout = undefined;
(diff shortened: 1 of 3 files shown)


packages/apps/src/plugins/http/stream.ts
```diff
@@ -161,6 +177,22 @@ export class HttpStream implements IStreamer {
     }
   }
 
+  protected async send_activity(activity: TypingActivity) {
```
aacebo: this should be camel case, conforming to the project standards

packages/apps/src/plugins/http/stream.ts
```diff
@@ -161,6 +177,22 @@ export class HttpStream implements IStreamer {
...
+        logger: this._logger
+      });
+      this.events.emit('chunk', res);
+      this.index++;
+      if (!this.id) {
+        this.id = res.id;
+      }
+  }
+
   protected async send(activity: ActivityParams) {
```
aacebo, line 197: why do we have a `send` that sends an activity and also a `send_activity` that sends an activity? There should be a way to consolidate these into one comprehensive method or to name one of the methods more accurately, right now the code is very confusing.

packages/apps/src/types/streamer.ts
```diff
@@ -29,6 +29,12 @@ export interface IStreamer {
    */
   emit(activity: Partial<IMessageActivity | ITypingActivity> | string): void;
```
aacebo, line 30: the `ITypingActivity` in this union was originally intended for informative updates, is it still valid to be here?

MehakBindra: Yes, update internally calls emit -  
```
update(text: string) {
    this.emit({
      type: 'typing',
      text: text,
      channelData: { streamType: 'informative' }
    });
  }
```

packages/apps/src/plugins/http/stream.ts
```diff
@@ -71,6 +80,10 @@ export class HttpStream implements IStreamer {
       await new Promise((resolve) => setTimeout(resolve, 200));
     }
 
+    if (this.text === '' && !this.attachments.length) {
+      this.text = 'Stream completed without content';
```
heyitsaamir: Will this ever send? - Seems risky to keep this in?

MehakBindra: @heyitsaamir when we only have informative updates and no actual text/attachment and then close the stream, yes it'll send. 
We could  just log this and return ?

heyitsaamir: Why would we want to send a hard-coded string in? Will this show up?

MehakBindra: Yep, we probably shouldn't .  I'll make a change to log and return

MehakBindra: Or would sending an empty string be better ? Otherwise it'll be stuck here : 
stream_starting

PR #305, opened 2025-08-20
Auto-generate (almost) all graph beta endpoints

Jesperholmbergmsft:
This big huge massive PR does two things:
a) updates the packages\graph-tools\src\allow-list.ts to include almost all endpoints
b) adds auto-generated endpoints based on the latest beta schema

For a), the allow-list was initially the following - super small to illustrate what the generated code looks like:
```
// the smallest API surface of the whole graph
export const allowList = [/^\/privacy(.*)$/];
```

Now it's changed to this to include almost everything:
```
// include anything but the /identityGovernance endpoints, as they pose MAX_PATH challenges
export const allowList = [/^(?!\/identityGovernance).*$/];
```

A separate PR will add special-treatment to identityGovernance so that we can support those also.

For b) I picked the most recent openapi.yaml from https://github.com/microsoftgraph/msgraph-metadata/blob/master/openapi/beta/openapi.yaml and used that as input to auto-generate the endpoints.  I'm not checking in that file though, as it's so big it doesn't really belong in git.


aacebo:
due to the size of these new graph packages our build time has been increasing quite a bit, I think we should set turbo to cache the graph package builds, which is currently disabled https://turborepo.com/docs/crafting-your-repository/caching

- https://github.com/microsoft/teams.ts/blob/main/packages/graph-endpoints-beta/turbo.json#L7
- https://github.com/microsoft/teams.ts/blob/main/packages/graph-endpoints/turbo.json#L7
- https://github.com/microsoft/teams.ts/blob/main/packages/graph/turbo.json#L7

Jesperholmbergmsft:
> due to the size of these new graph packages our build time has been increasing quite a bit, I think we should set turbo to cache the graph package builds, which is currently disabled https://turborepo.com/docs/crafting-your-repository/caching
> 
> * https://github.com/microsoft/teams.ts/blob/main/packages/graph-endpoints-beta/turbo.json#L7
> * https://github.com/microsoft/teams.ts/blob/main/packages/graph-endpoints/turbo.json#L7
> * https://github.com/microsoft/teams.ts/blob/main/packages/graph/turbo.json#L7

Oh nice catch, I'll make it so

aacebo (approved):

PR #307, opened 2025-08-21
Read port correctly in cloud deployments

MehakBindra:
On IISNode, process.env.PORT is a named pipe like `\\.\pipe\cab9024b-6962-4c74-96ec-d53160071338`

### DIFF
--- a/packages/apps/src/app.ts
+++ b/packages/apps/src/app.ts
@@ -185,7 +185,7 @@ export class App<TPlugin extends IPlugin = IPlugin> {
   protected tenantTokens = new LocalStorage<string>({}, { max: 20000 });
   protected events = new EventEmitter<AppEvents<TPlugin>>();
   protected startedAt?: Date;
-  protected port?: number;
+  protected port?: number | string;
 
   private readonly _userAgent = `teams.ts[apps]/${pkg.version}`;
 
@@ -329,7 +329,7 @@ export class App<TPlugin extends IPlugin = IPlugin> {
    * @param port port to listen on
    */
   async start(port?: number | string) {
-    this.port = +(port || process.env.PORT || 3978);
+    this.port = port || process.env.PORT || 3978;
 
     try {
       await this.refreshTokens(true);
--- a/packages/apps/src/plugins/http/plugin.ts
+++ b/packages/apps/src/plugins/http/plugin.ts
@@ -83,7 +83,7 @@ export class HttpPlugin implements ISender {
   get port() {
     return this._port;
   }
-  protected _port?: number;
+  protected _port?: number | string;
 
   protected express: express.Application;
   protected pending: Record<string, express.Response> = {};
--- a/packages/apps/src/types/plugin/plugin-start-event.ts
+++ b/packages/apps/src/types/plugin/plugin-start-event.ts
@@ -7,5 +7,5 @@ export interface IPluginStartEvent {
    * the port given to the
    * `app.start()` method
    */
-  readonly port: number;
+  readonly port: number | string;
 }


packages/apps/src/app.ts
```diff
@@ -329,7 +329,7 @@ export class App<TPlugin extends IPlugin = IPlugin> {
    * @param port port to listen on
    */
   async start(port?: number | string) {
-    this.port = +(port || process.env.PORT || 3978);
+    this.port = port || process.env.PORT || 3978;
```
heyitsaamir, line 333: won't this still break?
if the port's value is `\\.\pipe\cab9024b-6962-4c74-96ec-d53160071338`

MehakBindra: When port is a named pipe, then:
+(port) leads to NaN because + is not able to convert a string like \\.\pipe\... to a number.
This keeps the string 
server.listen(this.port, () => { .. } works fine

packages/dev/src/plugin.ts
```diff
@@ -103,11 +103,13 @@ export class DevtoolsPlugin implements ISender {
   }
 
   onStart({ port }: IPluginStartEvent) {
-    port += 1;
+    let numericPort = typeof port === 'string' ? parseInt(port, 10) : port;
+    numericPort += 1;
+    console.log(`Starting devtools on port ${numericPort}`);
```
MehakBindra: Wondering if this is the best way to do this, devtools is a completely local thing right? Should not be trying to start it in azure ?

If you do, we'll get this:
Starting devtools on port NaN

packages/apps/src/app.ts
heyitsaamir: but won't `server.listen("\\.\pipe\cab9024b-6962-4c74-96ec-d53160071338", () => { ... })` fail?

MehakBindra: in Node.js, the server.listen method can take a named pipe path [ref](https://nodejs.org/api/net.html?utm_source=chatgpt.com#:~:text=The%20node%3Anet%20module%20supports%20IPC%20with%20named%20pipes%20on%20Windows%2C%20and%20Unix%20domain%20sockets%20on%20other%20operating%20systems.)

packages/dev/src/plugin.ts
Jesperholmbergmsft: Might aid trouble shooting if this would test for Number.isNaN() and if so log something readable with this.log.error()?

packages/apps/src/app.ts
heyitsaamir: Oof! TIL! Thanks! that's good enough for me.

Jesperholmbergmsft (approved):

packages/dev/src/plugin.ts
heyitsaamir: Hmm, yeah this might not be an important enough issue for us to solve. 
But:
1. I'd warn if this happens
2. I'd provide an optional parameter in the DevTools plugin that someone can provide a custom port if they want. So we have a workaround.

heyitsaamir (commented):
Lg2m, added a comment on how to deal with devtools.

packages/dev/src/plugin.ts
MehakBindra: added optional parameter. we do log - " Starting devtools on port NaN. options.port should be >= 0 and < 65536. Received type number (NaN).", should be enough ?

MehakBindra: @heyitsaamir , @Jesperholmbergmsft will need a fresh review because of change

packages/dev/src/plugin.ts
```diff
@@ -69,7 +69,9 @@ export class DevtoolsPlugin implements ISender {
   protected pending: Record<string, ResolveRejctPromise> = {};
   protected pages: Array<Page> = [];
 
-  constructor() {
+  private readonly customPort?: number;
```
heyitsaamir: NIT: put this in an object (DevToolsPluginOptions). This way future additions will just be another property in this object, and we won't risk having to ship breaking changes.

packages/dev/src/plugin.ts
```diff
@@ -103,11 +106,13 @@ export class DevtoolsPlugin implements ISender {
   }
 
   onStart({ port }: IPluginStartEvent) {
-    port += 1;
+    const numericPort = this.customPort ?? (
+    typeof port === 'string' ? parseInt(port, 10) + 1: port + 1);
+    console.log(`Starting devtools on port ${numericPort}`);
```
heyitsaamir: this should use the logger, not console.log. Also (logger.debug)

heyitsaamir (approved):

aacebo (approved):

PR #309, opened 2025-08-22
Enable build output caching for graph projects

Jesperholmbergmsft:
This takes Alex's suggestion from https://github.com/microsoft/teams.ts/pull/305 to enable build output caching in the graph projects and avoid redundant rebuilds.

### DIFF
--- a/packages/graph-endpoints-beta/turbo.json
+++ b/packages/graph-endpoints-beta/turbo.json
@@ -3,9 +3,9 @@
   "tasks": {
     "build": {
       "inputs": ["$TURBO_DEFAULT$", ".env*"],
-      "outputs": [".next/**", "!.next/cache/**"],
-      "cache": false,
-      "dependsOn": ["@microsoft/teams.common#build", "@microsoft/teams.graph-tools#build", "@microsoft/teams.graph-endpoints#build"]
+      "outputs": ["dist/**", "!dist/cache/**"],
+      "cache": true,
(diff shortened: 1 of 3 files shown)


packages/graph-endpoints-beta/turbo.json
```diff
@@ -3,9 +3,9 @@
   "tasks": {
     "build": {
       "inputs": ["$TURBO_DEFAULT$", ".env*"],
-      "outputs": [".next/**", "!.next/cache/**"],
-      "cache": false,
-      "dependsOn": ["@microsoft/teams.common#build", "@microsoft/teams.graph-tools#build", "@microsoft/teams.graph-endpoints#build"]
+      "outputs": ["dist/**", "!dist/cache/**"],
```
Jesperholmbergmsft, line 6: TIL that output folder doesn't affect whether the output gets cached, but it does affect whether packages that are dependent on the output can actually find it. Who knew?

packages/graph-endpoints-beta/turbo.json
```diff
@@ -3,9 +3,9 @@
...
+      "cache": true,
+      "dependsOn": ["@microsoft/teams.graph-tools#build"]
```
Jesperholmbergmsft, line 8: This and graph-endpoints aren't depending on common or each other, just the tools.

heyitsaamir (approved):

PR #311, opened 2025-08-22
Update tests projects to use the new Graph calling pattern

Jesperholmbergmsft:
This PR updates the tab & graph tests to use the new graph calling pattern. It also makes some overall improvements to the tab test to better showcase some of the features of this library
 - pre-warming graph token consent
 - getting the teamsJs context
 - making a remote function call
 - having the remote function handler post a message to the current channel, and
 - a couple graph calls to fetch me & presence, and to set presence.

Tab test hosted in Teams:
image

### DIFF
--- a/tests/graph/src/index.ts
+++ b/tests/graph/src/index.ts
@@ -1,14 +1,15 @@
 import { App } from '@microsoft/teams.apps';
 import { ConsoleLogger } from '@microsoft/teams.common/logging';
+import * as endpoints from '@microsoft/teams.graph-endpoints';
 
 // :snippet-start: auth-config
-const app = new App({ 
+const app = new App({
   oauth: { // oauth configurations
     /**
      * The name of the auth connection to use.
      * It should be the same as the OAuth connection name defined in the Azure Bot configuration.
      */
-    defaultConnectionName: 'graph' 
+    defaultConnectionName: 'graph'
   },
   logger: new ConsoleLogger('@tests/auth', { level: 'debug' })
 });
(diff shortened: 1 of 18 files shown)


tests/tab/src/Tab/App.tsx
```diff
@@ -1,33 +1,107 @@
...
+  }, [app]);
+
+
+  const whoAmI = React.useCallback(async () => {
+    if (!app) {
+      return;
+    }
+
+    // get the current user from the Microsoft Graph
+    const me = await app.graph.call(endpoints.me.get);
```
Jesperholmbergmsft, line 59: I wanted to make this a remote function call to `await app.exec('whoAmI')` and use the userGraph prop in the function handler to get & return `me`, to show both graph on the bot side and passing return values -- but I realized, there's no userGraph prop in the function handler yet. I should have time to add the userGraph prop too before GA, I think, and I'll update this when I do.

tests/graph/src/index.ts
```diff
@@ -28,17 +29,17 @@ app.on('message', async ({ log, signin, userGraph, isSignedIn }) => {
...
       // Customize the OAuth card text (only applies to OAuth flow, not SSO)
       oauthCardText: 'Sign in to your account',
-      signInButtonText: 'Sign in' 
+      signInButtonText: 'Sign in'
     }); // call signin for your auth connection...
     return;
   }
 
-  const me = await userGraph.me.get();
+  const me = await userGraph.call(endpoints.me.get);
```
heyitsaamir, line 37: do you need to update the templates too?

Jesperholmbergmsft: yes - right you are - I forgot all this code is duplicated...

PR #317, opened 2025-08-27
[apps] add default route overrides

aacebo:
- add default route overrides so default routes registered by the app can be overridden by the user
- add tests for router

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@microsoft/teams.ts",
-  "version": "2.0.0-preview.9",
+  "version": "2.0.0-preview.10",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@microsoft/teams.ts",
-      "version": "2.0.0-preview.9",
+      "version": "2.0.0-preview.10",
       "workspaces": [
         "packages/*",
         "external/*",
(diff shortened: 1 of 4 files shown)


packages/apps/src/contexts/activity.ts
```diff
@@ -242,7 +259,7 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
     }
 
     const tokenExchangeState: TokenExchangeState = {
-      connectionName: this.connectionName,
+      connectionName: connectionName || this.connectionName,
```
heyitsaamir, line 262: if connectionName is passed in, they they have to do the work of overriding the verify-state event, because we use default to checking against the defaultOauthConnection

heyitsaamir: (it was the tradeoff we'd made to not make "connectionName" overrideable)

aacebo: yes, thats the point of this PR to enable custom auth handling for developers that want to handle the entire flow themselves.

Jesperholmbergmsft (approved):

packages/apps/src/contexts/activity.ts
heyitsaamir: We'll prob need to do this then right? https://github.com/microsoft/teams.ts/pull/280? 

Actually we prob need to do what you said in that PR - replace the onVerifyState handler?

heyitsaamir: Otherwise, how can the developer override the on-verify-state handler that the app sets by default?

aacebo: since the app now registers its default routes as default, the router will override them when the developer registers one

aacebo: this test case for example https://github.com/microsoft/teams.ts/blob/0d59f95067048f6fd03800e1d19613e1acfdc483/packages/apps/src/router.spec.ts#L40

heyitsaamir (approved):

yuqizhou77:
I think with overrideSignInActivity, the SDK will still call token service like `const resource = await this.api.bots.signIn.getResource({ state });` and `this.api.users.token.get`, which will fail if user choose not to use the bot service and OAuth connection for auth.

PR #318, opened 2025-08-27
Improved endpoint function typing and readme

Jesperholmbergmsft:
This PR improves the endpoint functions to not generate body & parameter arguments for APIs that don't need them. This reduces type clutter a little, and avoids generating impossible signatures.

For instance, `POST /me/reprocessLicenseAssignment` was generated with this effective signature:
`function create(body: never, params?: {} | undefined) : whatever`

This API doesn't take any body or params at all, so the types.ts `requestBody` type for this API is 'never'. Our code has assumed that all non-get, non-delete APIs will always require a body however, so we've been happily generating the `body` argument. The result is that `body: never` argument - which is problematic as there's nothing I can pass to satisfy it. 

The empty `params` type is not problematic as such, but it's a bit of clutter.

To make this PR reviewable, I only re-generated a handful of endpoints to show the difference before/after the changes. and keep the PR readable I'm planning to re-generate everything next week and that'll propagate the fixes.

Other minor things
 - paramDefs are made optional in the EndpointRequest type, to match params and reduce clutter
 - the EndpointRequest and CallOptions types are exported out of the Graph client, to make it easier to make custom endpoint functions.
 - the graph-endpoints and graph-endpoints-beta readmes are updated to correct links & instructions.

### DIFF
--- a/packages/graph-endpoints/src/me/appRoleAssignments.ts
+++ b/packages/graph-endpoints/src/me/appRoleAssignments.ts
@@ -104,14 +104,11 @@ export function update(
  *
  */
 export function create(
-  body: IEndpoints['POST /me/appRoleAssignments']['body'],
-  params?: IEndpoints['POST /me/appRoleAssignments']['parameters']
+  body: IEndpoints['POST /me/appRoleAssignments']['body']
 ): EndpointRequest<IEndpoints['POST /me/appRoleAssignments']['response']> {
   return {
     method: 'post',
     path: '/me/appRoleAssignments',
-    paramDefs: [],
-    params,
     body,
   };
 }
--- a/packages/graph-endpoints/src/me/removeAllDevicesFromManagement.ts
+++ b/packages/graph-endpoints/src/me/removeAllDevicesFromManagement.ts
@@ -12,15 +12,11 @@ export interface IEndpoints {
  *
  * Retire all devices from management for this user
  */
-export function create(
-  body: IEndpoints['POST /me/removeAllDevicesFromManagement']['body'],
-  params?: IEndpoints['POST /me/removeAllDevicesFromManagement']['parameters']
-): EndpointRequest<IEndpoints['POST /me/removeAllDevicesFromManagement']['response']> {
+export function create(): EndpointRequest<
+  IEndpoints['POST /me/removeAllDevicesFromManagement']['response']
+> {
   return {
     method: 'post',
     path: '/me/removeAllDevicesFromManagement',
-    paramDefs: [],
-    params,
-    body,
   };
 }
--- a/packages/graph-endpoints/src/types/common.ts
+++ b/packages/graph-endpoints/src/types/common.ts
@@ -4,7 +4,7 @@ import type { paths } from './types.ts';
 export type EndpointRequest<TResponse> = {
   method: 'get' | 'post' | 'patch' | 'delete' | 'put';
   path: string;
-  paramDefs: Array<{ name: string; in: 'query' | 'header' | 'path' }>;
+  paramDefs?: Array<{ name: string; in: 'query' | 'header' | 'path' }>;
   params?: Record<string, any>;
   body?: any;
   responseType?: TResponse; 
--- a/packages/graph-tools/src/endpoints.ts
+++ b/packages/graph-tools/src/endpoints.ts
@@ -62,6 +62,7 @@ interface IEndpoint {
   readonly parameters?: Array<OpenAPIV3.ParameterObject>;
   readonly description?: string;
   readonly deprecated?: boolean;
+  readonly hasRequestBody: boolean;
 }
 
 class Client {
@@ -205,6 +206,11 @@ class Client {
         name = name.slice(0, name.length - 1);
       }
 
+      if (method === 'delete') {
+        // 'delete' is a reserved word and can't be used as function name
+        name = 'del';
+      }
+
       this.endpoints[`${method.toUpperCase()} ${schema.url}`] = {
         method,
         name: this.getUniqueName(name),
@@ -212,6 +218,7 @@ class Client {
         parameters: params,
         description: def.description,
         deprecated: def.deprecated,
+        hasRequestBody: !!def.requestBody,
       };
     }
   }
--- a/packages/graph-endpoints-beta/README.md
+++ b/packages/graph-endpoints-beta/README.md
@@ -1,25 +1,30 @@
 # Teams: Graph
 
 <p>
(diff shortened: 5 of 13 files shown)


packages/graph-endpoints/src/me/appRoleAssignments.ts
```diff
@@ -104,14 +104,11 @@ export function update(
  *
  */
 export function create(
-  body: IEndpoints['POST /me/appRoleAssignments']['body'],
-  params?: IEndpoints['POST /me/appRoleAssignments']['parameters']
+  body: IEndpoints['POST /me/appRoleAssignments']['body']
```
Jesperholmbergmsft, line 107: This API don't take any parameters, so the effect params argument was `params?: {} | undefined`. That's just clutter, we don't need that.

packages/graph-endpoints/src/me/removeAllDevicesFromManagement.ts
```diff
@@ -12,15 +12,11 @@ export interface IEndpoints {
  *
  * Retire all devices from management for this user
  */
-export function create(
-  body: IEndpoints['POST /me/removeAllDevicesFromManagement']['body'],
-  params?: IEndpoints['POST /me/removeAllDevicesFromManagement']['parameters']
-): EndpointRequest<IEndpoints['POST /me/removeAllDevicesFromManagement']['response']> {
+export function create(): EndpointRequest<
```
Jesperholmbergmsft, line 15: This one was generated with `body: never, params?: {} | undefined` which is pretty problematic. Let's just omit the arguments when not needed.

packages/graph-endpoints/src/types/common.ts
```diff
@@ -4,7 +4,7 @@ import type { paths } from './types.ts';
 export type EndpointRequest<TResponse> = {
   method: 'get' | 'post' | 'patch' | 'delete' | 'put';
   path: string;
-  paramDefs: Array<{ name: string; in: 'query' | 'header' | 'path' }>;
+  paramDefs?: Array<{ name: string; in: 'query' | 'header' | 'path' }>;
```
Jesperholmbergmsft, line 7: `paramDefs` is used to build up the HTTP request, to figure out where exactly the incoming arguments go. Not all endpoints take parameters however, so it shouldn't be mandatory.

packages/graph-tools/src/endpoints.ts
```diff
@@ -205,13 +206,19 @@ class Client {
         name = name.slice(0, name.length - 1);
       }
 
+      if (method === 'delete') {
```
Jesperholmbergmsft, line 209: By fixing up the name here and adding the hasRequestBody variable, I could consolidate the 3 separate function patterns in the client.ts.hbs. We had one for 'get', one for 'delete' (same as get, but overrode the name), and one for everything else (same as get, but always added the body -- which is problematic).

heyitsaamir: can you explain this a bit more?

heyitsaamir (approved):

packages/graph-tools/src/endpoints.ts
Jesperholmbergmsft: Sure - the client.ts.hbs in this package is based on the one in packages\graph\scripts\client.ts.template, and I carried forward this logic to render the actual functions:
```
{{#if (eq method "get")}}
export function {{name}}(
  params?: IEndpoints['{{@key}}']['parameters'], 
  ) : EndpointRequest<IEndpoints['{{@key}}']['response']> {
  return {
    {{#if (eq ../apiVersion "beta")}}
    ver: 'beta',
    {{/if}}
    method: 'get',
    path: '{{url}}',
    paramDefs: [
      {{#if (notEmpty parameters)}}
      {{#each parameters}}
      { name: '{{name}}', in: '{{in}}' },
      {{/each}}
      {{/if}}
    ],
    params
  };
}
{{else if (eq method "delete")}}
export function del(
  params?: IEndpoints['{{@key}}']['parameters'], 
) : EndpointRequest<IEndpoints['{{@key}}']['response']> {
  return {
    {{#if (eq ../apiVersion "beta")}}
    ver: 'beta',
    {{/if}}
    method: 'delete',
    path: '{{url}}',
    paramDefs: [
      {{#if (notEmpty parameters)}}
      {{#each parameters}}
      { name: '{{name}}', in: '{{in}}' },
      {{/each}}
      {{/if}}
    ],
    params
  };
}
{{else}}
export function {{name}}(
  body: IEndpoints['{{@key}}']['body'],
  params?: IEndpoints['{{@key}}']['parameters'],
) : EndpointRequest<IEndpoints['{{@key}}']['response']> {
  return {
    {{#if (eq ../apiVersion "beta")}}
    ver: 'beta',
    {{/if}}
    method: '{{method}}',
    path: '{{url}}',
    paramDefs: [
      {{#if (notEmpty parameters)}}
      {{#each parameters}}
      { name: '{{name}}', in: '{{in}}' },
      {{/each}}
      {{/if}}
    ],
    params,
    body,
  };
}
{{/if}}
```

That thing had three cases - one for 'get' APIs, one for 'delete', and one for other. They're so similar that it's hard to see how they vary and why there's three.

The only difference between 'get' and 'delete' was that the name was tweaked for the latter. From 'delete' to 'del', to avoid using a reserved keword. By doing that rename here in endpoints.ts instead, I'm able to collapse the first two cases into just one.

The difference between 'get' and the others was that the others include the body parameter. Now that I found that body should only be included conditionally, well we can safely do that for get & delete too as they never have a request body anyway. 

Thanks to that we can go from the three patterns we started with to just a single one, here:
```
export function {{name}}(
  {{#if hasRequestBody}}
  body: IEndpoints['{{@key}}']['body'],
  {{/if}}
  {{#if (notEmpty parameters)}}
  params?: IEndpoints['{{@key}}']['parameters'],
  {{/if}}
) : EndpointRequest<IEndpoints['{{@key}}']['response']> {
  return {
    {{#if (eq ../apiVersion "beta")}}
    ver: 'beta',
    {{/if}}
    method: '{{method}}',
    path: '{{url}}',
    {{#if (notEmpty parameters)}}
    paramDefs: [
      {{#each parameters}}
      { name: '{{name}}', in: '{{in}}' },
      {{/each}}
    ],
    params,
    {{/if}}
    {{#if hasRequestBody}}
    body,
    {{/if}}
  };
}
```

Much nicer  - this should be easier to maintain without having to play spot-the-difference between the three cases.

aacebo (approved):

PR #320, opened 2025-08-29
Remove old graph endpoints and scripts

Jesperholmbergmsft:
This removes the old graph endpoints & their scripts, in order to fully move to the new calling pattern.

Besides deleting a bunch of things, I'm making some minor tweaks to the remaining code:
 - we're using the package.json version number in the HTTP client user-agent header. So far we've been importing it directly from `src/../package.json`. It's a little odd to import from outside the src folder, so instead I'm injecting the version as a global variable through tsup.config.json
 - we were importing from `@utils/url` which looks like an external dependency, but was actually a mapping to code within the package itself, with tsconfg.json having: `"paths": { "@utils/*": ["src/utils/*"] }`. I'm guessing that that was done to make it easy to import that URL utility from all the auto-generated endpoints, but that is no longer needed. I removed the mapping and changed to import from `./utils/url` as you normally would.
 - in addition, a little cleanup & renames, and added unit tests for the URL utility.

How tested:
 - built the packages and installed in a separate test app
 - verified that I can still call graph functions using the new calling pattern
 - verified that the package version is properly set when making graph functions. after installing the package, I see this in node_modules\@microsoft\teams.graph\dist\index.mjs: `this.http=t.clone({baseUrl:"https://graph.microsoft.com/v1.0",headers:{"Content-Type":"application/json","User-Agent":"teams.ts[graph]/2.0.0-preview.10"}}):`
 - verified that the package size and final bundle size of my app has dropped significantly

Before the change, source-map-explorer showed this:
```
        "../../../node_modules/@microsoft/teams.graph/dist/index.mjs": {
          "size": 1348868
        },
```
After the change, this:
```
        "../../../node_modules/@microsoft/teams.graph/dist/index.mjs": {
          "size": 1198
        },
```


packages/graph/jest.config.js
```diff
@@ -1,16 +1,9 @@
...
-    'src/index.ts', // Only collect coverage for this file
-    '!**/*.d.ts',
-    '!**/dist/**',
-    '!**/coverage/**',
-    '!**/node_modules/**',
-  ],
-  moduleNameMapper: {
-    '^@utils/(.*)$': '<rootDir>/src/utils/$1',
-    '^src/../package.json$': '<rootDir>/package.json',
+  globals: {
```
aacebo, line 6: what does this do out of curiosity? Should we be doing it for all our packages since they are all on the same version?

aacebo (approved):

packages/graph/jest.config.js
Jesperholmbergmsft: All of the endpoints create an HTTP client where the user-agent header includes the graph package version. It went like this:
```
import pkg from 'src/../package.json';
...
      this.http = new http.Client({
        baseUrl: 'https://graph.microsoft.com/v1.0',
        headers: {
          'Content-Type': 'application/json',
          'User-Agent': `teams.ts[graph]/${pkg.version}`,
        },
      });
```

It's a little weird to import package.json like that, from outside the src folder and everything. So I changed that to inject the package version as a global variable instead. I do that in the tsup.config.js, and also here for the unit tests. to compile. The change to the http client constructor is:
```
      this.http = new http.Client({
        baseUrl: 'https://graph.microsoft.com/v1.0',
        headers: {
          'Content-Type': 'application/json',
          'User-Agent': `teams.ts[graph]/${__PACKAGE_VERSION__}`,
        },
      });
```

I didn't see other places where we're importing from outside the src folder, but if we do need to reference the package version in code, this is a cleaner way to do it.

PR #322, opened 2025-08-30
Do not stream AI messages all the time

heyitsaamir:
We were sending in onChunk all the time, so all AI messages were streamed. I this PR, we only send onChunk if it's actually provided.

### DIFF
--- a/packages/ai/src/prompts/chat.ts
+++ b/packages/ai/src/prompts/chat.ts
@@ -243,7 +243,7 @@ export class ChatPrompt<
         messages,
         request: options.request,
         functions: fnMap,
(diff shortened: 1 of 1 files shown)


Jesperholmbergmsft (approved):

packages/ai/src/prompts/chat.ts
```diff
@@ -243,7 +243,7 @@ export class ChatPrompt<
         messages,
         request: options.request,
         functions: fnMap,
-        onChunk: async (chunk) => {
+        onChunk: onChunk ? async (chunk) => {
```
aacebo, line 246: wow good catch! Silly me

PR #328, opened 2025-09-04
Reduce graph endpoint file count

Jesperholmbergmsft:
This updates the graph endpoint generation script to reduce the number of files generated. When a client doesn't have any children, it's no longer put in its own separate .ts file, but instead exported as an object out of its parent's index.ts. This reduces the file count by about 75%, which in turn cuts build time in about half, reduces the uncompressed code size by about 20%, the NPM package size by about 15%. I didn't measure the time taken to `npm install` the graph-endpoints package, but that too is improved noticeably.

This will result in a big huge diff. To keep this PR readable I only re-generated endpoints under /me/teamwork - enough to illustrate the difference. 

In addition, there's a cheat to tweak the file system path generated for some of the nodes to enshorten the paths and help avoid the risk of creating paths so long that even Windows can't handle them. The enshortening will not be noticable to most developers, unless they go and do a deep import like
`import { accessPackageResourceEnvironment} from '@microsoft/teams.graph-endpoints/identityGovernance\entitlementManagement\accessPackageAssignments\accessPackage\accessPackageResourceRoleScopes\accessPackageResourceRole\accessPackageResource\accessPackageResourceScopes\accessPackageResource`;`

With the change, they'll have to import from our custom enshortened path:
```
import { accessPackageResourceEnvironment} from '@microsoft/teams.graph-endpoints/identityGovernance\entitlementManagement\apAssignments\ap\resourceRoleScopes\apResourceRole\apResource\scopes\apResource';
...
await graph.call(accessPackageResourceEnvironment.create);
```

or if they'd rather follow the actual Graph schema, they can do
```
...
await graph.call(identityGovernance.entitlementManagement.accessPackageAssignments.accessPackage.accessPackageResourceRoleScopes.accessPackageResourceRole.accessPackageResource.accessPackageResourceScopes.accessPackageResource.accessPackageResourceEnvironment.create);
```

This is not a breaking change, as I omitted these endpoints until now with the allow-list. 

The allow-list is no longer needed for us so that's removed too, but I kept most of the plumbing in place to keep us set up to build tooling with custom allow-lists - for when a developer would like to build their own custom set of endpoints.


packages/graph-endpoints/src/me/teamwork/associatedTeams/index.ts
```diff
@@ -111,14 +113,34 @@ export function update(
  *
  */
 export function create(
-  body: IEndpoints['POST /me/teamwork/associatedTeams']['body'],
-  params?: IEndpoints['POST /me/teamwork/associatedTeams']['parameters']
```
Jesperholmbergmsft, line 115: A prior change has us omitting params and paramDefs when the params is typed as just `{}`. This is the first time I re-generate the endpoints since that change, is why this diff shows up.

packages/graph-endpoints/src/me/teamwork/associatedTeams/index.ts
```diff
@@ -111,14 +113,34 @@ export function update(
...
   return {
     method: 'post',
     path: '/me/teamwork/associatedTeams',
-    paramDefs: [],
-    params,
     body,
   };
 }
+
+export const team = {
```
Jesperholmbergmsft, line 125: `team` doesn't have any children, so it no longer gets its own separate file. Instead it's exported as an object with a get function here. The consumer use it just as before, and if we ever need to move it to a sub-folder - if a future schema update adds children - that won't break consumers either.

packages/graph-tools/templates/client.ts.hbs
```diff
@@ -12,47 +12,61 @@ export interface IEndpoints {
...
-  * {{description}}
-  {{/if}}
-  {{#if deprecated}}
-  * @deprecated
-  {{/if}}
-  */
-export function {{name}}(
-  {{#if hasRequestBody}}
-  body: IEndpoints['{{@key}}']['body'],
+{{> functionDefinition apiVersion=../apiVersion isStandaloneExport=true }}
```
Jesperholmbergmsft, line 31: I moved the code to generate the endpoint function itself in the new function-definition.ts.hbs, so I can use the same `functionDefinition` partial here for the free floating functions as well as down on line 65 for the now-inlined children.

heyitsaamir (approved):

packages/graph-endpoints/src/me/teamwork/associatedTeams/index.ts
```diff
@@ -111,14 +113,34 @@ export function update(
  *
  */
 export function create(
-  body: IEndpoints['POST /me/teamwork/associatedTeams']['body'],
-  params?: IEndpoints['POST /me/teamwork/associatedTeams']['parameters']
+  body: IEndpoints['POST /me/teamwork/associatedTeams']['body']
```
heyitsaamir, line 116: no params for create?

Jesperholmbergmsft: In this earlier PR - https://github.com/microsoft/teams.ts/pull/318 - I started omitting params when endpoint doesn't actually take any parameters. I didn't re-generate all endpoints after making that change, but now that I generated the /me/teamwork ones to illustrate the effect of the script changes in this PR, those earlier changes get showcased too.

PR #329, opened 2025-09-04
Update graph client to allow for different national clouds and route beta endpoint requests correctly

Jesperholmbergmsft:
This PR has three changes to the graph client:

a) Adds a baseUrlRoot argument to let the graph client support different clouds. It still defaults to the public https://graph.microsoft.com, but can be overridden for special clouds.

b) In the call method, uses either /v1.0 or /beta baseUrl depending on the `ver` field in the endpoint request builder.

c) While testing, I realized that we never included the header parameters in the request. Query and path parameters - yes; header - no. This is a problem for instance for Graph APIs that rely on the `If-Match` header for optimistic concurrency -- e.g. [this one](https://learn.microsoft.com/en-us/graph/api/resources/planner-overview?view=graph-rest-1.0#planner-resource-versioning). The developer could work around that with the requestParams argument, but that's not nearly as convenient as just using the params - that are typed to support these headers anyway. Let's fix this too.

To test this, in a test app I call:
```
import * as betaEndpoints from '@microsoft/teams.graph-endpoints-beta';
...
    const graph: Graph = app.graph;
    const betaMe = await graph.call(betaEndpoints.me.get);
    const v1Me = await graph.call(endpoints.me.get);
```

Network tab shows that the requests are routed correctly and both give the expected results.
image

To test the request header change, I called:
```
graph.call(endpoints.agreementAcceptances.del, { "agreementAcceptance-id": "bogus", "If-Match": "abc123" });
```
The network tab showed that the `If-Match` request header was not included before the change, but is included after the change.

### DIFF
--- a/packages/graph/src/index.ts
+++ b/packages/graph/src/index.ts
@@ -1,13 +1,13 @@
 import * as http from '@microsoft/teams.common/http';
 
-import { getInjectedUrl } from './utils/url';
+import { getInjectedUrl, getInjectedRequestConfig } from './utils/url';
 
 import type { CallOptions, EndpointRequest, SchemaVersion } from './types';
 
 // Build-time constant injected by tsup
 declare const __PACKAGE_VERSION__: string;
 
-export { CallOptions, EndpointRequest } from './types';
+export { CallOptions, EndpointRequest, SchemaVersion } from './types';
 
 const defaultBaseUrlRoot = 'https://graph.microsoft.com';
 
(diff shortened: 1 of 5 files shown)


lilyydu (approved):

packages/graph/src/index.ts
```diff
@@ -1,34 +1,45 @@
...
-import type { CallOptions, EndpointRequest } from './types';
+import type { CallOptions, EndpointRequest, SchemaVersion } from './types';
 
 // Build-time constant injected by tsup
 declare const __PACKAGE_VERSION__: string;
 
-export { CallOptions, EndpointRequest } from './types';
+export { CallOptions, EndpointRequest, SchemaVersion } from './types';
+
+const defaultBaseUrlRoot = 'https://graph.microsoft.com';
```
heyitsaamir, line 12: NIT: I think convention is for constants like this to be `UPPER_SNAKE_CASE`

heyitsaamir (approved):
nice catch!

packages/graph/src/index.ts
Jesperholmbergmsft: 👍 I'll do that in the next PR, so as not to reset sign-off

PR #333, opened 2025-09-07
feat(cli): add Python AI template for CLI generator

rajan-chari:
Add a cli template for python sdk ai package

### DIFF
--- a/packages/cli/templates/python/ai/.env.example
+++ b/packages/cli/templates/python/ai/.env.example
@@ -1,7 +1,0 @@
-# OpenAI settings
-OPENAI_API_KEY=
-OPENAI_MODEL=gpt-4o-mini
-
-# Microsoft Teams settings (set by the CLI when creating the project)
-MICROSOFT_APP_ID={{ microsoftAppId }}
-MICROSOFT_APP_PASSWORD={{ microsoftAppPassword }}
--- a/packages/cli/templates/python/ai/.env.example
+++ b/packages/cli/templates/python/ai/.env.example
@@ -1,0 +1,7 @@
+# OpenAI settings
+OPENAI_API_KEY=
+OPENAI_MODEL=gpt-4o-mini
+
+# Microsoft Teams settings (set by the CLI when creating the project)
+MICROSOFT_APP_ID={{ microsoftAppId }}
+MICROSOFT_APP_PASSWORD={{ microsoftAppPassword }}
--- a/packages/cli/templates/python/ai/apppackage/manifest.json.hbs
+++ b/packages/cli/templates/python/ai/apppackage/manifest.json.hbs
@@ -1,42 +1,54 @@
 {
-  "$schema": "https://developer.microsoft.com/en-us/json-schemas/teams/v1.16/MicrosoftTeams.schema.json",
-  "manifestVersion": "1.16",
+  "$schema": "https://developer.microsoft.com/json-schemas/teams/v1.20/MicrosoftTeams.schema.json",
   "version": "1.0.0",
-  "id": "{{#if microsoftAppId}}{{ microsoftAppId }}{{else}}00000000-0000-0000-0000-000000000000{{/if}}",
-  "packageName": "com.example.{{ toKebabCase name }}",
-  "developer": {
-    "name": "Your Company",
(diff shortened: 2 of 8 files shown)


packages/cli/templates/python/ai/apppackage/manifest.json.hbs
```diff
@@ -0,0 +1,42 @@
+{
```
Jesperholmbergmsft, line 1: This manifest looks a bit different from the sibling https://github.com/Jesperholmbergmsft/spark.js/blob/main/packages/cli/templates/python/echo/apppackage/manifest.json.hbs for non-obvious reasons. Different manifest version, different developer name, .... should they be harmonized?

packages/cli/templates/python/ai/.env.example
```diff
@@ -0,0 +1,7 @@
+# OpenAI settings
```
Jesperholmbergmsft: Are the consumers expected to populate this manually? For the tab stuff we have ATK templates to create the env files and we don't have an example in the repo.

lilyydu: nit: we traditionally named this file `sample.env`

packages/cli/templates/python/ai/pyproject.toml
```diff
@@ -0,0 +1,15 @@
+[project]
+name = "py-ai-bot"
```
lilyydu: nit: following general standard, "ai-bot" should be sufficient

packages/cli/templates/python/ai/src/main.py
```diff
@@ -0,0 +1,36 @@
...
+app_options = AppOptions(
+    plugins=[DevToolsPlugin()]
+)
+
+app = App(app_options)
+
+@app.on_message
+async def handle_message(event):
+    model = OpenAICompletionsAIModel(
+        model="gpt-4o",
```
lilyydu: should this be pulled in from the env?

packages/cli/templates/python/ai/src/main.py
```diff
@@ -0,0 +1,36 @@
+# python
```
lilyydu: nit: can remove this note

packages/cli/templates/python/ai/src/main.py
```diff
@@ -0,0 +1,36 @@
...
+        model=model
+    )
+
+    # send text to LLM and stream chunks to the client
+    await prompt.send(
+        event.activity.text, 
+        on_chunk=lambda chunk: event.stream.emit(chunk))
+
+async def main():
+    port = int(os.environ.get("PORT", "3978"))
```
lilyydu, line 36: could provide option to set this field in `sample.env`

packages/cli/templates/python/ai/apppackage/manifest.json.hbs
rajan-chari: Fixed to match echobot

packages/cli/templates/python/ai/.env.example
rajan-chari: Removed.  Thanks

rajan-chari: Removed as suggested by @Jesperholmbergmsft

packages/cli/templates/python/ai/pyproject.toml
rajan-chari: Changed.  Thanks

packages/cli/templates/python/ai/src/main.py
rajan-chari: Done. Thanks

packages/cli/templates/python/ai/src/main.py
rajan-chari: Changed.  Read using os.environ.get().  Is that what you had in mind @lilyydu ?

packages/cli/templates/python/ai/src/main.py
rajan-chari: Changed.  Read using os.environ.get(). Is that what you had in mind @lilyydu ?

Jesperholmbergmsft (approved):

PR #335, opened 2025-09-08
Simplify graph function paramDefs type

Jesperholmbergmsft:
This PR changes the `paramDefs` type in the endpoint functions. The `paramDefs` values are used to construct the Graph HTTP request - for each parameter that the caller may supply, `paramDefs` decides whether they go in the path, the query string, or the request header.

The `paramDefs` type so far has been: `{ name: string; in: 'query' | 'header' | 'path' }>[]`. So - an  array of objects, each has a parameter name, and an `in` value to show where it goes.

The new type is: `Partial<Record<'query' | 'header' | 'path', string[]>>;` I'm pulling the 'in' values up a level to be a key in an object, the values being the parameters for that 'in'. 

In practice, the paramDefs for an API before may have been:
```
    paramDefs: [
      { name: '$top', in: 'query' },
      { name: '$skip', in: 'query' },
      { name: '$search', in: 'query' },
      { name: '$filter', in: 'query' },
      { name: '$count', in: 'query' },
      { name: '$orderby', in: 'query' },
      { name: '$select', in: 'query' },
      { name: '$expand', in: 'query' },
    ],
```
And now the same is:
```
    paramDefs: {
      query: ['$top', '$skip', '$search', '$filter', '$count', '$orderby', '$select', '$expand'],
    },
```

This change has a surprisingly large impact on the size of code generated. After building the beta package, looking at the combined size of the .js files, they shrink from 6207kB to 5204kB. That's about 1/6th less code on disk. 

There's no much improvement to the NPM package size, just a few kB, since the removed repetition compresses so well. But the uncompressed footprint makes it worthwhile all the same.

This change won't build until all endpoints are rebuilt, so in the first iteration I included only the changes to the generation script & the graph.call() method and a few updated endpoints - just enough to showcase the change. 

In the second iteration, I regenerated all the endpoints for both beta and v.10 (schema commits 4e63218 and 3bdcc26, both published yesterday).

### DIFF
--- a/packages/client/src/app.spec.ts
+++ b/packages/client/src/app.spec.ts
@@ -491,11 +491,9 @@ describe('App', () => {
       const getPresence = (): {
         method: 'get';
         path: '/me/presence';
-        paramDefs: [];
       } => ({
         method: 'get',
         path: '/me/presence',
-        paramDefs: [],
       });
 
       buildGraphClientSpy.mockReturnValue({ call });


packages/graph-endpoints/src/me/teamwork/associatedTeams/index.ts
```diff
@@ -33,10 +33,10 @@ export function del(
   return {
     method: 'delete',
     path: '/me/teamwork/associatedTeams/{associatedTeamInfo-id}',
-    paramDefs: [
-      { name: 'If-Match', in: 'header' },
-      { name: 'associatedTeamInfo-id', in: 'path' },
-    ],
+    paramDefs: {
```
Jesperholmbergmsft: This file and the new few showcase the effect of the changes. Just rejiggering the paramDefs, grouping by 'in' essentially, results in less code.

packages/graph-tools/src/endpoints.ts
```diff
@@ -194,9 +194,16 @@ class Client {
       const def = schema[method as keyof typeof methods];
       if (!def) continue;
 
-      const params = [...(def.parameters || []), ...(schema.parameters || [])]
+      const parameters = [...(def.parameters || []), ...(schema.parameters || [])]
         .map((param) => this.resolveParameter(param))
-        .filter((p): p is OpenAPIV3.ParameterObject => 'name' in p);
+        .filter((p): p is OpenAPIV3.ParameterObject => 'name' in p)
```
Jesperholmbergmsft: When generating, group the parameters by their 'in' value

packages/graph-tools/templates/function-definition.ts.hbs
```diff
@@ -28,11 +28,15 @@ function {{name}}(
     method: '{{method}}',
     path: '{{url}}',
     {{#if (notEmpty parameters)}}
-    paramDefs: [
+    paramDefs: {
       {{#each parameters}}
-      { name: '{{name}}', in: '{{in}}' },
+      {{@key}}: [
```
Jesperholmbergmsft: When generating the endpoints, loop through the grouped 'in' values and write out the parameter names

packages/graph/src/utils/url.ts
```diff
@@ -2,41 +2,45 @@ import qs from 'qs';
 
 import * as http from '@microsoft/teams.common/http';
 
-import { IParam } from './interfaces';
+import { ParamDefs } from '../types';
 
-export function getInjectedUrl(url: string, params: Array<IParam>, data: Record<string, any>) {
+export function getInjectedUrl(
```
Jesperholmbergmsft: Same results as before, but new input type. Before it'd loop over the params[] array and check the 'in' value for each to figure out what to do. Now it can loop over param.query to find the query string ones, and the params.path to find the path ones.

packages/graph/src/utils/url.ts
```diff
@@ -2,41 +2,45 @@ import qs from 'qs';
...
-      agg[param.name] = data[param.name];
-    }
-    return agg;
-   }, {});
-
-   return Object.keys(paramHeaders).length === 0 
-     ? requestConfig 
-    : {...requestConfig, headers: {...requestConfig?.headers, ...paramHeaders}};
-
+  const paramHeaders = (params.header ?? []).reduce<Record<string, any>>(
```
Jesperholmbergmsft: similar here -- no change in logic or results, but instead of looping over params and checking the 'in' value, this can now reduce directly over params.header.

packages/graph/src/index.spec.ts
```diff
@@ -962,7 +955,7 @@ describe('Client', () => {
             (): EndpointRequest<any> => ({
               method: 'get',
               path: '/users',
-              paramDefs: [],
+              paramDefs: {},
```
heyitsaamir: could prob get rid of this instead of providing empty {}

heyitsaamir (dismissed):

packages/graph/src/index.spec.ts
Jesperholmbergmsft: Agreed - I'll make it so.

PR #346, opened 2025-09-13
python dev deployment

MehakBindra:


packages/cli/configs/atk/basic/python/env/.env.dev
```diff
@@ -0,0 +1,16 @@
+# Built-in environment variables
```
heyitsaamir, line 1: do we need something similar for atk.oauth?

heyitsaamir (approved):

Jesperholmbergmsft (approved):

packages/cli/configs/atk/basic/python/env/.env.dev
MehakBindra: Yes, adding in seperate PR today

PR #350, opened 2025-09-15
Fix connection name handling in token exchange and verify state functions

rido-min:
Update the handling of connection names in the token exchange and verify state functions to use the context's connection name directly, improving consistency and clarity.

### DIFF
--- a/packages/apps/src/app.oauth.ts
+++ b/packages/apps/src/app.oauth.ts
@@ -16,11 +16,11 @@ export async function onTokenExchange<TPlugin extends IPlugin>(
   this: App<TPlugin>,
   ctx: contexts.IActivityContext<ISignInTokenExchangeInvokeActivity, PluginAdditionalContext<TPlugin>>
 ) {
-  const { api, activity, log, next } = ctx;
+  const { api, activity, log, connectionName, next } = ctx;
 
-  if (this.oauth.defaultConnectionName !== activity.value.connectionName) {
+  if (connectionName !== activity.value.connectionName) {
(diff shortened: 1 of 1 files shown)


packages/apps/src/app.oauth.ts
```diff
@@ -16,11 +16,11 @@ export async function onTokenExchange<TPlugin extends IPlugin>(
...
-  const { api, activity, log, next } = ctx;
+  const { api, activity, log, connectionName, next } = ctx;
 
-  if (this.oauth.defaultConnectionName !== activity.value.connectionName) {
+  if (connectionName !== activity.value.connectionName) {
     log.warn(
-      `default connection name "${this.oauth.defaultConnectionName}" does not match activity connection name "${activity.value.connectionName}"`
+      `default connection name "${connectionName}" does not match activity connection name "${activity.value.connectionName}"`
     );
   }
```
heyitsaamir: I think the problem is with binding @rido-min. I think we should still use this.oauth.defaultConnectionName, but i think in `app.ts` this function doesn't bind to `app`, so this.oauth is undefined.

rido-min: yeah, but since we are only using the `connectionName`, and it's already available in the `context` I thought this might be an easier fix. 

Binds are always hard to spot on (imho). Where should we bind? this goes deep into the routing infrastructure. Happy to try it if this fix is not sufficient.

rido-min: outdated

PR #354, opened 2025-09-17
TS Changes for atk.basic in azure and atk.oauth 

MehakBindra:
skip-test-verification

1. Oauth in Typescript - verified in local and dev
2. Basic in typescript Dev env (switched from managed identity back to client id and secret - single tenant) 
For managed identity we would have had to update each template to use azure identity
3.  Remove conversion of port to number as ports are strings in dev env

### DIFF
--- a/packages/cli/configs/atk/basic/python/.vscode/launch.json
+++ b/packages/cli/configs/atk/basic/python/.vscode/launch.json
@@ -49,6 +49,7 @@
         "name": "Start",
         "type": "debugpy",
         "request": "launch",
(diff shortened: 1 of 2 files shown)


heyitsaamir (changes requested):
With this change, we also need to create a Service Principal for the target tenant. As such, the app becomes unusable, unless the user manually does this. ATK has this slated for next week, so we shouldn't do the single-tenant switch just yet.

singhk97 (approved):

PR #364, opened 2025-09-26
DevTools: Fix React useEffect dependency cycle

corinagum:
- Handlers with `useCallback` created a `useEffect` dependency cycle, intermittently causing websocket failures due to re-renders.
- At some point, the main `App.tsx` file got changed to double quotes instead of single quotes. Updated to match the rest of the package.

### DIFF
--- a/packages/devtools/src/App.tsx
+++ b/packages/devtools/src/App.tsx
@@ -1,7 +1,6 @@
 import {
   useEffect,
   useMemo,
-  useCallback,
   useRef,
   useDebugValue,
   useState,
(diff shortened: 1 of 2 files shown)


heyitsaamir (approved):

packages/devtools/src/App.tsx
```diff
@@ -57,97 +56,91 @@ export default function App() {
...
+      socket.off('connect');
+      socket.off('disconnect');
+      socket.off('activity');
+      socket.off('metadata');
       socket.disconnect();
     };
-  }, [handleConnect, handleDisconnect, handleActivity, handleMetadata]);
+  }, [putActivity, onActivity, setMetadata]);
 
   const fluentTheme = useMemo(() => {
```
Jesperholmbergmsft, line 100: nit: this don't look like it needs to be memoized.

packages/devtools/src/App.tsx
```diff
@@ -57,97 +56,91 @@ export default function App() {
...
-      socket.off("activity");
-      socket.off("metadata");
+      socket.off('connect');
+      socket.off('disconnect');
+      socket.off('activity');
+      socket.off('metadata');
       socket.disconnect();
     };
-  }, [handleConnect, handleDisconnect, handleActivity, handleMetadata]);
+  }, [putActivity, onActivity, setMetadata]);
```
Jesperholmbergmsft, line 98: The change here looks safe and all, but I don't understand how it solves the re-renders. Which one of the old callbacks was unstable and caused the effect to run multiple times?

Jesperholmbergmsft (approved):

packages/devtools/src/App.tsx
corinagum: TBH I'm not sure the 'why', but Zustand `useStore` returns function references that keep changing despite being defined outside of the component. So, presumably the selectors are stable because they are outside of `App()`, but in reality they're changing intermittently*. Then, when those references change, `useCallback` causes unneccessary rerenders and the websocket closes.

*intermittently, because I don't run into this every time I run DevTools, and afaik they should be stable. Possible race condition.

Perhaps I'm doing something wrong with [`ActivityStore.ts`](https://github.com/microsoft/teams.ts/blob/main/packages/devtools/src/stores/ActivityStore.ts) and not seeing it? I'll leave the PR open and ping you on Monday in case you see something that I don't.

packages/devtools/src/App.tsx
corinagum: `teamsDarkTheme` and `teamsLightTheme` are expensive objects, which is why I use useMemo. lmk if you want to discuss further :)

packages/devtools/src/App.tsx
corinagum: TL;DR: we had a chat conversation and couldn't determine what may be causing the re-renders. @Jesperholmbergmsft correctly points out that zustand appears to be stable and `useCallback` wouldn't make a difference. I will keep these changes but refer to this later if the websocket close issue pops back up; this removes a dependency but doesn't change behavior

Potential cleanup option: put subscription management in a hook since `App.tsx` is focused on rendering.

PR #365, opened 2025-09-29
fix(apps): add `userToken` to event callback context

RaulPROP:
The attribute `userToken` in the event callback context was always undefined, even if `issignedin` was true.


corinagum (approved):
Thank you @RaulPROP for the catch!

PR #369, opened 2025-10-07
remove "transpile-only" in tsconfig, use tsx

heyitsaamir:
1. Remove nodemon/ts-node deps in favor of [tsx](https://tsx.is). Tsx is much faster than ts-node, but it doesn't do typechecking by default. (We were actually not doing typechecking either because tsconfig was set to transpile-only (which this PR also fixes)). 
2. Switches tsconfig to not be transpile only because users and agents should know when types are wrong, so they can fix those errors proactively. Now, `npm run build` will actually throw if the typescript has errors.

Tested it by creating new apps, and running them. Making sure they pick up the .env files automatically.


packages/cli/templates/typescript/ai/package.json.hbs
```diff
@@ -13,7 +13,7 @@
 "clean": "npx rimraf ./dist",
 "build": "npx tsup",
 "start": "node .",
-"dev": "npx nodemon -w \"./src/**\" -e ts --exec \"node -r ts-node/register -r dotenv/config ./src/index.ts\""
+"dev": "npx tsx watch -r dotenv/config src/index.ts"
```
rido-min: since tsx is already installed, we could avoid npx

Also, tsx allows `--env-file` so we could avoid dotenv

heyitsaamir: removing dotenv deps will takes some testing that i'm avoiding right now since our code internally depends on it. Fair thing to tackle in the future.

Agree not needing npx. Will remove

corinagum (approved):

aacebo (approved):

PR #372, opened 2025-10-10
Update python templates

heyitsaamir:
* Fix pyproject.toml for ai template
* Add "uv run start" capability
* Add "uv run typecheck" capability (with pyright dev-dep)

These are all ways to simplify running the project for users and agents.


packages/cli/templates/python/graph/README.md
```diff
@@ -0,0 +1,14 @@
+# Python Teams AI Bot
```
lilyydu: nit: Python Teams AI Graph Bot

Issue #374, opened 2025-10-13
Streamable content only starts streaming after it's been completely streamed from the model


heyitsaamir:
The whole point of streaming is that the user starts receiving chunks before the whole message has been downloaded. Right now, if you put logs in, you can see that the chunks are fully downloaded, and THEN the stream starts to actually flow to Teams. This is wrong. We need to send chunks out as soon as they come in (barring rate limitations).

Hint:
The main issue is that the timeout gets reset any time a new chunk comes in.

PR #378, opened 2025-10-16
[Fix] streaming: do not reset timeout for each chunk emit

MehakBindra:
Issue : https://github.com/microsoft/teams.ts/issues/374

Main change::
- In emit, instead of cancelling and scheduling a new timeout, 
      - flush immediately if no timeout pending, 
      - else push to queue and wait.
- Added tests for streaming

https://github.com/user-attachments/assets/d8fc0a16-21f9-4ddc-a681-5db5e200e192

Devtools: Streaming always starts before the full response is received (check with logs, see attached video).
Teams: Even though the first chunk is emitted immediately, (sometimes) by the time the stream starts on Teams, the full response is ready. In the second msg, we can see the stream starts a little bit before the full response is printed.

### DIFF
--- a/packages/apps/src/plugins/http/stream.spec.ts
+++ b/packages/apps/src/plugins/http/stream.spec.ts
@@ -38,7 +38,8 @@ describe('HttpStream', () => {
         if (calls <= successAfter) {
           throw new Error('timeout');
         }
-        return { id: `activity-${calls}` };
+
+        return { _activity, id: `activity-${calls}` };
       }
     );
     return () => calls;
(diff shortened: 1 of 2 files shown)


corinagum (approved):

packages/apps/src/plugins/http/stream.ts
```diff
@@ -76,15 +104,23 @@ export class HttpStream implements IStreamer {
...
     }
 
-    while (!this.id || this.queue.length) {
-      await new Promise((resolve) => setTimeout(resolve, 200));
+    // Wait until all queued activities are flushed
+    const start = Date.now();
+
+    while (this.queue.length || !this.id) {
+      if (Date.now() - start > this._totalTimeout) {
+        this._logger.warn('Timeout while waiting for id and queue to flush');
```
heyitsaamir, line 112: shouldn't you break out when this happens?

heyitsaamir: let's also make sure you test this behavior

packages/apps/src/utils/promises/retry.ts
```diff
@@ -27,7 +27,7 @@ export async function retry<T = any>(factory: () => Promise<T>, options?: RetryO
   try {
     return await factory();
   } catch (err) {
-    if (max > 0) {
+    if (max > 1) {
```
heyitsaamir, line 30: why max > 1?

packages/apps/src/plugins/http/stream.spec.ts
```diff
@@ -0,0 +1,129 @@
+import { HttpStream } from './stream';
```
heyitsaamir, line 1: So these are pretty simple test cases. If you want to test anything that tests the edge cases, you'll need to mimic ticks.

Basically:
Emit
Advance Time by x
Emit
Advance Time by x
Assert if activity was sent as expected
Assert if activity contains what is expected (chunk size?)
Emit
Advance Time by y
Assert if activity is not sent

etc.

This is how time-related activities are tested. Checkout this gist that displays a lot of these - https://gist.github.com/apieceofbart/e6dea8d884d29cf88cdb54ef14ddbcc4

Anyway, as rido said, don't sink too much time in this, but something to consider.

packages/apps/src/utils/promises/retry.ts
MehakBindra: max decreases by 1 for each attempt
max = no of attempts 
while max > 1, try again . (then no of tries = max )

packages/apps/src/plugins/http/stream.spec.ts
MehakBindra: updated some tests to use advanceTimersByTimeAsync instead of runAllTimersAsync to test time constraints better

singhk97 (approved):

PR #382, opened 2025-10-27
Revert "[🐞fix] Change AttachmentLayout type to replace 'carousel' with correct value: 'grid'" (And add MessagingExtensionAttachmentLayout)

heyitsaamir:
Reverts microsoft/teams.ts#381

This is a bit of mess - 
* There is an `AttachmentLayout` that exists on `Activity` which should be `list | carousel`.
* There is an `AttachmentLayout` that exists on `MessageExtension` which should be `list | grid`.

### DIFF
--- a/packages/api/src/models/messaging-extension/index.ts
+++ b/packages/api/src/models/messaging-extension/index.ts
@@ -6,4 +6,4 @@ export * from './messaging-extension-parameter';
 export * from './messaging-extension-query';
 export * from './messaging-extension-action';
 export * from './messaging-extension-action-response';
-export * from './message-extension-attachment-layout';
+export * from './messaging-extension-attachment-layout';


corinagum (dismissed):

packages/api/src/models/messaging-extension/messaging-extension-result.ts
```diff
@@ -1,6 +1,6 @@
 import { Activity } from '../../activities';
-import { AttachmentLayout } from '../attachment';
 
+import { MessageExtensionAttachmentLayout } from './message-extension-attachment-layout';
```
aacebo: nit: would be nice to either keep the `MessagingExtension*` or pivot completely to the `MessageExtension*` name to stay consistent with our type names.

aacebo (dismissed):

packages/api/src/models/messaging-extension/messaging-extension-result.ts
heyitsaamir: Good call. Updating to MessagingExtension

aacebo (approved):

Issue #367, opened 2025-10-07
Support of regional Azure bots


Benjiiim:
I'm not sure if I should open this one here or in the main teams-ai repo. Don't hesitate to tell me if I should do something else.

In Teams AI Library v1, we were able to use the ConfigurationBotFrameworkAuthenticationOptions to pass a regional setting for OAuthApiEndpoint and OAuthUrl (example: https://europe.api.botframework.com).
I don't find any way to do something similar with Teams AI Library v2 and that seems to block the use of regional Azure bots.

Is it possible to use regional Azure bots with Teams AI Library v2 today? If not, is it something planned?

lilyydu:
@Benjiiim,

Looked into it more, and we do support regional bots through customizing the [SignInOptions](https://github.com/microsoft/teams.ts/blob/main/packages/apps/src/contexts/activity.ts#L114) - an example of the `signInLink` (aka your `config.oathUrl`) would be https://europe.token.botframework.com/.auth/web/redirect.

I'm still unsure of the OauthApiEndpoint value +necessity as your OAuthApiEndpoint and OAuthUrl seem to have the same value, but the botbuilder link you linked above is referencing the general API endpoint (https://europe.api.botframework.com). I will try and validate this myself as well

PR #386, opened 2025-11-06
[apps/api] feat: allow oauth for regional bots

lilyydu:
resolves: #367

- exposed a new config option called `ClientSettings` thats passed into our clients from the app layer
- the `tokenUrl` specifies the regional token endpoint to use (e.g., europe.token.botframework)

I didn't reuse `OauthSettings` because 
a) its associated w/ the Apps ecosystem vs `ClientSettings` is concerned w the API layer 
b) eventually may diverge with additional params
b) would cause circular dependencies 

- tested with the graph sample, works w/ global and a west-europe regional bot
- add tests

**INSTRUCTIONS FOR SETTING UP** 
(documenting here for now, should add to docs, partial courtesy to @Benjiiim):
- in azurebot.bicep, replace all 'global' occurrences by 'westeurope'
- in manifest.json, *.botframework.com should be replaced by europe.token.botframework.com (or equivalent for other locations) in validDomains
- in aad.manifest.json, replace "https://token.botframework.com/.auth/web/redirect" by "https://europe.token.botframework.com/.auth/web/redirect"
- in index.ts, update OauthSettings to incl clientSettings

> const app = new App({
  oauth: {
    defaultConnectionName: 'graph',
    // This is an example of overriding the token URL for a specific region (e.g., Europe).
    // Uncomment this block if needed.
    // clientSettings: {
    //   tokenUrl: "https://europe.token.botframework.com",
    // }
  },
  logger: new ConsoleLogger('@tests/auth', { level: 'debug' })
});

resources:
- https://learn.microsoft.com/en-us/azure/bot-service/bot-builder-concept-regionalization?view=azure-bot-service-4.0
- https://learn.microsoft.com/en-us/azure/bot-service/ref-oauth-redirect-urls?view=azure-bot-service-4.0
- https://learn.microsoft.com/en-us/azure/bot-service/bot-builder-authentication?view=azure-bot-service-4.0&tabs=multitenant%2Caadv2%2Ccsharp#prepare-the-bot-code

### DIFF
--- a/packages/api/src/clients/api-client-settings.ts
+++ b/packages/api/src/clients/api-client-settings.ts
@@ -1,4 +1,4 @@
-export type ClientSettings = {
+export type ApiClientSettings = {
   /**
    * the URL to use for managing user oauth tokens.
    * Specify this value if you are using a regional bot.
@@ -8,19 +8,19 @@ export type ClientSettings = {
   readonly oauthUrl: string;
 };
 
-export const DEFAULT_CLIENT_SETTINGS: ClientSettings = {
+export const DEFAULT_API_CLIENT_SETTINGS: ApiClientSettings = {
   oauthUrl: 'https://token.botframework.com',
 };
 
-export function mergeClientSettings(
-  clientSettings?: Partial<ClientSettings>
-): ClientSettings {
+export function mergeApiClientSettings(
+  apiClientSettings?: Partial<ApiClientSettings>
+): ApiClientSettings {
   const env = typeof process === 'undefined' ? undefined : process.env;
   
   return {
     oauthUrl: 
-      clientSettings?.oauthUrl ?? 
+      apiClientSettings?.oauthUrl ?? 
       env?.OAUTH_URL ?? 
-      DEFAULT_CLIENT_SETTINGS.oauthUrl,
+      DEFAULT_API_CLIENT_SETTINGS.oauthUrl,
   };
 }
--- a/packages/api/src/clients/bot/sign-in.ts
+++ b/packages/api/src/clients/bot/sign-in.ts
@@ -3,7 +3,7 @@ import qs from 'qs';
 import { Client, ClientOptions } from '@microsoft/teams.common/http';
 
 import { SignInUrlResponse } from '../../models';
-import { ClientSettings, mergeClientSettings } from '../client-settings';
+import { ApiClientSettings, mergeApiClientSettings } from '../api-client-settings';
 
 export const BOT_SIGNIN_ENDPOINTS = {
   URL: 'api/botsignin/GetSignInUrl',
@@ -32,9 +32,9 @@ export class BotSignInClient {
     this._http = v;
   }
   protected _http: Client;
-  protected _clientSettings: Partial<ClientSettings>;
+  protected _apiClientSettings: Partial<ApiClientSettings>;
 
-  constructor(options?: Client | ClientOptions, clientSettings?: Partial<ClientSettings>) {
+  constructor(options?: Client | ClientOptions, apiClientSettings?: Partial<ApiClientSettings>) {
     if (!options) {
       this._http = new Client();
     } else if ('request' in options) {
@@ -42,13 +42,13 @@ export class BotSignInClient {
     } else {
       this._http = new Client(options);
     }
-    this._clientSettings = mergeClientSettings(clientSettings);
+    this._apiClientSettings = mergeApiClientSettings(apiClientSettings);
   }
 
   async getUrl(params: GetBotSignInUrlParams) {
     const q = qs.stringify(params);
     const res = await this.http.get<string>(
-      `${this._clientSettings.oauthUrl}/${BOT_SIGNIN_ENDPOINTS.URL}?${q}`
+      `${this._apiClientSettings.oauthUrl}/${BOT_SIGNIN_ENDPOINTS.URL}?${q}`
     );
 
     return res.data;
@@ -57,7 +57,7 @@ export class BotSignInClient {
   async getResource(params: GetBotSignInResourceParams) {
     const q = qs.stringify(params);
     const res = await this.http.get<SignInUrlResponse>(
-      `${this._clientSettings.oauthUrl}/${BOT_SIGNIN_ENDPOINTS.RESOURCE}?${q}`
+      `${this._apiClientSettings.oauthUrl}/${BOT_SIGNIN_ENDPOINTS.RESOURCE}?${q}`
     );
 
     return res.data;
--- a/packages/api/src/clients/index.ts
+++ b/packages/api/src/clients/index.ts
@@ -1,7 +1,7 @@
 import * as http from '@microsoft/teams.common/http';
 
+import { ApiClientSettings, mergeApiClientSettings } from './api-client-settings';
 import { BotClient } from './bot';
-import { ClientSettings, mergeClientSettings } from './client-settings';
 import { ConversationClient } from './conversation';
 import { MeetingClient } from './meeting';
 import { TeamClient } from './team';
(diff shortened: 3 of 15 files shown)


packages/api/src/clients/client-settings.ts
```diff
@@ -0,0 +1,13 @@
+export type ClientSettings = {
```
Jesperholmbergmsft: I lik this ClientSettings construct. A while back I was thinking on how to support [national clouds](https://learn.microsoft.com/en-us/graph/deployments) for the grpah client and wasn't sure how to plumb it properly, but this will extend nicely to that too!

packages/api/src/clients/bot/sign-in.ts
```diff
@@ -26,21 +27,23 @@ export class BotSignInClient {
...
-  constructor(options?: Client | ClientOptions) {
+  constructor(options?: Client | ClientOptions, clientSettings?: ClientSettings) {
     if (!options) {
       this._http = new Client();
     } else if ('request' in options) {
       this._http = options;
     } else {
       this._http = new Client(options);
     }
+    this._clientSettings = clientSettings ?? DEFAULT_CLIENT_SETTINGS;
```
Jesperholmbergmsft: In the clientSettings, tokenUrl is typed to be optional. Which is good, as it'll help extend to more type of settings, letting the developer override whatever they like.
That means though, that even if `clientSettings` is defined, tokenUrl might not be. In that case we should probably still fall back on the default value. One approach could be to merge the defaults & provided client settings here, e.g.
```suggestion
    this._clientSettings = {...DEFAULT_CLIENT_SETTINGS, ...(clientSettings ?? {})};
```

packages/api/src/clients/bot/sign-in.ts
```diff
@@ -26,21 +27,23 @@ export class BotSignInClient {
...
   }
 
   async getUrl(params: GetBotSignInUrlParams) {
     const q = qs.stringify(params);
     const res = await this.http.get<string>(
-      `https://token.botframework.com/api/botsignin/GetSignInUrl?${q}`
+      `${this._clientSettings.tokenUrl}/api/botsignin/GetSignInUrl?${q}`
```
corinagum: similar to Aamir's MSAL changes, do we want to consider consts to avoid non-magic strings for the latter part of the token urls? (applicable to below as well)

packages/api/src/clients/bot/sign-in.ts
corinagum: This seems like a good guard for future extensibility

Jesperholmbergmsft: I agree with Corina.

lilyydu: Oo good catch Jesper! I like this more - will update :)

packages/api/src/clients/client-settings.ts
```diff
@@ -0,0 +1,13 @@
...
+  /**
+   * the URL to use for managing user oauth tokens.
+   * Specify this value if you are using a regional bot.
+   * For e.g., https://europe.token.botframework.com
+   * Default is https://token.botframework.com
+   */
+  readonly OAuthUrl?: string;
+};
```
heyitsaamir: camelCase?

packages/api/src/clients/client-settings.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const DEFAULT_CLIENT_SETTINGS: Required<ClientSettings> = {
```
heyitsaamir: NIT: Instead of making all keys optional in ClientSettings, and making DEFAULT_CLIENT_SETTINGS `Required`, you could make all keys required in ClientSettings, and make all parameters `Partial`. This way, in places you need `ClientSettings`, the value of the keys isn't string | undefined.
For eg, in sign-in.ts, you have:
```ts
`${this._clientSettings.OAuthUrl}/${BOT_SIGNIN_ENDPOINTS.URL}?${q}`
```
here OauthUrl is actually string | undefined, which is less understandable than if it was just string.

packages/api/src/clients/client-settings.ts
```diff
@@ -0,0 +1,13 @@
+export type ClientSettings = {
+  /**
+   * the URL to use for managing user oauth tokens.
+   * Specify this value if you are using a regional bot.
+   * For e.g., https://europe.token.botframework.com
+   * Default is https://token.botframework.com
+   */
+  readonly OAuthUrl?: string;
```
heyitsaamir: Also oauthUrl, or oauthBaseEndpoint? It's not really the full url. Thoughts?

packages/api/src/clients/index.ts
```diff
@@ -44,8 +46,10 @@ export class Client {
       });
     }
 
-    this.bots = new BotClient(this.http);
-    this.users = new UserClient(this.http);
+    this._clientSettings = {...DEFAULT_CLIENT_SETTINGS, ...(clientSettings ?? {})};
```
heyitsaamir: this is copied in many places. Consider a helper method that takes in `Partial<ClientSettings>` and returns `ClientSettings`

packages/apps/src/oauth.ts
```diff
@@ -1,12 +1,25 @@
...
    * @default `graph`
    */
   readonly defaultConnectionName?: string;
+
+  /**
+   * the client settings to use for
+   * authentication. This is important to
+   * configure for regional bots.
+   */
+  readonly clientSettings?: ClientSettings
```
heyitsaamir: why is this in "OauthSettings". this seems more general than just Oauth, right?

packages/api/src/clients/client-settings.ts
heyitsaamir: Also, each of these should have an associated environment variable that populates them through a config, if not provided

lilyydu: I prefer sticking with `oauthUrl` (updated casing) even though internals can differ - devs can map it easily from here https://learn.microsoft.com/en-us/azure/bot-service/ref-oauth-redirect-urls?view=azure-bot-service-4.0

packages/apps/src/oauth.ts
lilyydu: no, this is specifically for oauth

packages/api/src/clients/api-client-settings.ts
```diff
@@ -0,0 +1,26 @@
...
+};
+
+export const DEFAULT_CLIENT_SETTINGS: ClientSettings = {
+  oauthUrl: 'https://token.botframework.com',
+};
+
+export function mergeClientSettings(
+  clientSettings?: Partial<ClientSettings>
+): ClientSettings {
+  const env = typeof process === 'undefined' ? undefined : process.env;
```
lilyydu, line 18: @Benjiiim responding here for threading, thanks for these suggestions!

- updated the endpoint, agree on matching with these docs
- removed the comment on resourceGroup().location and added in the manifest.json (these will go into official docs, just had them in this PR notes for ref)
- removed the code change to the template, let's capture this in the official docs instead

packages/apps/src/oauth.ts
heyitsaamir: > no, this is specifically for oauth

Then we shouldn't call it "ClientSettings" since that implies it's for all clients, not just for oauth. 

Why not keep it ClientSettings, and pass it to all Clients?

lilyydu: fair, how do you feel about `OauthClientSettings`?

heyitsaamir: That's fine, but why not just ClientSettings? This way you could pass additional hard coded client settings wherever they are necessary?

heyitsaamir: Now you'll need to have a different option for each type of client

lilyydu: (resolved offline to rename to ApiClientSettings)

PR #387, opened 2025-11-06
Introduce MSAL with UMI and FIC support

heyitsaamir:
This PR introduces MSAL as a dependency and switches our supported bot authentication to use it. It also includes support for User Managed Identity support and FIC such that we don't need to provide custom code to do that anymore as we had previously suggested. Instead, we can do that simply by the environment variables.

| CLIENT_ID | CLIENT_SECRET | MANAGED_IDENTITY_CLIENT_ID | Output |
|-|-|-|-|
| not_set | | | No-Auth |
| set | set | | SecretsAuth |
| set | not_set | | User Managed Identity Auth |
| set | not_set | set (same as CLIENT_ID) | User Managed Identity Auth |
| set | not_set | set (diff from CLIENT_ID) | FIC (user managed identity) |
| set | not_set | "system" | FIC (system identity) |

Changes to note:
1. We are now no longer _proactively_ refreshing tokens. Instead, we put the token fetcher inside the token resolver logic for the client. This ends up being a call to `tokenManager.getBotToken`.
2. This fixes a bug where proactive scenarios were going to fail because we weren't refreshing the token for them. For reactive scenarios, we refresh the token when the request comes in, so it was fine.
3. We are deprecating the `id` and `name` fields because those used to depend on the token being present. Now that we are not proactively fetching the token, those are not as useful (or reliable). For now, I have chosen to simply use the value from the manifest to populate them. I don't think these fields were that useful anyway, so opting to deprecate.
4. Removes graphToken from being injected into plugins. Plugins weren't using it, and we want to remove tight dependency on graph package.

#### PR Dependency Tree

* **PR #387** 👈
* Merged #388, #389, #392


rido-min (dismissed):

packages/api/src/clients/bot/token.ts
```diff
@@ -75,6 +79,10 @@ export class BotTokenClient {
       };
     }
 
+    if (!('clientSecret' in credentials)) {
+      throw new Error('Bot Token Client only supports auth via secrets');
```
rido-min: why is this limitation? the user token service does not care about the credentials used to mint the token.

heyitsaamir: talked offline. this is for the old way of minting tokens using Bot TokenClient

PR #391, opened 2025-11-07
Updated Graph Endpoints from latest schema (commit 08ecd2f)

Jesperholmbergmsft:
This updates the Graph Beta Endpoints schema from the latest published schema. 

I added the schema commit hash to the readme that gets published to NPM. Besides that, all code changes in this PR were auto-generated from the existing scripts.


heyitsaamir (approved):
lgtm.

We should really think about pulling this package out so that it can live alongside the schema and be up-to-date

PR #394, opened 2025-11-11
[api/apps]: refactored ApiClientSettings to AppOptions

lilyydu:


packages/apps/src/app.ts
```diff
@@ -87,6 +87,12 @@ export type AppOptions<TPlugin extends IPlugin> = Partial<Credentials> & {
    * Skip authentication for HTTP requests
    */
   readonly skipAuth?: boolean;
+
```
heyitsaamir: NIT: rm extra line

packages/apps/src/app.ts
```diff
@@ -87,6 +87,12 @@ export type AppOptions<TPlugin extends IPlugin> = Partial<Credentials> & {
...
+  
+  /**
+   * API client settings to use.
```
heyitsaamir: A bit more detail? More like "overrides"?

heyitsaamir (dismissed):

packages/apps/src/app.ts
```diff
@@ -87,6 +87,12 @@ export type AppOptions<TPlugin extends IPlugin> = Partial<Credentials> & {
    * Skip authentication for HTTP requests
    */
   readonly skipAuth?: boolean;
+  
+  /**
+   * API client settings used for overriding.
+   * If you have a regional bot, specify your oauthUrl here.
```
heyitsaamir: Don't put something specific to oauthurl here, because you'll need to change it anytime ApiClientSettings changes. Just say something like add any overrides for clients with this option.

heyitsaamir: First line is probably good enough

heyitsaamir (approved):

PR #396, opened 2025-11-12
Remove short circuit in `app.$process()` if no routes found.

### LINKED ISSUE #393: Request hangs indefinitely on activities without handlers defined
How to reproduce:
```ts
const app = new App({
    skipAuth: true,  // not needed, just for easier local testing
});

app.on("message", async ({ send }) => {
    await send("Hello there!");
});
app.start();
```

With the above setup, activities of type message will be handled as expected by the handler we defined via `app.on`. However, for any other kind of activity (like, for example, `installationUpdate`) the request will hang indefinitely.
I believe the cause is here: https://github.com/microsoft/teams.ts/blob/13334d6472ab770f3a34119f829d9fb9a687be95/packages/apps/src/app.process.ts#L133-L135
By exiting `process` early, there's never a chance for `onActivityResponse` to be called here:
https://github.com/microsoft/teams.ts/blob/13334d6472ab770f3a34119f829d9fb9a687be95/packages/apps/src/app.process.ts#L175-L180
And without `onActivityResponse` being called, the `HttpPlugin` never sends a response, leading the request to hang indefinitely, which I don't think was desired.

singhk97:
fixes #393 

The returned `{ status: 200 }` from `app.$process` is not handled by the plugin, and so the http response is not set causing the server to "hang". By not short circuiting, the program execution continues leading to `onActivityResponse` event being called and the HTTP response being set to default status of 200.

- [x] **Is it an issue in .NET SDK?** Nope, there is no short-circuiting being done like this.
- [x] **Is it an issue in Python SDK?** Yes, it should be fixed: https://github.com/microsoft/teams.py/pull/218


singhk97:
Since #398 is merged, this PR should be good to merge.

PR #398, opened 2025-11-13
Simplify setting http response after activity processing

singhk97:
Ensure that HTTP responses are returned directly from handlers, rather than through event-based callbacks.

**Activity processing and response flow:**

* The `onActivity` handler in `app.events.ts` now returns a `Promise<InvokeResponse>` and directly returns the result of processing the activity, rather than just awaiting it.
* The `$process` function in `app.process.ts` is refactored to always return an `InvokeResponse`, handling both normal and error cases, and ensuring response construction is consistent.

**HttpPlugin interface and implementation updates:**

* The `$onActivity` event handler in `HttpPlugin` is now asynchronous and returns a `Promise<InvokeResponse>`, aligning with the new app contract.
* The HTTP request handler in `HttpPlugin` now awaits the activity response and sends it directly to the client, removing the previous pattern of storing pending responses and using callback events.
* Old error and activity response event handlers (`onError`, `onActivityResponse`) and related state management are removed from `HttpPlugin`, as response handling is now synchronous with the HTTP request.

**Type and import consistency:**

* The `InvokeResponse` type is imported and used consistently across files to improve type safety and clarity.

### DIFF
--- a/examples/message-extensions/src/card.ts
+++ b/examples/message-extensions/src/card.ts
@@ -1,4 +1,4 @@
-import { Account, Message, ThumbnailCard } from '@microsoft/teams.api';
+import { Account, Message, TeamsChannelAccount, ThumbnailCard } from '@microsoft/teams.api';
 import {
   ActionSet,
   AdaptiveCard,
@@ -108,7 +108,7 @@ export function createMessageDetailsCard(messagePayload: Message) {
   return new AdaptiveCard(...cardElements);
 }
 
-export function createConversationMembersCard(members: Account[]) {
+export function createConversationMembersCard(members: (Account | TeamsChannelAccount)[]) {
   const membersList = members.map((member) => member.name).join(', ');
 
   return new AdaptiveCard(
@@ -157,7 +157,7 @@ export async function createDummyCards(searchQuery: string) {
         // When a user clicks on a list item in Teams:
         // - If the thumbnail has a `tap` property: Teams will trigger the `message.ext.select-item` activity
         // - If no `tap` property: Teams will insert the full adaptive card into the compose box
-        // tap: { 
+        // tap: {
         //   type: "invoke",
         //   title: item.title,
         //   value: {
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "examples/*"
       ],
       "devDependencies": {
-        "@changesets/cli": "*",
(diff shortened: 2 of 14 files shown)


heyitsaamir (commented):
Makes sense to me. Add tests?

packages/apps/src/app.events.ts
```diff
@@ -39,9 +40,9 @@ export async function onActivity<TPlugin extends IPlugin>(
   this: App<TPlugin>,
   sender: ISender,
   event: IActivityEvent
-) {
+): Promise<InvokeResponse> {
   this.events.emit('activity', event);
-  await this.process(sender, { ...event, sender });
+  return await this.process(sender, { ...event, sender });
```
heyitsaamir: Looks good to me. 

One thing I think we should do is have an `onActivity` in apps (core, not app.events). That should call emit's onActivity, and then call process from there.

singhk97: One issue with that is that other event handler methods will be here... do we want to have that discrepancy? Let's plan that refactor for later?

heyitsaamir: Right, but now process is no longer tied to an event right?

singhk97: https://github.com/microsoft/teams.ts/blob/ee110bafc226ca8dca69ca518690510eaa81d86a/packages/apps/src/app.ts#L507
the process is not really tied to an event (it's tied to the event manager) and the App's `OnActivity` implementation is just ^^.

heyitsaamir: right that's what i mean. Why does `onActivity` needs to be in events anymore? It can be in the main app file. The flow of data is weird right now, app.events is powering onActivity for no clear reason

singhk97: It's because the responsibility of the app.events is to handle events, and `OnActivity` is one of them. Moving it to app will split the responsibility of managing events between app and app.events.

heyitsaamir: it's responsible for emitting events, not really handling them. 
"process" is part of handling an event that's happening. I think what makes sense is:

Http plugin gets the http event -> forwards the event to app -> app sends the event to be processed -> emits it out/

wdyt?

singhk97: moved

Issue #408, opened 2025-11-25
[Bug]: type definition for `Account` is either wrong or `context.api.conversations.members.get` has wrong return type


Mutantpenguin:
### Bug Description

The type definition for `Account` (in `node_modules\@microsoft\teams.api\dist\models\account.d.ts`) is either wrong or `context.api.conversations.members.get` has the wrong return type.

`Account` is defined as:
```ts
type Account<P = any> = {
    readonly id: string;
    readonly aadObjectId?: string;
    readonly role: Role;
    readonly name: string;
    readonly properties?: P;
    readonly membershipSources?: MembershipSource[];
};
```

but during runtime, `context.api.conversations.members.get` returns this (in an array):
```ts
    email
    givenName
    id
    name
    objectId
    surname
    tenantId
    userPrincipalName
    userRole
```

### Steps to Reproduce

```ts
app.on("install.add", async (context) => {
  const members = await context.api.conversations.members(context.activity.conversation.id).get();

  const user = members[0];
  await context.send(new MessageActivity(`Hello ${user.givenName}, thanks for adding me! 🖐️`).addAiGenerated());
});
```

### Expected Behavior

`user.givenName` is not an error

### Actual Behavior

`user.givenName` is an error

### SDK Version

2.0.4

### Node.js Version

22.21.1

PR #409, opened 2025-11-27
TeamsChannelAccount required for certain APIs

MehakBindra:
fixes https://github.com/microsoft/teams.ts/issues/408
Teams uses two different account schemes depending on the API, this PR updates the type definitions so that  conversation.members and meetingInfo/meetingParticipant APIs return `TeamsChannelAccount` instead of the generic `Account`

### DIFF
--- a/packages/api/src/models/account.ts
+++ b/packages/api/src/models/account.ts
@@ -10,16 +10,60 @@ export type Account<P = any> = {
   readonly membershipSources?: MembershipSource[];
 };
 
+/**
+ * Represents a Teams channel account, extending the basic channel account with Teams-specific properties.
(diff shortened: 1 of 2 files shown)


packages/api/src/models/account.ts
```diff
@@ -10,6 +10,19 @@ export type Account<P = any> = {
   readonly membershipSources?: MembershipSource[];
 };
 
+export type TeamsChannelAccount<P = any> = {
```
heyitsaamir, line 18: @copilot can you please add documentation? 
https://learn.microsoft.com/en-us/dotnet/api/microsoft.bot.schema.teams.teamschannelaccount?view=botbuilder-dotnet-stable

PR #414, opened 2025-12-04
Add azure pipelines

heyitsaamir:
Also fixes some leftover bluehawk snippets scripts, and runs an npm audit.

### DIFF
--- a/.azdo/azure-pipelines.yml
+++ b/.azdo/azure-pipelines.yml
@@ -47,6 +47,9 @@ steps:
   - script: npm run build
     displayName: 'Build'
 
+  - script: npm run test
+    displayName: 'Test'
+
   - script: |
       mkdir -p $(Build.ArtifactStagingDirectory)/packages
       for dir in packages/* external/*; do
@@ -63,9 +66,3 @@ steps:
       artifactName: 'npm-packages'
       publishLocation: 'Container'
     displayName: 'Publish package artifacts'
-
-  - script: npm run test
-    displayName: 'Test'
-
-  - script: npm run validate:snippets
-    displayName: 'Validate Snippets'


azdo/azure-pipelines.yml
```diff
@@ -0,0 +1,49 @@
...
+      - '*'
+
+# Manual trigger support
+schedules: []
+
+pool:
+  vmImage: 'ubuntu-latest'
+
+strategy:
+  matrix:
```
rido-min: what's the reason to build with multiple node versions? this might be useful for e2e tests but for build/pack I think it's ok to just build with the latest node version

azdo/azure-pipelines.yml
rido-min, line 1: nit. can you name the folder `.azdo` ?

azdo/azure-pipelines.yml
```diff
@@ -0,0 +1,49 @@
...
+  - script: npm install
+    displayName: 'Install Dependencies'
+
+  - script: npm run lint
+    displayName: 'Lint'
+
+  - script: npm run build
+    displayName: 'Build'
+
+  - script: npm run test
```
rido-min: missing `npm pack` and push to an internal feed

azdo/azure-pipelines.yml
```diff
@@ -0,0 +1,49 @@
...
+steps:
+  - checkout: self
+    clean: true
+
+  - task: NodeTool@0
+    displayName: 'Use Node.js $(node_version)'
+    inputs:
+      versionSpec: '$(node_version)'
+
+  - script: npm install
```
rido-min: we should install from a Central Feed in ADO, see [this](https://github.com/microsoft/Agents-for-js/blob/926091f9f2cdfc6094867d804d568a5c69d15350/.azdo/ci-pr.yaml#L14)

.azdo/azure-pipelines.yml
rido-min, line 1: seems this file is a dupe

external/mcpclient/src/mcp-client-plugin.spec.ts
rido-min, line 1: is this file related to azdo pipeline?

package.json
```diff
@@ -36,13 +36,11 @@
...
-    "test": "npx turbo test --concurrency=100%",
-    "validate:snippets": "node ./scripts/validate-snippets.js"
+    "test": "npx turbo test --concurrency=100%"
   },
   "devDependencies": {
     "@changesets/cli": "latest",
     "@turbo/gen": "^2.5.7",
-    "bluehawk": "^1.6.0",
     "turbo": "^2.4.0",
     "typescript": "^5.4.5"
```
rido-min, line 45: I would use this PR to use latest tsc

azdo/azure-pipelines.yml
heyitsaamir: fair!

external/mcpclient/src/mcp-client-plugin.spec.ts
heyitsaamir: Split this PR up so this one is just related to azdo stuff.

package.json
heyitsaamir: Let's do build improvements separately

.azdo/azure-pipelines.yml
```diff
@@ -0,0 +1,68 @@
...
+
+  - task: npmAuthenticate@0
+    inputs:
+      workingFile: .npmrc
+
+  - task: Npm@1
+    inputs:
+      command: ci
+      customRegistry: useFeed
+      customFeed: OneCRM/DPX-Tools-Upstream
```
rido-min, line 39: we should use our one feed, not this one

.azdo/azure-pipelines.yml
```diff
@@ -0,0 +1,68 @@
...
+
+  - script: npm run build
+    displayName: 'Build'
+
+  - script: npm run test
+    displayName: 'Test'
+
+  - script: |
+      mkdir -p $(Build.ArtifactStagingDirectory)/packages
+      for dir in packages/* external/*; do
```
rido-min, line 55: nit: why do we have a tree here? can we flat the source tree under packages?

Also, I'd suggest to use npm workspaces, so we can run single commands from the root

PR #422, opened 2025-12-10
Move from teamsapp -> m365

heyitsaamir:


heyitsaamir:
No more ATK support, or CLI. Removing

PR #424, opened 2025-12-16
Simplify HTTP Plugin architecture

heyitsaamir:
Separate activity sending from HTTP transport layer

The previous architecture tightly coupled HTTP transport concerns with activity sending logic:

**Previous Architecture:**
```
HttpPlugin (transport) → implements ISender (sending)
                      → has send() method (creates new Client per call)
                      → has createStream() method
                      → knows about Activity protocol details

ActivityContext → depends on ISender plugin
               → cannot work without transport plugin
               → conflates transport and sending concerns
```

There are a few issues with this:
- HttpPlugin created NEW Client instances on every send() call. So there's really no benefit of this logic being in the "httpclient" plugin.
- Transport plugins (HttpPlugin) were forced to implement send/createStream. This makes it more cumbersome to build your own HttpPlugin with your own servier.
- Users couldn't "bring their own server" without implementing ISender
- ActivityContext was tightly coupled to plugin architecture. ("Sender" was coupled with an activity, without any necessary benefits.)

## New Architecture

```
HttpPlugin (transport) → only handles HTTP server/routing/auth
                      → emits ICoreActivity (minimal protocol knowledge)
                      → just passes body payload to app

ActivitySender (NEW)  → dedicated class for sending activities
                     → receives injected, reusable Client
                     → handles all send/stream logic
                     → private to App class

ActivityContext       → uses ActivitySender now (which is not a plugin) 
```

In this PR, I am mainly decoupling responsibilities of HttpPlugin from being BOTH a listener AND a sender, to being just a listener. The sender bit is now separated to a different `ActivitySender` class. Other than better code organization, the main thing this lets us do is **not require the app to run to be able to send proactive messages**. This is a huge plus point because now the App can be used in scenarios where it doesn't necessarily need to _listen_ to incoming messages (like agentic notifications!)

### 1. Created ActivitySender Class
- Centralized all activity sending logic
- Receives reusable Client in constructor (no per-send instantiation)
- Private to App class - internal implementation detail
- Provides send() and createStream() methods
- **Separate from HttpPlugin**

### 2. Introduced ICoreActivity Interface
- Minimal fields transport layer needs: serviceUrl, id, type
- Extensible via [key: string]: any for protocol-specific fields
- Transport plugins work with this instead of full Activity type. So it's easier to create these.
- Parsing to Activity happens in app.process.ts now, NOT in HttpPlugin.

### 3. Removed ISender Interface
- No longer needed - plugins don't send activities
- Plugins only handle transport (receiving requests)
- Breaking change, but simplifies plugin architecture. This pattern wasn't documented (intentionally) because the design was subject to change. So it should be okay hopefully to change this.

### For Plugin Authors:
1. **ISender removed** - Custom plugins should implement IPlugin only
2. **IActivityEvent changed** - Now has body: ICoreActivity instead of activity: Activity

#### PR Dependency Tree

* **PR #424** 👈
  * **PR #433**

### DIFF
--- a/examples/proactive-messaging/package.json
+++ b/examples/proactive-messaging/package.json
@@ -18,13 +18,13 @@
     "dev": "tsx -r dotenv/config src/index.ts"
   },
   "dependencies": {
-    "@microsoft/teams.api": "2.0.5",
-    "@microsoft/teams.apps": "2.0.5",
-    "@microsoft/teams.cards": "2.0.5",
-    "@microsoft/teams.common": "2.0.5"
+    "@microsoft/teams.api": "*",
+    "@microsoft/teams.apps": "*",
+    "@microsoft/teams.cards": "*",
+    "@microsoft/teams.common": "*"
   },
   "devDependencies": {
-    "@microsoft/teams.config": "2.0.5",
+    "@microsoft/teams.config": "*",
     "@types/node": "^22.5.4",
     "dotenv": "^16.4.5",
     "rimraf": "^6.0.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -283,6 +283,25 @@
         "typescript": "^5.4.5"
       }
     },
+    "examples/proactive-messaging": {
+      "name": "@examples/proactive-messaging",
+      "version": "0.0.6",
+      "license": "MIT",
+      "dependencies": {
+        "@microsoft/teams.api": "*",
+        "@microsoft/teams.apps": "*",
+        "@microsoft/teams.cards": "*",
+        "@microsoft/teams.common": "*"
+      },
+      "devDependencies": {
+        "@microsoft/teams.config": "*",
+        "@types/node": "^22.5.4",
+        "dotenv": "^16.4.5",
+        "rimraf": "^6.0.1",
+        "tsx": "^4.20.6",
+        "typescript": "^5.4.5"
+      }
+    },
     "examples/reactions": {
       "name": "@examples/reactions",
       "version": "0.0.1",
@@ -1951,6 +1970,10 @@
     },
     "node_modules/@examples/message-extensions": {
       "resolved": "examples/message-extensions",
+      "link": true
+    },
+    "node_modules/@examples/proactive-messaging": {
+      "resolved": "examples/proactive-messaging",
       "link": true
     },
     "node_modules/@examples/reactions": {
@@ -10028,6 +10051,8 @@
     },
     "node_modules/env-cmd": {
       "version": "11.0.0",
+      "resolved": "https://registry.npmjs.org/env-cmd/-/env-cmd-11.0.0.tgz",
+      "integrity": "sha512-gnG7H1PlwPqsGhFJNTv68lsDGyQdK+U9DwLVitcj1+wGq7LeOBgUzZd2puZ710bHcH9NfNeGWe2sbw7pdvAqDw==",
       "dev": true,
       "license": "MIT",
       "dependencies": {


packages/apps/src/events/activity.ts
```diff
@@ -1,24 +1,45 @@
...
+   * Activity ID for correlation
+   */
+  id?: string;
+
   /**
-   * the sender of the event
+   * Activity type for basic routing
    */
-  sender: ISender;
+  type?: string;
```
lilyydu, line 20: instead of string, should we be more specific here? or wait till it passes into app.process

heyitsaamir: yeah, similar to the core in dotnet, this layer should be as unopinionated as possible. So ideally it doesn't really even know about Activity

packages/apps/src/types/event.ts
```diff
@@ -1,12 +1,5 @@
...
  * some event emitted from
  * either the App or a Plugin
  */
-export interface IEvent {
-  /**
-   * the sender of the event
-   */
-  sender?: IPlugin;
-}
+export interface IEvent {}
```
aacebo: empty interface ?

heyitsaamir: Kept it in case we want to expand in the future, but happy to remove. 

Actually might be okay to remove.

packages/apps/src/types/plugin/sender.ts
```diff
@@ -2,21 +2,18 @@ import { ActivityParams, ConversationReference, SentActivity } from '@microsoft/
...
 import { IStreamer } from '../streamer';
 
-import { IPlugin } from './plugin';
-
 /**
- * a plugin that can send activities
+ * Interface for activity sending (NOT a plugin)
+ * Separates sending concerns from transport concerns
  */
-export interface ISender<TCustomEvents extends {} = {}> extends IPlugin<TCustomEvents> {
```
corinagum, line 10: Could we instead mark ISender as deprecated and send them to `IActivitySender`?

corinagum: (I'm not super convinced on this, more asking than anything)

packages/apps/src/app.ts
```diff
@@ -372,30 +366,37 @@ export class App<TPlugin extends IPlugin = IPlugin> {
...
-      // initialize plugins
-      for (const plugin of this.plugins) {
-        // inject dependencies
-        this.inject(plugin);
-
-        if (plugin.onInit) {
-          plugin.onInit();
-        }
-      }
+      await this.initialize();
```
lilyydu, line 411: missing check if its already been initialized

packages/botbuilder/src/plugin.ts
```diff
@@ -72,8 +71,8 @@ export class BotBuilderPlugin extends HttpPlugin implements ISender {
     this.handler = options?.handler;
   }
 
-  onInit() {
-    super.onInit();
+  async onInit() {
+    await super.onInit();
```
lilyydu, line 75: also need to update IPlugin to be `onInit?(): void | Promise<void>`

packages/apps/src/app.ts
lilyydu, line 450: can reuse `params` - called twice

packages/apps/src/plugins/http/plugin.ts
```diff
@@ -98,7 +83,6 @@ export class HttpPlugin implements ISender {
     this.use = this.express.use.bind(this.express);
 
     this.express.use(cors());
-    this.express.use('/api*', express.json());
```
lilyydu, line 101: removal concern raised by claude:

```
express.json() scope narrowed silently
  Was: this.express.use('/api*', express.json()) (all /api/* routes)
  Now: applied only to POST /api/messages
  Any custom routes added under /api/* via plugin.use(...) will silently stop receiving parsed bodies — subtle breaking change.
```

packages/apps/src/events/activity.ts
```diff
@@ -1,24 +1,45 @@
...
 
+  /**
+   * Allow any additional protocol-specific fields
+   */
+  [key: string]: any;
```
corinagum: Instead of this, could we cast `ICoreActivity` to Activity in `app.process.ts` ~~after validation~~ to protect type safety better? 

`export type ActivityBody = ICoreActivity & Record<string, unknown>;`

Or, instead, we could remove `key: string]: any;`, then in `HttpPlugin.onRequest` cast `body: req.body as ICoreActivity` - doesn't strip extra fields, but they aren't visible to transport plugin.

packages/apps/src/plugins/http/plugin.ts
```diff
@@ -121,7 +105,7 @@ export class HttpPlugin implements ISender {
       });
       messageHandlers.unshift(jwtMiddleware);
     }
-    this.express.post('/api/messages', ...messageHandlers);
+    this.express.post('/api/messages', express.json(), ...messageHandlers);
```
corinagum: I recommend calling this out as a part of the changes, wdyt? Anyone with custom routes would need to have custom parsing now, right? How likely is it for bigger apps to have custom routes that parsed anything going to `/api*`?

I don't think we need to make a big deal of it, but worth mentioning.

packages/apps/src/app.ts
corinagum: `startedAt` seems like dead code (protected). Do we want to use that in conjunction with this or remove it?

packages/apps/src/types/plugin/sender.ts
heyitsaamir: I dno't think it's necessary. Only the ActivitySender is a sender, and you can't have multiple ISenders. Noone would be implementing their own ISender because it's  not documented and v complicated.
If someone really wanted to control the sending, we can expose swapping out IActivitySender maybe? IMO, we should just wait until someone asks, and then expose it. Thoughts?

packages/apps/src/app.ts
```diff
@@ -372,30 +366,37 @@ export class App<TPlugin extends IPlugin = IPlugin> {
...
+   * initialize the app.
+   */
+  async initialize() {
+    // initialize plugins
+    for (const plugin of this.plugins) {
+      // inject dependencies
+      this.inject(plugin);
+
+      if (plugin.onInit) {
+        plugin.onInit();
```
corinagum: Should this be awaited? `IPlugin` allows returning promises, and `BotBuilderPlugin.onInit` is now async.

packages/apps/src/app.ts
heyitsaamir: I wanna remove startedAt, because now the server may not be under our control, so we don't really know if anyone should rely on that

packages/botbuilder/src/plugin.ts
heyitsaamir: It already is that right? The reason this is okay is b/c HttpPlugin (it's parent class) has a defined init.

packages/apps/src/app.ts
heyitsaamir: yes, good point! Interesting, why were we not awaiting it before...

packages/apps/src/app.ts
heyitsaamir: hm, wdym?

packages/apps/src/activity-sender.ts
```diff
@@ -0,0 +1,51 @@
...
+import { ILogger } from '@microsoft/teams.common/logging';
+
+import { HttpStream } from './http-stream';
+import { IActivitySender, IStreamer } from './types';
+
+/**
+ * Handles sending activities to the Bot Framework
+ * Separate from transport concerns (HTTP, WebSocket, etc.)
+ */
+export class ActivitySender implements IActivitySender {
```
corinagum, line 12: unit tests?

packages/apps/src/plugins/http/plugin.ts
heyitsaamir: This is a fair criticism. Let me fix this.

packages/apps/src/plugins/http/plugin.ts
heyitsaamir: Yeah I think low risk, but to avoid the thrash, I'll mark this is as a TODO to remove in the future. And for HttpServer route, I'm going to remove it. Thoughts?

packages/apps/src/app.spec.ts
```diff
@@ -120,9 +125,9 @@ describe('App', () => {
 
       await app.start();
```
corinagum, line 126: Looks like proactive messages need tests. `initialize()` -> `send()` without `start()`

packages/apps/src/plugins/http/plugin.ts
corinagum: Yeah that's a good idea. It helps avoid surprising anyone. for `HttpServer`, add it to list of breaking changes for posterity?

packages/apps/src/activity-sender.ts
heyitsaamir: oh weird, i thought i added them. fixing...

packages/apps/src/app.spec.ts
heyitsaamir: Good idea

packages/apps/src/app.ts
heyitsaamir: fixed

packages/apps/src/activity-sender.ts
heyitsaamir: Added

packages/apps/src/plugins/http/plugin.ts
heyitsaamir: I reverted this, and added a comment

packages/apps/src/app.spec.ts
heyitsaamir: Added a test and an example!

examples/proactive-messaging/README.md
```diff
@@ -0,0 +1,77 @@
+# Proactive Messaging Example
```
corinagum, line 1: Nice :)

PR #427, opened 2026-01-05

ShanmathiMayuramKrithivasan:
**Add support for Targeted Messages**
This PR introduces support for sending targeted messages - messages delivered privately to a specific recipient within a conversation.

Key Updates:

Added an isTargeted boolean parameter to the send, update, reply and delete APIs. When enabled, the message is sent privately to the Recipient.Id specified in the activity payload.

We append isTargetedActivity=true as a query parameter in API URLs when isTargeted is set, allowing backend services to correctly process these requests.

### DIFF
--- a/packages/apps/src/contexts/activity.ts
+++ b/packages/apps/src/contexts/activity.ts
@@ -228,7 +228,7 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
       }
     }
 
-    return this.send(activity, undefined, isTargeted);
+    return this.send(activity, this.ref, isTargeted);
   }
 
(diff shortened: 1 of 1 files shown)


packages/apps/src/contexts/activity.ts
```diff
@@ -228,7 +228,7 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
       }
     }
 
-    return this.send(activity);
+    return this.send(activity, undefined, isTargeted);
```
kavins14: can you update `undefined` to `this.ref`

kavins14 (changes requested):

PR #430, opened 2026-01-08
Add support for Targeted Messages"

heyitsaamir:
NOTE: This is reopening #427 because it needed some design work.
---
This PR introduces support for sending targeted messages - messages delivered privately to a specific recipient within a conversation.

Key Updates:

Added an isTargeted boolean parameter to the send, update, reply and delete APIs. When enabled, the message is sent privately to the Recipient.Id specified in the activity payload.

We append isTargetedActivity=true as a query parameter in API URLs when isTargeted is set, allowing backend services to correctly process these requests.

### DIFF
--- a/packages/api/src/clients/conversation/activity.ts
+++ b/packages/api/src/clients/conversation/activity.ts
@@ -32,35 +32,44 @@ export class ConversationActivityClient {
     this._apiClientSettings = mergeApiClientSettings(apiClientSettings);
   }
 
-  async create(conversationId: string, params: ActivityParams) {
-    const res = await this.http.post<Resource>(
-      `${this.serviceUrl}/v3/conversations/${conversationId}/activities`,
-      params
-    );
+  async create(conversationId: string, params: ActivityParams, isTargeted: boolean = false) {
+    let url = `${this.serviceUrl}/v3/conversations/${conversationId}/activities`;
+    if (isTargeted) {
+      url += '?isTargetedActivity=true';
+    }
+
+    const res = await this.http.post<Resource>(url, params);
     return res.data;
   }
 
-  async update(conversationId: string, id: string, params: ActivityParams) {
-    const res = await this.http.put<Resource>(
-      `${this.serviceUrl}/v3/conversations/${conversationId}/activities/${id}`,
-      params
-    );
+  async update(conversationId: string, id: string, params: ActivityParams, isTargeted: boolean = false) {
+    let url = `${this.serviceUrl}/v3/conversations/${conversationId}/activities/${id}`;
+    if (isTargeted) {
+      url += '?isTargetedActivity=true';
+    }
+
+    const res = await this.http.put<Resource>(url, params);
     return res.data;
   }
 
-  async reply(conversationId: string, id: string, params: ActivityParams) {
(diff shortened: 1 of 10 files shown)


packages/api/src/clients/conversation/activity.ts
```diff
@@ -32,35 +32,44 @@ export class ConversationActivityClient {
...
-      `${this.serviceUrl}/v3/conversations/${conversationId}/activities`,
-      params
-    );
+  async create(conversationId: string, params: ActivityParams, isTargeted: boolean = false) {
+    let url = `${this.serviceUrl}/v3/conversations/${conversationId}/activities`;
+    if (isTargeted) {
+      url += '?isTargetedActivity=true';
+    }
+
+    const res = await this.http.post<Resource>(url, params);
```
heyitsaamir, line 41: isTargetedActivity should be a URL param. `http.post` should have support for that

heyitsaamir: Do the same for all the methods below as well.

packages/apps/src/plugins/http/stream.ts
```diff
@@ -246,25 +246,28 @@ export class HttpStream implements IStreamer {
...
+   * @param isTargeted when true, sends the message privately to the specified recipient
    */
-  protected async send(activity: ActivityParams) {
+  protected async send(activity: ActivityParams, isTargeted: boolean = false) {
     activity = {
       ...activity,
       from: this.ref.bot,
       conversation: this.ref.conversation,
+      // Set recipient from ref.user if provided (for targeted messages)
+      ...(this.ref.user && { recipient: this.ref.user }),
```
heyitsaamir, line 257: @ShanmathiMayuramKrithivasan does this work with streaming? have you tested that?

packages/apps/src/app.ts
```diff
@@ -399,12 +399,20 @@ export class App<TPlugin extends IPlugin = IPlugin> {
    * send an activity proactively
    * @param conversationId the conversation to send to
    * @param activity the activity to send
+   * @param isTargeted when true, sends the message privately to the recipient specified in activity.recipient
    */
-  async send(conversationId: string, activity: ActivityLike) {
+  async send(conversationId: string, activity: ActivityLike, isTargeted: boolean = false) {
```
heyitsaamir, line 404: Can we do targetedRecipientUserId here? 
So users can do `app.send(conversationId, "string message", targetedRecipientUserId);` Right now it's not clear that `isTargeted` will lead to the method throwing under some circumstances.

With `targetedRecipientUserId`, `toActivityParams` does the work of setting the recipient id in the activity.

packages/apps/src/plugins/http/plugin.ts
```diff
@@ -160,16 +160,18 @@ export class HttpPlugin implements ISender {
       ...activity,
       from: ref.bot,
       conversation: ref.conversation,
+      // Set recipient from ref.user if provided (for targeted messages)
+      ...(ref.user && { recipient: ref.user }),
```
heyitsaamir, line 164: Putting this logic in toActivityParams might be better/cleaner

PR #433, opened 2026-01-16
Introduce HttpServer (and begin deprecating HttpPlugin)

heyitsaamir:
In this PR, we introduce a new object called HttpServer and begin to deprecate HttpPlugin.

## Main changes
1. Create `HttpServer` internal class. It accepts an `IHttpServerAdapter` which is the server implementation.
2. Pulled out the express implementation as an `IHttpServerAdapter`
3. Deprecated HttpPlugin, and made it use HttpServer + ExpressAdapter.
4. Changed BotBuilder/A2A/Mcp plugins to depend on HttpServer vs. HttpPlugin.
5. Added examples to show how powerful IHttpServerAdapter can be with different types of servers (hono, fastify).
6. Minor refactor of the jwt middleware such that we can reuse it in HttpServer and app.embed.

## Why:
HTTP is a core part of our sdk. Our App object uses HTTP to set up a server, perform auth validations, and pipe the request to the handlers that are attached, and then return the response. Key part is that Http is a *core* part of App, not a plugin, since core functionality is dependent on it.
Even inside the App object, we were doing special casing for this Http"Plugin" whereas it should never have really been a plugin to begin with. By making it a plugin, we were exposing many non-plugin essential things to the plugin system in general.

So what should it have been? Well, HTTP Plugin had these responsibilities
1. Set up the express server
2. Perform validations if credentials were present
3. Pass the incoming request to App
4. Once App handlers have had a chance to process this incoming request, pass the response back to the server.

So, we introduce a new object called `HttpServer` whose responsibilities are essentially that ^. This object is not a plugin, but an object that's created by App itself.

## Customization

Now this idealogical shift doesn't really warrant us doing this refactor, but we started seeing requests from folks who wanted to hook Teams functionality into existing servers, or replace the underlying server infra with a non-express server. Our recommendation was to rebuild a new HttpPlugin. But rebuilding this plugin is not simple (since we don't really document it anywhere, and didn't expect folks to build their own).
So `HttpServer` exposes an `IHttpServerAdapter` concept. To build the adapter, one simply needs to build out a handler for extracting request data, and a handler for responses. This means that you can build simple custom adapters for your own _existing_ servers. (And if you don't pass one in, we'll build a default express one.) Examples of servers are in the http-adapters folder under examples/.

## Adapter Interface

The `IHttpServerAdapter` interface adapters need to implement:

```typescript
interface IHttpServerAdapter {
  registerRoute(method: HttpMethod, path: string, handler: HttpRouteHandler): void;
  serveStatic?(path: string, directory: string): void;
  start?(port: number): Promise<void>;
  stop?(): Promise<void>;
}
```

Handlers are pure functions — `({ body, headers }) → { status, body }`. No framework-specific request/response objects leak through the abstraction.

### Why `registerRoute`?

Some adapter patterns have the adapter own routing internally and just receive a single callback. But our SDK creates routes dynamically — `app.function('myFunc')` registers `/api/functions/myFunc` at runtime, in addition to the core `/api/messages` endpoint. The adapter needs a `registerRoute` method so that both `HttpServer` and `app.function()` can tell it what paths to listen on.

### Optional methods

`start`/`stop` are optional — serverless adapters (Vercel, Lambda) don't need them. `serveStatic` is optional — only needed for tab hosting.

`HttpMethod` is currently just `'POST'` (the only method the Teams protocol uses). It may expand to a union if needed.

## Backward Compat

We've updated `HttpPlugin` to basically use `HttpServer` with an `ExpressAdapter` internally for backward compat. I don't think this should lead to any breaking changes (even if someone passes in their own `HttpPlugin`). (Tested BotBuilderPlugin, from examples, and it worked without any changes).
However, it should be noted that I marked HttpPlugin as deprecated in this PR, so it should be discouraged going forward, and after the next few versions, it'll be removed.

## Testing

I tested by running the following examples:

1. Echo bot
2. Devtools
3. BotBuilder
4. HttpPlugin
5. Tabs
6. AI (streaming and regular completions)

skip-test-verification (added manifest for tabs)

#### PR Dependency Tree

* **PR #424**
  * **PR #433** 👈
    * **PR #442**

#### PR Dependency Tree

* **PR #424**
  * **PR #433** 👈

### DIFF
--- a/examples/http-adapters/package.json
+++ b/examples/http-adapters/package.json
@@ -18,8 +18,8 @@
     "@microsoft/teams.common": "*",
     "cors": "^2.8.5",
     "express": "^5.0.0",
-    "restify": "^11.1.0",
-    "hono": "^4.6.14"
+    "hono": "^4.6.14",
+    "restify": "^11.1.0"
   },
   "devDependencies": {
     "@microsoft/teams.config": "*",
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,6 +20,9 @@
       },
       "engines": {
         "node": ">=20"
+      },
+      "optionalDependencies": {
+        "@rollup/rollup-linux-x64-gnu": "4.40.1"
       }
     },
     "examples/a2a": {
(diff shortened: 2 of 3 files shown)


examples/http-adapters/README.md
```diff
@@ -0,0 +1,232 @@
+# HTTP Adapter Examples
+
+This example demonstrates how to use **teams.ts** with different HTTP frameworks through custom adapters.
+
+## What is an HTTP Adapter?
+
+An HTTP adapter bridges a specific HTTP framework (Express, Hono, Next.js, etc.) with teams.ts. This allows you to:
+
+- ✅ Use any HTTP framework you prefer
```
lilyydu: Nice! I like how we showcase different adapters

examples/http-adapters/README.md
```diff
@@ -0,0 +1,232 @@
...
+# Run Hono adapter
+npm run dev:hono
+
+# Run Next.js adapter
+npm run dev:nextjs
+```
+
+All adapters start on `http://localhost:3978` with the Teams bot endpoint at `/api/messages`.
+
+## Architecture
```
lilyydu: Will be nice to capture this in our official docs later :)

packages/apps/src/http/http-server.ts
```diff
@@ -0,0 +1,264 @@
...
+  async start(port: number | string) {
+    const portNumber = typeof port === 'string' ? parseInt(port, 10) : port;
+    if (!this._adapter.start) {
+      throw new Error(
+        'Adapter does not implement start(). ' +
+        'Either implement start() in your adapter, or manage server lifecycle manually.'
+      );
+    }
+    await this._adapter.start(portNumber);
+    this.logger.info(`listening on port ${port} 🚀`);
```
heyitsaamir: This is most likely never going to hit.

examples/http-adapters/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "start:express": "NODE_ENV=production tsx express/index.ts",
+    "start:hono": "NODE_ENV=production tsx hono/index.ts",
+    "start:nextjs": "NODE_ENV=production tsx nextjs/server.ts",
+    "build:nextjs": "next build"
+  },
+  "dependencies": {
+    "@microsoft/teams.api": "2.0.5",
+    "@microsoft/teams.apps": "2.0.5",
+    "@microsoft/teams.common": "2.0.5",
+    "express": "^4.21.0",
```
rido-min: can we update to express 5?

examples/http-adapters/express/teams-app.ts
```diff
@@ -0,0 +1,50 @@
...
+  res.json({
+    users: [
+      { id: 1, name: 'Alice' },
+      { id: 2, name: 'Bob' }
+    ]
+  });
+});
+
+expressApp.get('/', (req, res) => {
+  res.send(`
```
rido-min, line 24: this is cute, should we do the same in other languages/samples?

packages/apps/src/app.ts
```diff
@@ -362,14 +398,17 @@ export class App<TPlugin extends IPlugin = IPlugin> {
...
     for (const plugin of this.plugins) {
-      // inject dependencies
       this.inject(plugin);
 
       if (plugin.onInit) {
         plugin.onInit();
       }
     }
 
+    // initialize server
```
heyitsaamir, line 438: We could technically move this to the constructor, as with above ^. So then we won't need `app.initialize()` anymore.

external/a2a/src/server/plugin.ts
```diff
@@ -76,8 +77,9 @@ export class A2APlugin implements IPlugin {
   @Event('custom')
   protected readonly emit!: EmitPluginEvent<A2AEvents>;
 
-  @Dependency()
-  protected readonly _httpPlugin!: HttpPlugin;
+  @HttpServer()
```
heyitsaamir, line 80: For these plugins, we are exposing the HttpServer because of backward compat. The A2A Plugin does a check to see if the adapter is Express or not since it was built with an express server in mind.

external/mcp/src/plugin.ts
```diff
@@ -111,8 +114,8 @@ export class McpPlugin implements IPlugin {
   @Logger()
   readonly logger!: ILogger;
 
-  @Dependency()
-  readonly httpPlugin!: HttpPlugin;
+  @HttpServer()
```
heyitsaamir, line 117: For these plugins, we are exposing the HttpServer because of backward compat. The MCP Plugin does a check to see if the adapter is Express or not since it was built with an express server in mind.

packages/apps/src/http/express-adapter.ts
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * Express adapter for HttpServer
+ *
+ * Handles Express-specific HTTP framework concerns:
+ * - Express app creation and middleware setup
+ * - Route registration via Express routing
+ * - Request/response data extraction and sending
+ * - Server lifecycle management
+ */
+export class ExpressAdapter implements IHttpAdapter {
```
heyitsaamir: This adapter does a lot, but that's mainly because we are maintaining some of the exposed functions from HttpPlugin.

heyitsaamir: This implementation is the "Default" one we ship with. Ideally we should not be shipping with a server at all -- If we want to do `serve(new App())` where `serve` comes from an `express` specific package, that'll lead to a pretty big breaking change for all our docs for our simplest use-cases. For this reason, I think we should leave `app.start` the way it is, and for v3 consider removing `.start`.

packages/apps/src/http/http-server.ts
```diff
@@ -0,0 +1,263 @@
...
+        if (!authHeader) {
+          sendResponse({
+            status: 401,
+            body: { error: 'Unauthorized' }
+          });
+          return;
+        }
+
+        try {
+          token = await this.validateJwt(authHeader, body);
```
heyitsaamir: This isn't using the middleware approach to validate tokens. Just doing it inline. I think that's okay, but if anyone thinks differently, lmk.

packages/apps/src/plugins/http/plugin.ts
```diff
@@ -1,53 +1,28 @@
...
+ * This wrapper will be removed in a few patch versions.
+ *
+ * NOTE: This plugin is named "HttpPlugin" for historical reasons and backwards compatibility.
+ * It is the default HTTP plugin that uses Express as the underlying framework.
  */
 @Plugin({
   name: 'http',
   version: pkg.version,
-  description: 'the default plugin for receiving activities via HTTP',
+  description: 'Will be deprecated: Use HttpServer with server option instead',
```
heyitsaamir: Specifies that this is scheduled for deprecation.

packages/apps/src/plugins/http/plugin.ts
```diff
@@ -56,34 +31,36 @@ export class HttpPlugin {
...
-  }
-  protected _server: http.Server;
+  protected expressAdapter: ExpressAdapter;
+  protected _server: HttpServer;
 
-  get port() {
-    return this._port;
+  constructor(server?: http.Server, options?: { skipAuth?: boolean }) {
+    this.expressAdapter = new ExpressAdapter(server);
+    this._server = new HttpServer(this.expressAdapter, options);
```
heyitsaamir: Uses ExpressAdapter internally, since that was the behavior from before

examples/http-adapters/package.json
heyitsaamir: Yup

heyitsaamir: Updated to 5!

examples/http-adapters/README.md
heyitsaamir: Yup! That's the plan!

examples/http-adapters/package.json
heyitsaamir: Actually the upgrade deserves to be its own PR so it's above in the stack https://github.com/microsoft/teams.ts/pull/442

packages/apps/src/http/adapter.ts
```diff
@@ -0,0 +1,43 @@
...
+export type HttpMethod = 'POST';
+
+export interface IHttpServerAdapter {
+  /**
+   * Register a route handler for a given HTTP method and path
+   * @param method HTTP method
+   * @param path URL path (e.g., '/api/messages')
+   * @param handler Pure function: ({ body, headers }) → { status, body }
+   */
+  registerRoute(method: HttpMethod, path: string, handler: HttpRouteHandler): void;
```
aacebo, line 32: if the move here is to switch from plugins to botbuilder adapters, this makes sense. If not then this adds a new unnecessary paradigm, I originally intended for us to make more specific plugin types, in this case maybe IHttpPlugin, but if the goal is to end with what botbuilder or agents sdk has that this makes sense.

packages/apps/src/http/http-server.ts
```diff
@@ -0,0 +1,212 @@
...
+      this._adapter.serveStatic(path, directory);
+    }
+  }
+
+
+  /**
+   * Handle incoming activity
+   * Validates JWT, signals app, sends response
+   */
+  protected async handleActivity(request: { body: unknown; headers: Record<string, string> }): Promise<{ status: number; body?: unknown }> {
```
aacebo: you may want to expose a TestAdapter to make building tests a bit less verbose. And these types you are writing inline exist as http.Request | Response and all the http packages in node I know of extend these types.

Image

packages/apps/src/test-utils.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import { App, AppOptions } from './app';
+import { HttpMethod, IHttpServerAdapter, HttpRouteHandler } from './http/adapter';
+import { IPlugin } from './types';
+
+/**
+ * Mock HTTP adapter for testing
+ * Provides no-op implementations for all methods
+ */
+export class TestAdapter implements IHttpServerAdapter {
```
aacebo, line 13: disregard my comment about TestAdapter from before, I see it now!

packages/apps/src/http/http-server.ts
```diff
@@ -0,0 +1,212 @@
...
+
+/**
+ * Interface for HTTP server - exposed to plugins
+ */
+export interface IHttpServer {
+  /**
+   * Get the underlying adapter
+   * Useful for plugins that need adapter-specific features
+   */
+  readonly adapter: IHttpServerAdapter;
```
aacebo, line 28: a bit confusing and questionable to have two interfaces with similar roles/methods, not sure I see a reason to have IHttpServer since it seems the adapter has the same methods and more. Or at least since you are exposing adapter publicly you don't really need the duplicated methods.

packages/apps/src/http/adapter.ts
heyitsaamir: It's not really like the botbuilder style tbh. The only similarity is the word "Adapter".

CloudAdapter in BotBuilder land takes a req, and res, and converts it into a TurnContext, and then passes it to the processor. The req and res are still coming from the outside.

In this case, we are converting HttpPlugin into an HttpServer. It's job is to basically sit between an actual server (adapter), and the app. Basically:

```
HttpServerAdapter -- body/headers --> HttpServer -- Activity --> App
HttpServerAdapter <-- status/responsejson -- HttpServer <-- InvokeResponse -- App
```

So the adapter knows nothing about activities or anything. All it knows is how to build the server. So a user can pass any type of HttpServerAdapter that can get body/headers, and send responses to the App. So it removes the need for the user to learn about any "Activity" specific things at all.

The HttpServer which is in the middle basically gets the incoming body/headers, checks authorization, and then passes it to the App.

packages/apps/src/http/http-server.ts
heyitsaamir: Great point!
I make HttpServer internal, removed this interface b/c we don't need it, and then just exposed the adapter via the plugin DI. Simple! 

Good suggestion!

packages/apps/src/app.ts
```diff
@@ -409,6 +454,7 @@ export class App<TPlugin extends IPlugin = IPlugin> {
    */
   async stop() {
     try {
```
lilyydu, line 476: do we need to stop the server here?

packages/apps/src/http/http-server.ts
```diff
@@ -0,0 +1,189 @@
...
+    await this._adapter.start(portNumber);
+  }
+
+  /**
+   * Stop the HTTP server
+   * Called by App.stop() if implemented
+   */
+  async stop() {
+    if (!this._adapter.stop) {
+      this.logger.warn('Adapter does not implement stop(). Skipping server shutdown.');
```
lilyydu, line 117: throwing error in start vs logger warning in stop - should we match these?

external/a2a/src/server/plugin.ts
```diff
@@ -105,6 +107,14 @@ export class A2APlugin implements IPlugin {
     this.middlewares.push(middleware);
   }
   onInit() {
+    const adapter = this.httpServerAdapter;
+    if (!(adapter instanceof ExpressAdapter)) {
+      throw new Error(
```
lilyydu, line 112: key was updated:

` 'Please use: new App({ httpServerAdapter: new ExpressAdapter() })'`

same for other plugins^

packages/apps/src/app.ts
heyitsaamir: Yes! Good catch!

(Tho tbh, we never use this lol)

packages/apps/src/app.ts
```diff
@@ -391,14 +433,17 @@ export class App<TPlugin extends IPlugin = IPlugin> {
...
           await plugin.onStart({ port: this.port });
         }
       }
       this.events.emit('start', this.log);
       this.startedAt = new Date();
+
+      // Start HTTP server
+      await this.server.start(this.port);
     } catch (error: any) {
       this.onError({ error });
```
heyitsaamir, line 468: call on-stop if "start" fails

packages/apps/src/plugins/http/plugin.ts
```diff
@@ -1,53 +1,28 @@
...
+import { HttpServer } from '../../http/http-server';
+import { Plugin } from '../../types';
 
 /**
- * Receives activities via HTTP
- * Handles HTTP server setup, routing, and authentication
+ * @deprecated Will be deprecated. Use HttpServer instead:
+ *   new App({ server: new HttpServer(new ExpressAdapter()) })
+ *
+ * This wrapper will be removed in a few patch versions.
```
corinagum: This feels a little aggressive for a breaking change. Next minor?

packages/apps/src/plugins/http/plugin.ts
```diff
@@ -1,53 +1,28 @@
...
-} from '../../types';
+import { ExpressAdapter } from '../../http';
+import { HttpServer } from '../../http/http-server';
+import { Plugin } from '../../types';
 
 /**
- * Receives activities via HTTP
- * Handles HTTP server setup, routing, and authentication
+ * @deprecated Will be deprecated. Use HttpServer instead:
+ *   new App({ server: new HttpServer(new ExpressAdapter()) })
```
corinagum: needs updating

packages/apps/src/http/index.ts
```diff
@@ -0,0 +1,3 @@
+// Public API exports
+export { HttpMethod, IHttpServerAdapter, HttpRouteHandler } from './adapter';
+export { ExpressAdapter } from './express-adapter';
```
corinagum, line 3: Does `HttpServer` need export? - `HttpPlugin` references it

packages/apps/src/http/http-server.ts
heyitsaamir: Hm, I think I prefer the approach here. Failing to start is fatal. Failing to stop is bad for cleaning up so a warning is good enough

packages/apps/src/plugins/http/plugin.ts
heyitsaamir: Sounds good. I'll keep it general and say future version.

packages/apps/src/http/index.ts
heyitsaamir: Good catch. It does... but I actually don't want to export it out of the package, since it's implementation detail. so I'm leave it out of the barrel file. Thoughts?

lilyydu (dismissed):
LGTM!

corinagum (commented):
LGTM!

external/mcp/src/plugin.ts
```diff
@@ -111,8 +114,8 @@ export class McpPlugin implements IPlugin {
   @Logger()
   readonly logger!: ILogger;
 
-  @Dependency()
-  readonly httpPlugin!: HttpPlugin;
+  @HttpServerDecorator()
```
lilyydu: nit: preference over `HttpServer` to match our other decorators

examples/http-adapters/restify/restify-adapter.ts
```diff
@@ -0,0 +1,65 @@
...
+  /**
+   * Register a route handler for a given HTTP method and path
+   */
+  registerRoute(method: HttpMethod, path: string, handler: HttpRouteHandler): void {
+    const m = method.toLowerCase() as Lowercase<HttpMethod>;
+    this.server[m](path, async (req: restify.Request, res: restify.Response) => {
+      try {
+        const response = await handler({
+          body: req.body,
+          headers: req.headers as Record<string, string>
```
lilyydu: should be `Record<string, string | string[]>`?

packages/apps/src/http/http-stream.ts
lilyydu, line 110: claude raised a polling issue:

Image

Image

heyitsaamir: Hm, yeah I think i've seen this, but I'd rather not increase the scope of this PR (since this PR didn't touch streaming at all). So let's address this separately.

PR #440, opened 2026-01-23
feat(openai): Add token usage logging

hggzm:
## Summary
Adds debug logging for token usage metrics from OpenAI API responses.

## Changes
- Added token usage logging in `OpenAIChatModel.send()`
- Logs prompt_tokens, completion_tokens, and total_tokens
- Uses existing logger infrastructure at debug level

## Benefits
- Enables cost tracking and optimization
- Helps debug context window issues
- Non-intrusive (debug level only)

Fixes #2119, #2002


heyitsaamir:
Ty for the contribution but we're moving in a diff direction with AI related stuff for now.

PR #443, opened 2026-01-26

ShanmathiMayuramKrithivasan:
Adds support for targeted messages that only a specific recipient can see in a conversation.

**Usage**

```
// Target the sender from context
await send(new MessageActivity('Only you see this').withTargetedRecipient(true));

// Target a specific user
await send(new MessageActivity('Private').withTargetedRecipient('user-id'));
```

**Key Changes**
- SDK appends ?isTargetedActivity=true to API calls, so backend services handles the messages accordingly
- Supports create, update, delete, and reply operations
-  Includes tests and example bot


packages/api/src/activities/message/message.ts
```diff
@@ -81,6 +81,16 @@ export interface IMessageActivity extends IActivity<'message'> {
...
+  /**
+   * Indicates if this is a targeted message visible only to a specific recipient.
+   */
+  isTargeted?: boolean;
+
+  /**
+   * The recipient ID for targeted messages. When set, the message will only be visible to this user.
+   */
+  targetedRecipientId?: string;
+
```
heyitsaamir: Can we just keep `isTargeted` here and when setting targetedRecipientId, you can set it directly to `self.recipient`?

packages/api/src/clients/conversation/index.ts
```diff
@@ -74,6 +74,12 @@ export class ConversationClient {
         this._activities.reply(conversationId, id, params),
       delete: (id: string) => this._activities.delete(conversationId, id),
       members: (activityId: string) => this._activities.getMembers(conversationId, activityId),
+      createTargeted: (params: ActivityParams) => this._activities.createTargeted(conversationId, params),
+      updateTargeted: (id: string, params: ActivityParams) =>
+        this._activities.updateTargeted(conversationId, id, params),
+      replyTargeted: (id: string, params: ActivityParams) =>
```
heyitsaamir: has this been tested? (`replyTargeted`?) what does it do?

packages/apps/src/contexts/activity.ts
```diff
@@ -213,7 +213,18 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
   }
 
   async send(activity: ActivityLike, conversationRef?: ConversationReference) {
-    return await this._plugin.send(toActivityParams(activity), conversationRef ?? this.ref);
+    const params = toActivityParams(activity);
+
+    // For targeted message creates (not updates), set the recipient
+    if (params.type === 'message' && 'isTargeted' in params && params.isTargeted && !params.id) {
```
heyitsaamir: if param.id exists, why do we want to skip doing this?

ShanmathiMayuramKrithivasan: Having params.id indicates an update. APX throws an error when we send recipient on update or delete operations, so we only set it for send scenarios as recipient cannot be changes in these scenarios.

packages/api/src/clients/conversation/index.ts
ShanmathiMayuramKrithivasan: Removed it - it wasn't necessary since replies to targeted messages use the same endpoint as creates (just with replyToId set). The routing is already handled correctly. Tested with the sample app.

packages/api/src/activities/message/message.ts
ShanmathiMayuramKrithivasan: Done - withTargetedRecipient(userId) now sets this.recipient directly.

packages/apps/src/contexts/activity.ts
heyitsaamir: Hmm, are you sure? I see this error:

Image

Image

Works fine for delete though

heyitsaamir: Can you double check this? This is the only blocker for this PR. Otherwise all looks good :)

heyitsaamir: Talked offline, The update behavior is being rolled out now in canary. Tested in canary and it worked as expected.

PR #445, opened 2026-01-31
Remove changesets, use dynamic versioning during release

heyitsaamir:
* We made a decision that we should not check-in versions in the git repo.
* In this PR, we replace all the versioning to be 0.0.0 for all packages since versions are now calculated. All workspace dependencies now are using '*' instead of exact versions. That's general convention, but we can also use '0.0.0' if we want. 
* We have a script to set the version during CI that will basically calculate the version for the given commit, and then set all the versions for all the packages to that version dynamically during CI.
* We use [nbgv](https://dotnet.github.io/Nerdbank.GitVersioning/docs/getting-started.html) to do this version calculation.
* In our case, `main` and `release` hold special meaning. `main` branches will allow the creation of "preview" releases, whereas `release` will be stable.
* We need to create a `release` branch and protect it.

skip-test-verification

### DIFF
--- a/examples/targeted-messages/package.json
+++ b/examples/targeted-messages/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@examples/targeted-messages",
-  "version": "0.0.1",
+  "version": "0.0.0",
   "private": true,
   "license": "MIT",
   "main": "dist/index",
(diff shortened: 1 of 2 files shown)


scripts/version.js
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env node
```
rido-min, line 1: does this work on windows hosts?

heyitsaamir: yep tested in pipeline - https://dev.azure.com/DomoreexpGithub/Github_Pipelines/_build/results?buildId=8467&view=results

corinagum:
Hey @rido-min, could you give a final pass? We'll need this for public preview work.

PR #458, opened 2026-02-19
Set nbgv cloud build version variable in publish script

corinagum:
- Set upcoming version number to be in ADO build run's title
- Remove non-working pathFilters from version.json
    - Wildcards do not work in glob
- Add `main` as allowed release branch
   - non-allowed release branches will have the suffix, which is why this is necessary.


rido-min (approved):
I think we can keep main as ReleaseRefSpec for now

PR #482, opened 2026-03-23
Update quoted replies & new quotes features

corinagum:
image

- Remove `withReplyToId()`:  `replyToId` body property has no effect on threading (APX only uses it for encryption)
- Add `QuotedReplyEntity` type with nested `quotedReply` data object (`messageId` required, `senderId`/`senderName`/`preview`/`time`/`isReplyDeleted`/`validatedMessageReference` optional)
- Add `getQuotedMessages()` on `MessageActivity` to read inbound quoted reply entities
- Add `addQuotedReply(messageId, response?)` builder on `MessageActivity` for constructing outbound quotes (build order = reading order)
- Update r`eply()` to stamp `quotedReply` entity `+ <quoted messageId="..."/>` placeholder instead of blockquote HTML; remove `replyToId` assignment
- Add `quoteReply(messageId, activity)` on `ActivityContext` for quoting arbitrary messages
- Mark all quoted reply types and methods as `@experimental`
- Fix bug in `ActivityContext` constructor where `Object.assign(this, rest)` ran before `MessageActivity.from().toInterface()`, causing interface methods (`getQuotedMessages`, `stripMentionsText`, `isRecipientMentioned`, `getAccountMention`) to be missing on `context.activity`. Added regression test.
- Add quoted-replies example exercising all APIs


packages/apps/src/contexts/activity.ts
```diff
@@ -246,19 +257,54 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
...
-      const blockQuote = this.buildBlockQuoteForActivity();
+    if (this.activity.id) {
+      const placeholder = `<quoted messageId="${this.activity.id}"/>`;
+      if (!activity.entities) {
+        activity.entities = [];
+      }
+      activity.entities.push({
+        type: 'quotedReply',
+        quotedReply: { messageId: this.activity.id },
+      });
```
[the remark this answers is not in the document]
corinagum: By design — the entity is metadata that is harmless on non-message types, and the text placeholder is only added when activity.type === 'message' (line 271). This is consistent across all three SDKs.

packages/apps/src/contexts/activity.ts
```diff
@@ -246,19 +257,54 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
...
 
-      if (blockQuote) {
-        activity.text = `${blockQuote}\r\n${activity.text}`;
+      if (activity.type === 'message') {
+        const text = activity.text?.trim() ?? '';
+        activity.text = text ? `${placeholder} ${text}` : placeholder;
```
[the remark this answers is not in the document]
corinagum: this.activity.id is a service-generated ID from the Teams/Bot Framework backend — always a safe alphanumeric string. Escaping would add complexity for a scenario that cannot occur with real Teams activity IDs.

packages/apps/src/contexts/activity.ts
```diff
@@ -246,19 +257,54 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
...
+    }
+    activity.entities.push({
+      type: 'quotedReply',
+      quotedReply: { messageId },
+    });
+
+    if (activity.type === 'message') {
+      const text = activity.text?.trim() ?? '';
+      activity.text = text ? `${placeholder} ${text}` : placeholder;
+    }
```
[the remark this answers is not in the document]
corinagum: Same as the reply() comment — entity is metadata, text placeholder only added for messages (line 300 check). Consistent behavior.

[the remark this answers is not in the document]
corinagum: The messageId parameter comes from a prior send() result or from the service. Same reasoning — these are service-generated IDs.

packages/api/src/activities/message/message.ts
```diff
@@ -394,6 +416,32 @@ export class MessageActivity extends Activity<'message'> implements IMessageActi
...
+   */
+  addQuotedReply(messageId: string, response?: string): this {
+    if (!this.entities) {
+      this.entities = [];
+    }
+    this.entities.push({
+      type: 'quotedReply',
+      quotedReply: { messageId },
+    });
+    this.addText(`<quoted messageId="${messageId}"/>`);
```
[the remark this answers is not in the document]
corinagum: Same reasoning — message IDs are service-generated. Also fixed the text stripping issue: now using a trimmed copy only for the empty-check decision, preserving the original text.

packages/api/src/activities/activity.ts
```diff
@@ -257,11 +257,6 @@ export class Activity<T extends string = string> implements IActivity<T> {
     return this;
   }
```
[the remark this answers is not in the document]
corinagum: Intentional — withReplyToId() is being replaced by the quoted-reply entity approach across all three SDKs (Python, .NET, TypeScript) as part of the design spec. The replyToId property itself still exists on the activity interface for raw access; the builder was the only thing removed.

packages/api/src/activities/message/message.ts
```diff
@@ -394,6 +416,32 @@ export class MessageActivity extends Activity<'message'> implements IMessageActi
...
+
+  /**
+   * Add a quotedReply entity for the given message ID and append a
+   * `<quoted messageId="..."/>` placeholder to text.
+   * If response is provided, it is appended after the placeholder.
+   * @param messageId - The IC3 message ID of the message to quote
+   * @param response - Optional response text to append after the placeholder
+   * @returns this instance for chaining
+   *
+   * @experimental This API is in preview and may change in the future.
```
heyitsaamir: One potential idea that we discussed (i'm just penning it down here). Maybe we could have a feature flag system for these experimental features. We could ship with them off, and interested parties just have to manually turn it on. Then we could flip them on by default. And then we could remove.

Potentially over engineering, but food for thought

packages/api/src/models/entity/quoted-reply-entity.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * preview text of the quoted message
+   */
+  preview?: string | null;
+
+  /**
+   * timestamp of the quoted message
+   */
+  time?: string | null;
```
heyitsaamir, line 32: Format? UTC?

packages/apps/src/contexts/activity.ts
```diff
@@ -163,11 +163,21 @@ export interface IBaseActivityContext<T extends Activity = Activity, TExtraCtx e
   send: (activity: ActivityLike, conversationRef?: ConversationReference) => Promise<SentActivity>;
 
   /**
-   * reply to the inbound activity
+   * reply to the inbound activity, automatically quoting the inbound message
    * @param activity activity to send
    */
   reply: (activity: ActivityLike) => Promise<SentActivity>;
```
heyitsaamir, line 170: i forget, did we consider `reply: (activity: ActivityLike, quotedMessageId?: string) => Promise<SentActivity>`?

packages/apps/src/contexts/activity.ts
```diff
@@ -207,27 +217,28 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
...
-      value.activity = MessageDeleteActivity.from(value.activity).toInterface();
+    if (rest.activity.type === 'messageDelete') {
+      rest.activity = MessageDeleteActivity.from(rest.activity).toInterface();
     }
 
-    if (value.activity.type === 'typing') {
-      value.activity = TypingActivity.from(value.activity).toInterface();
+    if (rest.activity.type === 'typing') {
+      rest.activity = TypingActivity.from(rest.activity).toInterface();
     }
```
heyitsaamir, line 236: NIT: Consider destructuring `activity` out, and then just setting it as you do below.

packages/apps/src/contexts/activity.ts
```diff
@@ -246,19 +257,54 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
...
     activity = toActivityParams(activity);
-    activity.replyToId = this.activity.id;
 
-    if (activity.type === 'message' && activity.text) {
-      const blockQuote = this.buildBlockQuoteForActivity();
+    if (this.activity.id) {
+      const placeholder = `<quoted messageId="${this.activity.id}"/>`;
+      if (!activity.entities) {
+        activity.entities = [];
+      }
```
heyitsaamir: NIT: Could we check if activity is MessageActivity and just use addQuote fluent api If we could avoid spreading this out.

packages/apps/src/contexts/activity.ts
```diff
@@ -246,19 +257,54 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
...
+   * Diagnostic: ExperimentalTeamsQuotedReplies
+   */
+  async quoteReply(messageId: string, activity: ActivityLike) {
+    activity = toActivityParams(activity);
+    const placeholder = `<quoted messageId="${messageId}"/>`;
+
+    if (!activity.entities) {
+      activity.entities = [];
+    }
+    activity.entities.push({
```
heyitsaamir: similar comment as above

packages/api/src/activities/message/message.ts
corinagum: Overall I'm fine with feature flags in the future. I think it should be incorporated in the design/plan going forward as opposed to tacked on towards the end. I looked at different implementations and I think for TS subpath exports would be good, but without prompting Claude seems to agree that making 'breaking' changes on the wire (`reply()`) before availability on the wire is probably best avoided. I think it's best we wait on these changes until it will be available in-client before incorporating it in a release. Plus, once the change IS available on APX/IC3 side, it's not like the blockquote 'old' format will break, there will be plenty of wiggle room (though I still would like a release asap once it is feature-available). 

 I would also like to discuss with the team preview + feature flags. For TS it's a non-issue because 'preview' is just an indicator in inline docs, but for the others it's an explicit opt-in, which makes preview + feature flag kind of awkward/double-opt-in. Having a system in place for when we want preview opt-in v.s. feature flags and whether or not there should/would be both would be a good idea.

packages/apps/src/contexts/activity.ts
corinagum: Do you mean `quoteReply`? We decided to go with that rather than breaking changes on a pre-existing method. There's also the builder method `addQuotedReply`.

packages/api/src/models/entity/quoted-reply-entity.ts
corinagum: IC3 epoch. Updated inline doc for all three langs.

packages/apps/src/contexts/activity.ts
corinagum: Good nit — this isn't specific to the quoted replies changes though (it's the pre-existing constructor enrichment pattern). Can you create a work item for it so we can address it separately?

packages/apps/src/contexts/activity.ts
corinagum: The key difference is prepend vs append — `reply()`/`quoteReply()` prepend the placeholder before existing text (quote above, response below), while `addQuotedReply()` appends via `addText()` for the build-order-equals-reading-order pattern. Delegating to `addQuotedReply()` would put the quote after the developer's text instead of before it.

They also operate on different types — `reply()`/`quoteReply()` work on `ActivityParams` (from `toActivityParams()`), while `addQuotedReply()` is on `MessageActivity`. Converting between them would add complexity for no gain.

packages/apps/src/contexts/activity.ts
corinagum: Same reasoning as above — prepend vs append difference.

packages/apps/src/contexts/activity.ts
heyitsaamir: Nah, not worth it. Next time someone's touching this, they can fix it. It's just an oppotunistic fix.

packages/apps/src/contexts/activity.ts
```diff
@@ -246,19 +257,54 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
 
   async reply(activity: ActivityLike) {
     activity = toActivityParams(activity);
-    activity.replyToId = this.activity.id;
 
-    if (activity.type === 'message' && activity.text) {
-      const blockQuote = this.buildBlockQuoteForActivity();
+    if (this.activity.id) {
```
heyitsaamir, line 280: can reply just call this.quotedReply with this.activity.id?

packages/apps/src/contexts/activity.ts
```diff
@@ -246,19 +257,54 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
...
+   * @param activity - The activity to send
+   *
+   * @experimental This API is in preview and may change in the future.
+   * Diagnostic: ExperimentalTeamsQuotedReplies
+   */
+  async quoteReply(messageId: string, activity: ActivityLike) {
+    activity = toActivityParams(activity);
+    const placeholder = `<quoted messageId="${messageId}"/>`;
+
+    if (!activity.entities) {
```
heyitsaamir: Before adding this, we should check if type of the activity is `message`. 

Also I don't think we should do all this activity specific stuff in this layer. Let's put all that in the activity model layer. Thoughts?

corinagum: C# can't easily delegate.` QuoteReply<T>()` accepts T where `T : IActivity` - it could be any activity type. To call `AddQuotedReply()`, you'd need to cast to MessageActivity. But the bigger issue is AddQuotedReply() appends (via AddText()), while QuoteReply() prepends. AddQuotedReply() returns MessageActivity, not T, so you'd lose the generic type. We could update to only be MessageActivity, but Reply<T> delegates to QuoteReply<T> and Reply<T> and you want to gate for activity type anyway. If QuoteReply only accepts MessageActivity, then Reply<T>, which can't be changed as it's pre-existing and lives on IContext<IActivity> without breaking, needs ugly casting ... it's not really making a cleaner solution imo.

Also, targeted messages set precedent for activity-specific logic in the context layer. I was told multiple times by multiple people to do what targeted messages does. I'll validate against message, I'm good with that suggestion, but otherwise no-op.

heyitsaamir: Yeah even if it just means adding a utility in the model files, let's do that so the code that modifies models is in one area.

heyitsaamir: (basically we just want to avoid doing model manipulation in too many places. If we did it for targeted messages, we should probably fix that too, but afaik, all of that happens in one place)

corinagum: It doesn't happen in one place. Targeted messages manipulates the activity on `context/activity.ts`.

Moved stamping to be on model. Targeted messages remains the same - recommend filing a new work item if you want it fixed.

heyitsaamir (dismissed):
lg2m. Lmk if we change between `quoteReply` and `quote`. Both sound ok to me!

heyitsaamir (dismissed):

corinagum:
**Re: Copilot review comments**

- **messageId escaping**: messageId is an APX-controlled IC3 ID, not user input. No escaping needed.
- **Non-message activity guard on reply()**: Handled in `quote()` — non-message activities bypass entity stamping.
- **withReplyToId() removal**: Intentional.
- **Entity order in prependQuote**: Entity position in the list doesn't affect rendering; APX matches by type.
- **"preview" wording**: Updated to "coming soon."

heyitsaamir (approved):
Looks good. Minor NIT. 

Also, does package-lock need changing?

heyitsaamir (approved):

PR #494, opened 2026-03-29
feat(apps): expose skipCache option on token acquisition methods

### LINKED ISSUE #493: TokenManager: expose skipCache option for token refresh after 401
## Problem

`TokenManager.getBotToken()` and `TokenManager.getGraphToken()` call MSAL's `acquireTokenByClientCredential({ scopes })` without passing `skipCache`. This means callers have no way to force a fresh token after receiving a 401 from the Bot Framework or Graph API.

MSAL caches tokens with a ~5 minute buffer before expiry. A token can expire between the moment it's fetched from cache and when the HTTP response arrives — particularly during:
- Long-running operations (file uploads to OneDrive/SharePoint)
- Streaming sessions that span minutes (Teams streaminfo protocol)
- Proactive messaging that happens long after the initial token was cached

When this race condition occurs, the caller gets a 401 and wants to retry with a fresh token. But calling `getBotToken()` / `getGraphToken()` again returns the same cached token because MSAL still considers it valid.

## Proposed Solution

Add an optional `skipCache` parameter to `getBotToken` and `getGraphToken` that threads through to MSAL's `acquireTokenByClientCredential({ scopes, skipCache })`:

```typescript
// token-manager.ts
async getBotToken(skipCache?: boolean): Promise<IToken | null> { ... }
async getGraphToken(tenantId?: string, skipCache?: boolean): Promise<IToken | null> { ... }

// app.ts
protected async getBotToken(skipCache?: boolean) { ... }
protected async getAppGraphToken(tenantId?: string, skipCache?: boolean) { ... }
```

This is fully backward-compatible — `skipCache` defaults to `false`, so all existing callers are unaffected. Only 401-retry code paths would pass `true`.

## Context

We hit this in the OpenClaw Teams plugin where Bot Framework REST calls (update/delete activity) and Graph API calls (OneDrive uploads, sharing links, chat member lookups) intermittently fail with 401 after the bot has been running for extended periods. We implemented a retry-on-401 pattern but it's ineffective without `skipCache` because the retry gets the same stale token from MSAL's cache.

MSAL's `ClientCredentialRequest` already supports `skipCache?: boolean` — the Teams SDK just doesn't expose it.

SidU:
## Summary

- Add optional `skipCache` parameter to `getBotToken()`, `getGraphToken()`, and `getAppGraphToken()` that threads through to MSAL's `acquireTokenByClientCredential({ skipCache })`
- Covers both credential paths that use `acquireTokenByClientCredential`: client credentials and federated identity credentials
- Fully backward-compatible — `skipCache` defaults to `false`

## Motivation

MSAL caches tokens with a ~5 minute expiry buffer. A token can become stale between when it's fetched from cache and when the API call completes — particularly during long-running operations (file uploads, streaming sessions, proactive messaging). When this happens, the caller gets a 401 and wants to retry with a fresh token, but calling `getBotToken()` / `getGraphToken()` again returns the same cached token because MSAL still considers it valid.

This change lets callers pass `skipCache: true` on retry to force MSAL to acquire a fresh token from Azure AD.

Closes #493

## Changes

| File | Change |
|------|--------|
| `token-manager.ts` | Add `skipCache?` param to `getBotToken`, `getGraphToken`, `getToken`, `getTokenWithClientCredentials`, `getTokenWithFederatedCredentials`; pass through to `acquireTokenByClientCredential({ scopes, skipCache })` |
| `app.ts` | Add `skipCache?` param to `getBotToken`, `getAppGraphToken`; pass through to TokenManager |
| `token-manager.spec.ts` | 4 new tests for `skipCache` behavior + 4 existing assertions updated for `skipCache: false` |

## Test plan

- [x] All 207 existing tests pass
- [x] 4 new tests verify `skipCache: true` is forwarded to MSAL for both bot and graph tokens
- [x] 4 new tests verify `skipCache: false` (default) behavior is unchanged
- [x] `tsc` build passes clean

### DIFF
--- a/packages/apps/src/app.ts
+++ b/packages/apps/src/app.ts
@@ -622,9 +622,9 @@ export class App<TPlugin extends IPlugin = IPlugin> {
   /// Token
   ///
 
-  protected async getBotToken() {
+  protected async getBotToken(skipCache?: boolean) {
     if (!this.tokenManager) return;
-    return await this.tokenManager.getBotToken();
+    return await this.tokenManager.getBotToken(skipCache);
   }
 
   protected async getUserToken(
@@ -640,8 +640,8 @@ export class App<TPlugin extends IPlugin = IPlugin> {
     return res.token;
   }
 
-  protected async getAppGraphToken(tenantId?: string) {
+  protected async getAppGraphToken(tenantId?: string, skipCache?: boolean) {
     if (!this.tokenManager) return;
-    return await this.tokenManager.getGraphToken(tenantId);
+    return await this.tokenManager.getGraphToken(tenantId, skipCache);
   }
 }
--- a/packages/apps/src/token-manager.spec.ts
+++ b/packages/apps/src/token-manager.spec.ts
@@ -89,7 +89,8 @@ describe('TokenManager', () => {
       );
 
       expect(mockAcquireTokenByClientCredential).toHaveBeenCalledWith({
-        scopes: ['https://api.botframework.com/.default']
+        scopes: ['https://api.botframework.com/.default'],
+        skipCache: false
       });
 
       expect(token).not.toBeNull();
@@ -141,7 +142,8 @@ describe('TokenManager', () => {
       const token = await tokenManager.getGraphToken();
 
       expect(mockAcquireTokenByClientCredential).toHaveBeenCalledWith({
-        scopes: ['https://graph.microsoft.com/.default']
+        scopes: ['https://graph.microsoft.com/.default'],
+        skipCache: false
       });
 
       expect(token).not.toBeNull();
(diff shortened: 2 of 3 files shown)


heyitsaamir:
This looks good to me. 

1. Can we use `forceRefresh` instead of `skipCache`? 
2. Instead if making it positional, make it an optional object. (`options?: TokenOptions`). This way this is extendible.

rido-min (changes requested):
holding this PR while we get a good repro, and answers to this [comment](https://github.com/microsoft/teams.ts/issues/493#issuecomment-4150990428)

heyitsaamir:
Closing because of pushback in Rido's comment. We can reopen this once we resolve the issue above.

PR #498, opened 2026-04-02
Add OpenDialogData and SubmitData convenience wrappers

corinagum:
## Summary

- Adds `OpenDialogData` class that extends `SubmitActionData` and abstracts away the `msteams: { type: 'task/fetch' }` protocol detail for opening dialogs. Takes a `dialogId` and optional extra data.
- Adds `SubmitData` class that extends `SubmitActionData` with a convenience constructor accepting an `action` identifier for handler routing. Aligns with [web form `action` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/form#action).
- Above matches Python SDK implementation (microsoft/teams.py#222).
- Updates dialogs and cards examples to use the new classes.
- Updates dialogs example handler to read `dialog_id` (the reserved field set by `OpenDialogData`) instead of `opendialogtype`.
- Regenerated core.ts from latest codegen: non-required hero constructor params now use `?` only when type-safe (arrays, `any`, or types that include `undefined`). Required hero params stay required.
- New generated types: CollabStageInvokeDataValue, OpenUrlDialogAction, PopoverAction, ProgressBar, ProgressRing, Resources, StringResource, TabInfo.

**Note:** Identifier-based handler routing (e.g. `on('dialog.open', 'simple_form')`) is not included in this PR and will be scoped separately.

## Test plan

- [x] 0 TypeScript compilation errors
- [x] 18/18 cards package tests pass
- [x] E2E: dialogs example tested against Teams (dialog open, form submit, multi-step)
- [x] E2E: cards example tested against Teams (submit_basic, save_profile, submit_feedback)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


heyitsaamir (dismissed):
Will you do the routing piece separately? That's when the benefits here get unlocked :)

corinagum:
@heyitsaamir Routing would be a separate effort. This PR covers the cards-side wrappers to address your  feedback on the data construction/DevEx. Still valuable: developers no longer need to know about `msteams` or `TaskFetchSubmitActionData` to open a dialog. Routing can be tracked as its own work item if the team wants to prioritize it, especially since it's in PY already. :)

heyitsaamir (dismissed):
Cool yeah, I think we should prioritize it (i.e. routing) since that's kind of the whole value add for this abstraction). Otherwise this doesn't add that much value.

corinagum:
@heyitsaamir Agreed the routing completes the story. Want to put together a work item for it so we can get it on the board?

dclaux (approved):

rido-min (approved):

PR #499, opened 2026-04-03
Add cancellation token to OnMessage handler in Program.cs template

rido-min:
Enhance the OnMessage handler to support cancellation tokens, allowing for better control over asynchronous operations.


corinagum (approved):
Lgtm, thank you @rido-min

PR #516, opened 2026-04-09
feat: add missing endpoints (Paged Members, Meeting Notifs) & address client gaps

lilyydu:
equivalent to python: https://github.com/microsoft/teams.py/pull/327

- went through the list of gaps and differences betw our clients vs. backend implementation
- added in the missing endpoints from BF, verified that these exist in backend codebase, and manually had Claude verify them w/ my test bot 

**Issues Addressed**
- in TeamsChannelAccount, `userRole` updated to be `str` instead of fixed value, matching BE.
- `aadObjectId` is either returned as `aadObjectId` or `objectId` (scenario dependent, created helper fn to map this)
- for `GET v3/teams/{id}/conversations`, needed to index into `response.json()["conversations"].` I manually tested this.
- missing `tenant_id` in `TeamDetails`
- No DELETE member route for` DELETE v3/conversations/{id}/members/{id}` **-> DEPRECATED**
- `GET v3/conversations/{id}/activities/{id}/members` returns `TeamsChannelAccount`, not `Account`
- Extra parameters `topic` and `bot` in `CreateConversationParams` for `{service_url}/v3/conversations`  **-> DEPRECATED**
- No `GET v3/conversations route`   **-> DEPRECATED**
- removed `is_group` from `CreateConversationParams` - ignored on backend  -> **DEPRECATED**

**New Endpoints**

1) **Paged Conversation Members**
Added get_paged() to ConversationMemberClient. Supports optional page_size and continuation_token query params.
For pageSize to work, the minimum is 50, by default its 200, and max is 500.

 2) **Meeting Notifications**
 Added send_notification() to MeetingClient. Sends a targeted in-meeting notification to specific recipients on specified surfaces (e.g. meetingTabIcon, meetingStage). This is different from targeted messages, this was introduced in 2022. Requires this RSC permission `OnlineMeetingNotification.Send.Chat` and ECS flag enabled for the tenant/bot.


packages/api/src/clients/conversation/member.ts
```diff
@@ -29,19 +29,41 @@ export class ConversationMemberClient {
   }
 
   async get(conversationId: string) {
-    const res = await this.http.get<TeamsChannelAccount[]>(
+    const res = await this.http.get<any[]>(
       `${this.serviceUrl}/v3/conversations/${conversationId}/members`
     );
-    return res.data;
+    return res.data.map(resolveAadObjectId);
```
corinagum, line 35:
```suggestion
    return res.data?.map(resolveAadObjectId);
```
nit - we should have had this optional chaining everywhere from the beginning

corinagum: or not, maybe the one method I saw with chaining was the outlier. Resolving.

packages/api/src/clients/conversation/member.ts
```diff
@@ -29,19 +29,41 @@ export class ConversationMemberClient {
...
+  async getPaged(conversationId: string, pageSize?: number, continuationToken?: string) {
+    const params: Record<string, string | number> = {};
+    if (pageSize !== undefined) params['pageSize'] = pageSize;
+    if (continuationToken !== undefined) params['continuationToken'] = continuationToken;
+
+    const res = await this.http.get<any>(
+      `${this.serviceUrl}/v3/conversations/${conversationId}/pagedMembers`,
+      { params }
+    );
+    return { ...res.data, members: res.data.members.map(resolveAadObjectId) };
```
corinagum, line 61: Should return `PagedMembersResult` type

packages/api/src/clients/meeting.ts
```diff
@@ -52,4 +56,31 @@ export class MeetingClient {
...
+  ) {
+    const body = {
+      type: params.type ?? 'targetedMeetingNotification',
+      value: params.value,
+    };
+    const res = await this.http.post<any>(
+      `${this.serviceUrl}/v1/meetings/${encodeURIComponent(meetingId)}/notification`,
+      body
+    );
+    return res.data || undefined;
```
corinagum, line 85: `MeetingNotificationResponse`?

packages/api/src/clients/conversation/member.ts
lilyydu: I had this originally but noticed none of the other clients have a return type, and all keep it as res.data

packages/api/src/clients/meeting.ts
lilyydu: same as above^ had this originally but noticed none of the other clients have a return type, and all keep it as res.data

corinagum, line 85:
```suggestion
    return res.data ?? undefined;
```

packages/api/src/clients/conversation/activity.ts
```diff
@@ -65,10 +65,10 @@ export class ConversationActivityClient {
   }
 
   async getMembers(conversationId: string, id: string) {
-    const res = await this.http.get<Account[]>(
+    const res = await this.http.get<any[]>(
```
corinagum: Do tests need to be updated? 
- with data `[]` property
- new `resolveAadObjectId` behavior

packages/api/src/clients/meeting.ts
corinagum: discussed in chat: axios resolves `JSON.parse('')` throws, but falls back to raw response text. Empty string should become undefined, therefore `||` is correct

corinagum (dismissed):
lgtm!

corinagum (approved):
lgtm, thanks for the typings updates :)

PR #520, opened 2026-04-11
fix: drain entire queue per flush cycle

heyitsaamir:
Removes the 10-item batch cap in HttpStream flush. Previously, each flush only processed 10 queued items, causing a long streaming tail when the LLM finishes faster than chunks are sent to Teams.

### DIFF
--- a/packages/apps/src/http/http-stream.spec.ts
+++ b/packages/apps/src/http/http-stream.spec.ts
@@ -66,17 +66,17 @@ describe('HttpStream', () => {
     stream.emit('Message 13');
 
     await jest.advanceTimersByTimeAsync(300);
-    // 500ms passed since first emit, second flush should happen
+    // 500ms passed since first emit, second flush drains entire queue
     expect(client.conversations.activities().create).toHaveBeenCalledTimes(2);
     stream.emit('Message 14');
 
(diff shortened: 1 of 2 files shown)


heyitsaamir:

- #513
    - #520 ◀

rajan-chari (dismissed):
Interesting.  Some thoughts as to why this limit might be there and also rebuttals.  TLDR.  seems like a clear win to me.

1) Perhaps there was concern about this text getting too long? 
           this.text += activity. Text;
However we just keep adding to it afik so not clear why limit it to 10.  

2) Keep the user waiting while we append strings
String append operation is orders of magnitude faster than the streaming so the user does not have to wait 

Verdict: PR seems like a clear win.  

Other future improvement:  I remember seeing something about 1/sec update limit.  However, origin for this seems unclear.  Changing that to 1/500 ms would also streaming responsiveness benefits.

rajan-chari:
would love to hear from @aacebo

rajan-chari:
Thanks @heyitsaamir

heyitsaamir:
@copilot  can you go through the history of that file and figure out why that chunk limit was added? Use old PRs if needed too.

lilyydu (dismissed):

lilyydu (approved):

PR #524, opened 2026-04-14
fix: surface Graph API error body in GraphError

heyitsaamir:
## Summary
- Add `GraphError` class to `packages/graph` that extracts the Graph API response body (error code + message) from AxiosErrors
- Wrap `Client.call()` with error handling so Graph errors produce readable messages like `Graph GET /chats/.../messages failed (403): Insufficient privileges to complete the operation.` instead of just `AxiosError: Request failed with status code 403`
- `GraphError` exposes `statusCode`, `code`, and full `body` for programmatic error handling

## Test plan
- [x] Added test for Graph API error with standard `{ error: { code, message } }` body
- [x] Added test for non-standard error body (e.g. plain string)
- [x] Existing tests pass (59/59) — non-Axios errors still propagate unchanged


Jesperholmbergmsft (dismissed):
Ship it!

Jesperholmbergmsft (dismissed):

PR #525, opened 2026-04-14
address model gaps

lilyydu:
equivalent to these 2 python PRs:
https://github.com/microsoft/teams.py/pull/340 (not all applicable bc TS doesnt have defined inbound/outbound interfaces)
https://github.com/microsoft/teams.py/pull/336

This pull request adds deprecation warnings to several activity-related interfaces and types to clarify that certain properties and activities are no longer recognized or returned by the service backend. This improves the clarity of the API and helps developers avoid using obsolete features. Verified with APX repo and manual testing.

**Deprecation of properties and activities:**
* Marked the `relatesTo` property as deprecated in the base `IActivity` interface, the `Activity` class, and in all derived activity interfaces where it appears, indicating it is no longer returned by the backend.
* Added `@deprecated` annotations to the `withRelatesTo` method in the `Activity` class for consistency.

**Deprecation of activity types not recognized by the backend:**
* Marked the `IHandoffActivity` and `ITraceActivity` interfaces as deprecated
* Marked the `IEndOfConversationActivity` interface and `EndOfConversationCode` type as deprecated


packages/api/src/activities/conversation/conversation-update.ts
```diff
@@ -19,6 +19,7 @@ export interface IConversationUpdateActivity extends IActivity<'conversationUpda
 
   /**
    * Indicates whether the prior history of the channel is disclosed.
+   * @deprecated Not recognized by the service backend.
```
heyitsaamir: hm what's the harm in removing this if it never worked and always returned undefined?

heyitsaamir: One thing I think we should think about it to include when it was deprecated, and maybe we even state when it will be removed (maybe in 3 months?)

heyitsaamir: also let's not say "Not recognized by the service backend". Just leave it as @deprecated.

heyitsaamir (dismissed):

packages/api/src/activities/conversation/conversation-update.ts
lilyydu: (resolving after discussing offline)

lilyydu: I think it'll be hard to 100% confirm when these were deprecated / if they ever existed or if we just added it because BF had these fields. Timeline note would be nice - let me raise this w/ team for our ETA on next release!

heyitsaamir (dismissed):

corinagum (changes requested):
I don't think we can fully remove fields that are part of the official activity protocol from Activity or MessageActivity or MessageUpdateActivity themselves for compat reasons. Similar to `channelData` and `replyToId`. I totally agree with cleaning up our usage and deprecating old ABS code, but not acknowledging what's on the protocol doesn't set us up well for the future.

https://github.com/microsoft/teams.py/issues/239 stems from a problem of Teams SDK models diverging from the activity protocol contarct. It applies to these as well, especially for `speak` and `inputHint` which are related to accessibility.

I think for the time being, the deprecation notes should be removed from the items on activity and message activity, message update activity and we continue the discussion on channels support internally.

corinagum (dismissed):
Sorry for accidentally retriggering Copilot. I think we're waiting on one last confirmation? but otherwise lgtm. Thanks for facillitating my concerns.

corinagum (approved):

PR #536, opened 2026-04-21
Prompt Preview Support

ShanmathiMayuramKrithivasan:
Prompt preview allows a bot to reference the original targeted message in its reply, so Teams can render a collapsible preview of the user's prompt.

When a bot replies to a targeted message (reactive scenarios), the SDK now passes the original message's messageId via a targetedMessageInfo entity in the activity's entities array.

For proactive replies, developers can attach the entity themselves with the messageId of the targeted message.


corinagum:
The new public API for prompt preview is missing the `@experimental` / `Diagnostic: ExperimentalTeamsTargeted` JSDoc that `withRecipient(value, isTargeted)` carries at `packages/api/src/activities/activity.ts:292-293`.

Consider adding the same annotation to:
- `addTargetedMessageInfo` in `packages/api/src/activities/activity.ts`
- `TargetedMessageInfoEntity` in `packages/api/src/models/entity/targeted-message-info-entity.ts`

so consumers see the preview warning consistent with the rest of the targeted-messages surface.

packages/apps/src/contexts/activity.ts
```diff
@@ -251,6 +251,23 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
...
+      !params.entities?.some((e) => e.type === 'targetedMessageInfo')
+    ) {
+      if (!params.entities) {
+        params.entities = [];
+      }
+
+      params.entities.push({
+        type: 'targetedMessageInfo',
+        messageId: this.activity.id,
+      });
```
corinagum, line 268: Also scan `params.entities` for any entity with `type === 'quotedReply'` targeting `this.activity.id`, and scan `params.text` for a matching `<quoted messageId="<tm-id>"/>` placeholder. If either is found, strip both and log. Per §6.1 of the prompt preview design, PP owns the preview surface for targeted message flows and `qtdMsgs` must not be set. Using the string `'quotedReply'` avoids taking a dependency on the quoted replies PR, which has not landed on main.

packages/api/src/activities/activity.ts
```diff
@@ -408,6 +408,18 @@ export class Activity<T extends string = string> implements IActivity<T> {
...
+  /**
+   * Add a targeted message info entity.
+   * This links the reply to the original targeted message for prompt preview.
+   * @param messageId the message ID of the targeted message
+   */
+  addTargetedMessageInfo(messageId: string) {
+    return this.addEntity({
+      type: 'targetedMessageInfo',
+      messageId,
+    });
```
corinagum, line 420: A couple of gaps on this helper.

**Duplicate handling.** Calling this twice appends two entities. Per §1 P0 of the design ("One targeted message reply = one blockquote"), and APX's encoder taking only the first (`TargetedMessageInfoEntityEncoder.cs:19-21` via `FirstOrDefault`), the helper should replace any existing `targetedMessageInfo` entity instead of appending. `addAiGenerated` at `activity.ts:352-355` already uses this "skip if already present" pattern.

**Docstring.** Add guidance on where to source `messageId` (from the incoming TM activity's `id`), note that the reactive `ctx.send` / `ctx.reply` paths populate this automatically so this helper is for deferred or cross-context sends, and mention that an invalid or expired `messageId` causes APX to drop the preview silently while still delivering the message.

ShanmathiMayuramKrithivasan: Addressed both - addTargetedMessageInfo now skips if a targetedMessageInfo entity already exists (matching  pattern in this file), and the docstring covers messageId sourcing, reactive auto-population, and APX silent-drop behavior.

examples/targeted-messages/src/index.ts
```diff
@@ -8,154 +8,73 @@ const app = new App({
...
-      '- `test prompt preview` - Reply with targeted message + prompt preview (reactive, auto-managed)\n' +
-      '- `test proactive preview` - Send targeted message with prompt preview (proactive, manual)\n' +
-      '- `test update` - Send then update a targeted message\n' +
-      '- `test delete` - Send then delete a targeted message\n\n' +
-      '💡 *Test in a group chat to verify others don\'t see targeted messages!*\n' +
-      '💡 *Use /slash commands to trigger targeted messages for prompt preview testing.*'
+      '- `send` - Send a targeted message (only you see it)\n' +
+      '- `update` - Send a targeted message, then update it after 3 seconds\n' +
+      '- `delete` - Send a targeted message, then delete it after 3 seconds\n' +
+      '- `public` - Send a public message (visible to all)\n\n' +
```
corinagum: `test` was removed

ShanmathiMayuramKrithivasan: Updated

packages/apps/src/contexts/activity.ts
ShanmathiMayuramKrithivasan: Updated - Now avoids QR collision by stripping any "quotedReply" entities and matching <quoted .../> text placeholder before adding PP. No dependency on the quoted replies PR. Reverted the Reply skip.

corinagum (approved):

PR #538, opened 2026-04-22
 fix: route `app.graph` + `.ctx.graph` to sovereign endpoints via `cloud.graphScope`

corinagum:
- `app.graph` was routing all Microsoft Graph calls to the public-cloud endpoint (`https://graph.microsoft.com`) regardless of the sovereign cloud configured on the app. Sovereign customers (GCCH, DoD, China) would be unable to use
 - `app.graph` - calls would hit the wrong cloud and fail. This PR fixes that by deriving the Graph base URL from the cloud's `graphScope` and plumbing it through `GraphClient`.

  - Public cloud (default): `graphScope = "https://graph.microsoft.com/.default"` → derived base URL = `https://graph.microsoft.com` = previous hardcoded default. **Zero behavior change.**
  - Sovereign clouds: behavior changes from broken (silently public) to correct (per-cloud). This is the fix.

### DIFF
--- a/packages/graph/src/index.ts
+++ b/packages/graph/src/index.ts
@@ -64,10 +64,13 @@ export class Client {
     return this._http;
   }
 
+  /**
+   * Creates a Graph client.
+   * Base URL resolution order: positional `baseUrlRoot` argument, then `options.baseUrlRoot`
+   * (read whether options is a `Client` instance or `ClientOptions`), then the public-cloud default.
+   */
   constructor(options?: Options, baseUrlRoot?: string) {
-    this.baseUrlRoot = baseUrlRoot
-      ?? (options && !('request' in options) ? options.baseUrlRoot : undefined)
-      ?? defaultBaseUrlRoot;
+    this.baseUrlRoot = baseUrlRoot ?? options?.baseUrlRoot ?? defaultBaseUrlRoot;
     if (!options) {
       this._http = new http.Client({
         baseUrl: `${this.baseUrlRoot}/v1.0`,
--- a/packages/graph/src/index.ts
+++ b/packages/graph/src/index.ts
@@ -47,6 +47,12 @@ type Options = (http.Client | http.ClientOptions) & {
   baseUrlRoot?: string;
 };
 
+/** Graph-specific client options. */
+type GraphOptions = {
+  /** Graph service root override for routing Graph calls to sovereign cloud. */
+  baseUrlRoot?: string;
+};
+
 /**
  * /
  * Provides an entry point for invoking Microsoft Graph APIs.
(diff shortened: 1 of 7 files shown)


packages/graph/src/index.ts
```diff
@@ -64,8 +64,13 @@ export class Client {
...
 
-  constructor(options?: Options) {
-    this.baseUrlRoot = options?.baseUrlRoot ?? defaultBaseUrlRoot;
+  /**
+   * Creates a Graph client.
+   * Base URL resolution order: positional `baseUrlRoot` argument, then `options.baseUrlRoot`
+   * (read whether options is a `Client` instance or `ClientOptions`), then the public-cloud default.
+   */
+  constructor(options?: Options, baseUrlRoot?: string) {
+    this.baseUrlRoot = baseUrlRoot ?? options?.baseUrlRoot ?? defaultBaseUrlRoot;
```
heyitsaamir: why not just pass it in the Options? looks like it already exists

corinagum: Haha, I originally followed the same path. It does already exist in `Options`, but it only works cleanly when `options` is the plain-object `ClientOptions` overload. But where it's actually called pass an `http.Client` instance (via `this.client.clone(...)`). Attaching `baseUrlRoot` there means every existing call site where someone passes `http.Client` or `ClientOptions` would need to update to wrap the first arg in `{ client: ...}`, becoming a breaking change. Fwiw I think the liklihood of this breaking a user is extremely low, but is it a public class so it's not a 0 chance, which is why I went with the less clean but non-breaking approach.

heyitsaamir: Hmm, that sort of breaks `Options` in general then. Adding this parameter works for this particular field, but what if more client options need to be overrided?

- With `Options` you can say that the original http.client needs to handle that particular case. Which was the original assumption with this pattern.
- Arguably, if we want overrides on top of the passed in Options, then we should just pastt in `ClientOptions` as the secondary object instead of just the field directly.

I'm honestly okay with either, but i slightly prefer the first. When creating the graph client, wherever that gets created, it should override it there imo.

heyitsaamir: You're right passing `baseUrlRoot` through `clone` doesn't quite work — `http.Client.options` is `protected`, so `GraphClient` can't read it back off the cloned client. And adding `baseUrlRoot` to `http.ClientOptions` is confusing next to `baseUrl` since it's really a Graph concept, not a general HTTP one.

I think the second param is fine, but let's make it a `GraphOptions` object instead of a raw string — that way it scales if more graph-specific config is needed later:

```typescript
type GraphOptions = {
  baseUrlRoot?: string;
};

constructor(options?: Options, graphOptions?: GraphOptions)
```

heyitsaamir (approved):

PR #515, opened 2026-04-08
Security hardening: token validation and service URL improvements

corinagum:
## Summary

Security hardening for token validation, service URL handling, and development tooling.

- **Service URL validation**: Validate inbound `serviceUrl` against allowed hostnames from the configured cloud environment preset. Configurable via `additionalAllowedDomains` for non-standard channels or sovereign clouds without presets.
- **Scope validation**: Use exact set membership instead of substring matching for JWT scope checks.
- **MCP tool results**: Fix `isCallToolResult` returning incorrect results for valid tool responses.
- **DevTools**: Prevent the DevTools plugin from starting in production environments.

## Test plan

- Unit tests for domain allowlist (cloud preset FQDNs, rejected domains, attacker trafficmanager, localhost, custom domains, wildcard, cross-cloud rejection, botframework.com rejected by default)
- Unit tests for exact scope matching (substring rejected, exact match among multiple scopes)
- Unit tests for `isCallToolResult` (valid results, null, primitives, unknown types)
- Unit tests for DevTools production guard
- E2E validated in Teams -- real Bot Framework JWT and serviceUrl pass the allowlist with no regressions
- E2E validated DevTools blocked on `NODE_ENV=production`
- E2E confirmed unauthenticated DevTools WebSocket is unreachable when guard is active

PR #541, opened 2026-04-24
Tighten additionalAllowedDomains API contract

corinagum:
Follow-up to the allowlist feature (shipped via #515). No runtime behavior change for correct callers. Adds hardening informed by review feedback on the sibling Python PR (microsoft/teams.py#404):

1. **Docstring clarity.** `App.additionalAllowedDomains`, `HttpServerOptions`, and `ServiceTokenValidator` constructor all document that entries must be bare hostnames matched exactly (case-insensitive); wildcard patterns like `*.example.com`, URL suffixes, and full URLs are NOT supported. `['*']` is the sole wildcard, used as a disable-validation sentinel.
2. **Defensive copy.** `HttpServer` and `ServiceTokenValidator` no longer store the caller's array by reference. A post-construction mutation of the caller's array would previously have changed validator behavior at runtime; now it is isolated.
3. **Tests** covering the class of bug Python shipped in #404: plumbing from `HttpServerOptions` through `HttpServer.initialize()` into the `ServiceTokenValidator` constructor.

### DIFF
--- a/packages/apps/src/http/http-server.ts
+++ b/packages/apps/src/http/http-server.ts
@@ -21,9 +21,10 @@ export type HttpServerOptions = {
   readonly skipAuth?: boolean;
   /**
    * Additional service URL hostnames accepted beyond the cloud preset.
-   * Entries must be bare hostnames matched exactly (case-insensitive)
+   * Entries must be bare hostnames matched exactly (case-insensitive);
    * wildcard patterns like `'*.example.com'`, URL suffixes, or full URLs are NOT supported.
-   * Pass `['*']` as the sole wildcard to accept any hostname (disables service-URL validation).
+   * If `'*'` is present anywhere in the list, hostname allowlist checks are disabled,
+   * but service URLs must still be valid URLs and use `https:` (except localhost).
    */
   readonly additionalAllowedDomains?: string[];
   readonly logger?: ILogger;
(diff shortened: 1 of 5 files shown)


packages/apps/src/http/http-server.ts
```diff
@@ -75,7 +82,11 @@ export class HttpServer implements IHttpServer {
   constructor(adapter: IHttpServerAdapter, options: HttpServerOptions) {
     this._adapter = adapter;
     this.skipAuth = options.skipAuth ?? false;
-    this.additionalAllowedDomains = options.additionalAllowedDomains;
+    // Defensive copy so post-construction mutation of the caller's array
+    // does not change validator behavior at runtime.
+    this.additionalAllowedDomains = options.additionalAllowedDomains
+      ? [...options.additionalAllowedDomains]
+      : undefined;
```
heyitsaamir: NIT: `this.additionalAllowedDomans = options.additionalAllowedDomains?.slice()` would be more concise

packages/apps/src/middleware/auth/service-token-validator.ts
```diff
@@ -79,7 +85,11 @@ export class ServiceTokenValidator {
     }, logger);
 
     this.credentials = { clientId: appId, tenantId };
-    this.additionalAllowedDomains = additionalAllowedDomains;
+    // Defensive copy so post-construction mutation of the caller's array
```
heyitsaamir, line 88: (same as above)

heyitsaamir (dismissed):

heyitsaamir (approved):

PR #543, opened 2026-04-27
fix: add .mjs outExtension for ESM builds to fix Webpack 5 strict ESM…

### LINKED ISSUE #542: [Bug]: Cannot import @microsoft/teams.client into webpack 5 javascript react project: Module has no exports.
### Bug Description

I try to import @microsoft/teams.client into webpack 5 javascript react-scripts project. Webpack 5 fails to compile in strict mode (the default in webpack 5) because @microsoft/teams.client does not export the App class in a way webpack 5 expects it.

### Steps to Reproduce

package.json:
```
    "dependencies": {
        "@microsoft/teams-js": "^2.22.0",
        "@microsoft/teams.client": "2.0.8",
        "@microsoft/teams.graph-endpoints": "2.0.8",
   }
    "devDependencies": {
        "react-scripts": "5.0.1"
     },
```

Then run:
`react-scripts start`

Webpack complains:
```
Failed to compile.

Attempted import error: 'App' is not exported from '@microsoft/teams.client' (imported as 'teamsClient').
ERROR in ./src/App.jsx 136:22-37
export 'App' (imported as 'teamsClient') was not found in '@microsoft/teams.client' (module has no exports)

ERROR in ./node_modules/@microsoft/teams.client/dist/index.mjs 1:0-28
Module not found: Error: Can't resolve './app' in 'C:\Users\Oliver\Downloads\ms-identity-javascript-react-spa-main\node_modules\@microsoft\teams.client\dist'
Did you mean 'app.mjs'?
BREAKING CHANGE: The request './app' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.

webpack compiled with 2 errors
```

### Expected Behavior

I expected `import {App} from '@microsoft/teams.client';` to import App class from teams.client, but Webpack 5 stumples over the import.

### Actual Behavior

Right now Webpack 5 javascript project cannot import '@microsoft/teams.client' as the npm available package does not provide *.mjs-files which are compatible with webpack 5 strict mode.

### SDK Version

2.0.8

### Node.js Version

22.16.0

### Additional Context

I am not Webpack and JS compiler expert. Please be kind.

Webpack 5 parses javascript modules in strict mode so the file extension (like *.mjs) HAS TO match EXACTLY.
@microsoft/teams.client   ./dist/index.mjs  line 1:   `export { App } from './app';` fails. Imo it should be `export { App } from './app.mjs';`

I tried a workaround in the webpack.config.js so that webpack 5 is more forgiving:
```
module.exports = {
  // ... other config
  module: {
    rules: [
      // ... other rules
      {
        test: /\.m?js/,
        resolve: {
          fullySpecified: false, // Disables the requirement for extensions
        },
      },
    ],
  },
};
```

Now webpack compiles but webpack cannot resolve the import at runtime. Interestingly when I look into browser variables & memory I do find `teamsClient.App` class (containing the expected function names) when I `import * as teamsClient from '@microsoft/teams.client'`. It seems in principle it should work. 

My guess is webpack 5 workaround mixes up the files (app.js, app.mjs) of the imports thus failing during runtime. When the files in dist-folder are generated the js / mjs file extensions should be named explcitly in the import statements so that webpack uses the right one.

Hamzaa6296:
## Bug
When importing `@microsoft/teams.client` in a Webpack 5 / react-scripts 5 project,
the build fails with:

> Module not found: Error: Can't resolve './app' in dist/
> BREAKING CHANGE: The request './app' failed to resolve only because it was
> resolved as fully specified

## Root Cause
tsup is configured with `bundle: false` and `format: ['cjs', 'esm']`. When tsup
builds the ESM output without bundling, it copies each file individually and keeps
bare import specifiers like `'./app'` without file extensions.

Webpack 5 treats `.mjs` files as strict ES Modules and requires all import paths
to be fully specified with explicit extensions. Since `dist/index.mjs` references
`'./app'` instead of `'./app.mjs'`, Webpack 5 cannot resolve it and the build fails.

## Fix
Added `esbuildOptions` to the shared tsup config in `packages/config/tsup.config.js`:

```js
esbuildOptions: (options, { format }) => {
  if (format === 'esm') {
    options.outExtension = { '.js': '.mjs' };
  }
},
```

This tells esbuild to name ESM output files with `.mjs` extensions so that
import paths and file names match exactly, which is what Webpack 5 requires.

## Files Changed
- `packages/config/tsup.config.js`, added `esbuildOptions` for ESM builds

## Testing
- Import `@microsoft/teams.client` in a Webpack 5 react-scripts 5 project
- Run `react-scripts start`
- Verify build completes without extension resolution errors
- Verify `App` class is correctly importable and works at runtime

Closes #542

### DIFF
--- a/packages/config/tsup.config.js
+++ b/packages/config/tsup.config.js
@@ -7,7 +7,12 @@ module.exports = {
   treeshake: true,
   splitting: true,
   clean: true,
-  outDir: 'dist',
-  entry: ['src/**/*.ts', '!src/**/*.spec.ts'],
-  format: ['cjs', 'esm'],
+  outDir: "dist",
+  entry: ["src/**/*.ts", "!src/**/*.spec.ts"],
+  format: ["cjs", "esm"],
+  esbuildOptions: (options, { format }) => {
+    if (format === "esm") {
+      options.outExtension = { ".js": ".mjs" };
+    }
+  },
 };


Hamzaa6296:
Hey! Just wanted to add some context for the maintainers and anyone 
hitting this issue.

**Why this is a silent failure:**
The package installs fine, `npm install` succeeds, and no errors appear
until you actually try to run `react-scripts start`. At that point Webpack 5
throws a cryptic BREAKING CHANGE error that doesn't obviously point to a
missing file extension, so it can take a while to track down.

**Why only Webpack 5 is affected:**
Webpack 4 and most other bundlers are forgiving about missing extensions in
ESM imports. Webpack 5 changed this behavior, when it encounters a `.mjs`
file it enforces strict ES Module resolution, meaning every import path MUST
include an explicit file extension. Since our dist output uses bare specifiers
like `'./app'` instead of `'./app.mjs'`, Webpack 5 refuses to resolve it.

**Why the workaround in the issue doesn't fully work:**
Setting `fullySpecified: false` in webpack config gets past the compile error
but causes a runtime mixup, the dist folder has both `app.js` and `app.mjs`,
and without strict extension matching Webpack grabs the wrong one, causing
the import to fail silently at runtime.

**The fix:**
One `esbuildOptions` block in the shared tsup config, no changes to any
source files, no changes to individual package configs. Since this config is
shared across all packages in the monorepo, all packages get the correct
`.mjs` extensions automatically on next build.

If you've already installed `@microsoft/teams.client@2.0.8`, you'll need to
wait for a patched release and then update your dependency. No changes needed
on the consumer side once the fix is published.

corinagum (changes requested):
Thanks for tackling this @Hamzaa6296. The bug in #542 is real, I reproduced it locally: `.mjs` has `export { App } from './app';`, which Webpack 5 strict ESM rejects. 

Unfortunately I don't think this patch fixes it. I applied it locally on top of main and rebuilt `@microsoft/teams.client`. The output `dist/index.mjs` is byte-identical before and after, still from `'./app'`, no `.mjs` extension on the specifier.

To rule out a stale install / wrong tsup version, I instrumented the callback:
```ts
esbuildOptions: (options, { format }) => {
  console.log('before:', format, options.outExtension);
  if (format === 'esm') options.outExtension = { '.js': '.mjs' };
    console.log('after:', options.outExtension);
  }
```

Output:
before: esm { '.js': '.mjs' }
after:  { '.js': '.mjs' }

So two things are going on:
1. tsup already defaults `outExtension` to `{ '.js': '.mjs' }` for ESM, which is why `.mjs` files exist today. The patch sets it to the same value.
2. With `bundle: false`, esbuild runs in per-file transform mode, and `outExtension` doesn't rewrite import specifiers in that mode. Specifier rewriting is bundle-mode behavior. That's the actual reason the bare from `'./app'` survives.

Actual fix probably needs:
- An esbuild plugin that rewrites relative-import specifiers to add `.mjs` (e.g. esbuild-plugin-file-path-extensions), wired in via `esbuildPlugins`
- A small postbuild script that walks `dist/**/*.mjs` and appends `.mjs` to relative imports
- Switching to bundle: true (probably not desirable; loses per-file output / tree-shaking story)

Could you try one of those and share a `cat dist/index.mjs` from your local build so we can confirm the specifier actually changes? Happy to help review the next iteration.

[Edit]

The repo already has a fix script at `packages/graph-tools/src/fix-esm-import-paths.ts`. It walks `dist/**/*.mjs` and rewrites bare relative specifiers to add `.mjs` (handling both import 'x' and import('x'), plus `index.mjs` directory resolution). `@microsoft/teams.graph-endpoints` and `graph-endpoints-beta` already wire it in via a `fix:esm-imports` script and as expected, they have 0 bare specifiers in their output.

So a good fix is likely just to wire that same step into the shared tsup config. That'd close #542 for all affected packages, not just teams.client.

Hamzaa6296:
> Thanks for tackling this @Hamzaa6296. The bug in #542 is real, I reproduced it locally: `.mjs` has `export { App } from './app';`, which Webpack 5 strict ESM rejects.
> 
> Unfortunately I don't think this patch fixes it. I applied it locally on top of main and rebuilt `@microsoft/teams.client`. The output `dist/index.mjs` is byte-identical before and after, still from `'./app'`, no `.mjs` extension on the specifier.
> 
> To rule out a stale install / wrong tsup version, I instrumented the callback:
> 
> ```ts
> esbuildOptions: (options, { format }) => {
>   console.log('before:', format, options.outExtension);
>   if (format === 'esm') options.outExtension = { '.js': '.mjs' };
>     console.log('after:', options.outExtension);
>   }
> ```
> 
> Output: before: esm { '.js': '.mjs' } after: { '.js': '.mjs' }
> 
> So two things are going on:
> 
> 1. tsup already defaults `outExtension` to `{ '.js': '.mjs' }` for ESM, which is why `.mjs` files exist today. The patch sets it to the same value.
> 2. With `bundle: false`, esbuild runs in per-file transform mode, and `outExtension` doesn't rewrite import specifiers in that mode. Specifier rewriting is bundle-mode behavior. That's the actual reason the bare from `'./app'` survives.
> 
> Actual fix probably needs:
> 
> * An esbuild plugin that rewrites relative-import specifiers to add `.mjs` (e.g. esbuild-plugin-file-path-extensions), wired in via `esbuildPlugins`
> * A small postbuild script that walks `dist/**/*.mjs` and appends `.mjs` to relative imports
> * Switching to bundle: true (probably not desirable; loses per-file output / tree-shaking story)
> 
> Could you try one of those and share a `cat dist/index.mjs` from your local build so we can confirm the specifier actually changes? Happy to help review the next iteration.
> 
> [Edit]
> 
> The repo already has a fix script at `packages/graph-tools/src/fix-esm-import-paths.ts`. It walks `dist/**/*.mjs` and rewrites bare relative specifiers to add `.mjs` (handling both import 'x' and import('x'), plus `index.mjs` directory resolution). `@microsoft/teams.graph-endpoints` and `graph-endpoints-beta` already wire it in via a `fix:esm-imports` script and as expected, they have 0 bare specifiers in their output.
> 
> So a good fix is likely just to wire that same step into the shared tsup config. That'd close #542 for all affected packages, not just teams.client.

Hi @corinagum, I've wired in the fix-esm-import-paths script from graph-tools into the teams.client build as suggested.
However after testing locally, the script doesn't fully fix the imports. Here's the output of cat dist/index.mjs after running the script:
jsexport { App } from './app';
After investigating, the script's regex only matches import * as x from './y' style imports, but the client package uses:

export { X } from './y' , re-export syntax
import { X } from './y' , named imports

These aren't matched by the current regex so they don't get .mjs appended.
Would you like me to also update the regex in fix-esm-import-paths.ts to cover these cases? Or is there a different approach you'd prefer?

corinagum:
@Hamzaa6296 Thanks for the catch on the regex limitation, you're right, and `graph-endpoints` only worked because that happens to emit only `import * as etc from '.etc.mjs'` Hand-written packages emit re-exports and named imports, which the current regex misses. 

If you're amenable, it would be great if you could do the following. It closes #542 across all six affected packages at once, and graph-endpoints/graph-endpoints-beta can drop their explicit `fix:esm-imports` step.  I realize this is a different scope from what you originally committed to, but your finding is appreciated!

1. Move `fix-esm-import-paths.ts` to `packages/config/` (drop explicit esm fix js step since the `onSuccess` hook will handle it). Note to self: need to discuss with team public re-export and `teams-graph-tools fix-esm-imports` subcommand in `graph-tools/src/index.ts` - no in-repo consumers, but as a published package there could be external pickup. Could leave thin shims that re-export from new location or just delete. @heyitsaamir do you have thoughts on this?
2. Extend the regex to cover the missing shapes:
  - `export { X } from './y'`
  - `export * from './y', export * as etc from './y'`
  - `import { X } from './y'`
  - `import X from './y'` (default)
  - keep existing `import * as` and dynamic `import('./y')

Potential regex for static cases:
`/(from\s*['"])(\.\.?\/[^'"]+)(['"])/g`

3. Wire up into shared `tsup.config.js` via `onSuccess`
```ts
const { fixEsmImports } = require('./fix-esm-import-paths');

module.exports = {
  // ...
  onSuccess: async () => { fixEsmImports('./dist'); };
}
```
4. Simplify `graph-endpoints`/`graph-endpoints-beta` build scripts now that the fix runs automatically
5. Add unit tests with fixtures for each import shape, plus smoke check for grepping `dist/**/*.mjs` for bare specifiers returns nothing across all affected packages.

Bigger than the original change, but it's the correct fix and still well-scoped. Happy to review once it's together, thanks for sticking with this and letting me know about the regex gap.

corinagum:
@Hamzaa6296 oops, after bringing this to my team I found out @Jesperholmbergmsft is already on the case for this topic. Thanks so much for your contribution, but to avoid double work I'm going to close this PR. Appreciate your time; hope to see other interactions from you in the future!

PR #548, opened 2026-04-30
fix: call plugin onStart hooks inside initialize() for custom adapter support

### LINKED ISSUE #544: [Bug]: ExpressAdapter and DevToolsPlugin
### Bug Description

The docs says when i use custom http adapter like ExpressAdapter, i should call `app.initialize()`  and should not be calling `app.start()` .but when i do that DevToolsPlugin is not getting started. when i tried to look at the code , i see below code for the start method
```typescript
async start(port) {
        this.port = port || process.env.PORT || 3978;
        try {
            await this.initialize();
            // Start plugins
            for (const plugin of this.plugins) {
                if (plugin.onStart) {
                    await plugin.onStart({ port: this.port });
                }
            }
            this.events.emit('start', this.log);
            // Start HTTP server
            await this.server.start(this.port);
        }
        catch (error) {
            await this.stop();
            this.onError({ error });
        }
    }
```
so if we just call initialize, when will the plugins get started? if i explicitly call onStart for the devToolsPlugin, i see it working
```diff
 await this.app.initialize();
+await this.devToolsPlugin.onStart({port:8080});
```

### Steps to Reproduce

Add DevtoolsPlugin to the ExpressAdapter sample and if you start the app, you can see devtools url(http://localhost:3001/devtools) not accessible.
If you call the `onStart` method after `initialize`, it will start working

### Expected Behavior

devtools should work with custom http adapter

### Actual Behavior

devtools are not accessible

### SDK Version

2.0.8

### Node.js Version

20.20.0

### Additional Context
btw i tried McpPlugin and that works without calling onStart.

_No response_

Hamzaa6296:
Fixes #544, plugins like DevToolsPlugin that rely on onStart() to start their own server were never started when using a custom HTTP adapter (e.g. ExpressAdapter), because onStart hooks only ran inside start(). Moved onStart loop into initialize() and removed the duplicate from start().

### DIFF
--- a/packages/apps/src/app.ts
+++ b/packages/apps/src/app.ts
@@ -1,4 +1,4 @@
-import { AxiosError } from 'axios';
+import { AxiosError } from "axios";
 
 import {
   ActivityLike,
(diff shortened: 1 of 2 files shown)


Hamzaa6296:
## Summary

This PR fixes #544, `DevToolsPlugin` (and any other plugin relying on `onStart`) not working when using a custom HTTP adapter like `ExpressAdapter`.

## Root Cause

When using a custom adapter, the docs correctly say to call `app.initialize()` instead of `app.start()`. However, `onStart` hooks for plugins were only ever called inside `start()`, meaning plugins that spin up their own server (like `DevToolsPlugin` on port 3001) were silently never started.

```ts
// Before, onStart only ran inside start(), never inside initialize()
async start(port) {
  await this.initialize();         // ← only onInit ran here
  for (const plugin of this.plugins) {
    if (plugin.onStart) {
      await plugin.onStart(...)    // ← custom adapter users never reached this
    }
  }
}
```

## Fix

- Moved the `onStart` loop into `initialize()` so it runs regardless of whether the user calls `start()` or manages their own HTTP server
- Removed the duplicate `onStart` loop from `start()` to avoid calling `onStart` twice
- `initialize()` now accepts an optional `port` parameter so plugins get the correct port in both flows

## Behavior

| Scenario | Before | After |
|---|---|---|
| `app.start()` | `onInit` ✅ `onStart` ✅ | `onInit` ✅ `onStart` ✅ |
| `app.initialize()` + custom adapter | `onInit` ✅ `onStart` ❌ | `onInit` ✅ `onStart` ✅ |

## Testing

Added `DevToolsPlugin` to the `ExpressAdapter` sample, called only `app.initialize()`, and confirmed `http://localhost:3001/devtools` is accessible without manually calling `devToolsPlugin.onStart()`.

Also verified `McpPlugin` continues to work as before since it doesn't rely on `onStart`.

corinagum:
Hi @Hamzaa6296, thank you for the changes. Unfortunately right now we are not making DevTools explicit updates, so I have gone with confirming the issue and initiating a small update to our docs. We appreciate your contribution regardless!

PR #553, opened 2026-05-01
fix(http-stream): resolve race condition in close() and streamType override in final message

### LINKED ISSUE #549: [Bug]: HttpStream - race condition in close() and streamType override in final message
### Bug Description

I've found two bugs in `HttpStream` (`packages/apps/src/http/http-stream.ts`):

found in @microsoft/teams.apps version 2.0.5

### Bug 1: `streamType` always remains `informative` in final message
`close()` builds the final activity with `.addStreamFinal().withChannelData(this.channelData)`. Since `addStreamFinal()` sets `streamType: 'final'` but `withChannelData()` merges `this.channelData` afterwards (which contains `streamType: 'informative'` accumulated from typing updates), the final message's `streamType` gets overwritten. Teams never receives the correct final stream state.

**Reproduction:**
```ts
stream.update('Thinking...')
stream.emit('Response text')
await stream.close()
// Final message has streamType: 'informative' instead of 'final'
```

### Bug 2: Race condition in close() - final message sent before flush completes
The while loop in close() checks queue.length and this.id but not this._flushing. If flush() has drained the queue but is still awaiting pushStreamChunk() network calls, close() proceeds immediately and sends the final message. This causes the final message to arrive before intermediate stream chunks.

**Reproduction:**
```ts
stream.emit('chunk 1')
stream.emit('chunk 2')
stream.close()
// close() may send final message before chunk 2 finishes sending
```

Proposed fixes
1. Bug 1: Swap method order in close() so .withChannelData() runs before .addStreamFinal()
2. Bug 2: Add `|| this._flushing` to the while condition in close()

I have a working fix with unit tests ready to submit as a PR once approved.

chonlaphoom:
Fixes #549 

## Summary
Fixes two bugs in HttpStream:

### Bug 1: streamType overwritten to informative in final message
close() calls .addStreamFinal().withChannelData(this.channelData). Since this.channelData contains streamType: informative from typing updates, it overwrites the final value.

### Bug 2: Race condition in close()
The while loop checks queue.length and this.id but not this._flushing. If flush() drained the queue but is still awaiting pushStreamChunk() calls, close() proceeds immediately.

### Fixes
1. Swap .withChannelData() and .addStreamFinal() order
2. Add || this._flushing to the while condition"

### DIFF
--- a/packages/apps/src/http/http-stream.spec.ts
+++ b/packages/apps/src/http/http-stream.spec.ts
@@ -241,4 +241,54 @@ describe('HttpStream', () => {
     // Further emits throw
     expect(() => stream.emit('Should fail')).toThrow(StreamCancelledError);
   });
+
+  test('close sends final message with streamType final after update', async () => {
+    const stream = new HttpStream(client, ref, logger);
+    mockCreate();
+
+    stream.update('Thinking...');
+    stream.emit('first message');
+    stream.emit('last message');
+    stream.close();
+    await jest.runAllTimersAsync();
+
+    expect(client.conversations.activities().create).toHaveBeenLastCalledWith(
+      expect.objectContaining({
+        type: 'message',
+        text: 'first messagelast message',
+        channelData: expect.objectContaining({
+          streamType: 'final',
+        }),
+      })
+    );
+  });
+
+  test('close waits for flush to complete before sending final message', async () => {
+    mockCreate();
+
+    const stream = new HttpStream(client, ref, logger);
+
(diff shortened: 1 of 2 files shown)


chonlaphoom:

corinagum (approved):
Thank you for the contribution! 
As a side note, Python and C# will need some equivalent updates. Filed here:
https://github.com/microsoft/teams.py/issues/418
https://github.com/microsoft/teams.net/issues/473 

I've started working on these :)

PR #561, opened 2026-05-07
Switch to named imports without subpaths

Jesperholmbergmsft:
While setting up some sample projects to repro & test a fix for bug 542, I realized that on older webpack I had to add babel configuration to work around subpath imports inside of the teams.ts libraries. I had to add the alias section here to my webpack.frontend.config.js just for a basic tab test app. Which is doable, but also unnecessary friction:
```
  resolve: {
    extensions: [".tsx", ".ts", ".js"],
    mainFields: ["main", "module", "browser"],
    alias: {
      "@microsoft/teams.common/http$": path.resolve(
        __dirname,
        "node_modules/@microsoft/teams.common/dist/http/index.js"
      ),
      "@microsoft/teams.common/logging$": path.resolve(
        __dirname,
        "node_modules/@microsoft/teams.common/dist/logging/index.js"
      ),
    },
  },
``` 

In addition, I noticed that we often use wildcard imports for readability, but on older bundlers this may impact tree shakability.

Let's clean up both of these and switch to named imports without subpaths. There are no code changes from this, just a bunch of routine imports & type alias changes.

How tested:
 - Built & packed the packages locally; npm installed them into my test project; verified that I could now remove the alias section
 - Build passes, linter passes, unit tests pass.
 - Verified that the diffs detected by Template Sync Verification are false positives, as they didn't have the same imports in the first place. skip-test-verification

### DIFF
--- a/examples/graph/src/index.ts
+++ b/examples/graph/src/index.ts
@@ -1,5 +1,5 @@
 import { App } from '@microsoft/teams.apps';
-import { ConsoleLogger } from '@microsoft/teams.common/logging';
+import { ConsoleLogger } from '@microsoft/teams.common';
 import * as endpoints from '@microsoft/teams.graph-endpoints';
 
 const app = new App({
@@ -10,7 +10,7 @@ const app = new App({
      */
     defaultConnectionName: 'graph'
   },
-  // Instead of setting in ConsoleLogger like below, you can also 
+  // Instead of setting in ConsoleLogger like below, you can also
   // set LOG_LEVEL=debug or LOG_LEVEL=trace env var for verbose SDK logging
   logger: new ConsoleLogger('@tests/auth', { level: 'debug' }),
     // This is an example of overriding the token URL for a specific region (e.g., Europe).
--- a/packages/apps/src/app.oauth.ts
+++ b/packages/apps/src/app.oauth.ts
@@ -6,7 +6,7 @@ import {
   ISignInVerifyStateInvokeActivity,
   TokenExchangeInvokeResponse,
 } from '@microsoft/teams.api';
-import { Client as GraphClient} from '@microsoft/teams.graph';
+import { Client as GraphClient } from '@microsoft/teams.graph';
 
 import { App } from './app';
 import * as contexts from './contexts';
--- a/packages/apps/src/app.oauth.ts
+++ b/packages/apps/src/app.oauth.ts
@@ -6,7 +6,7 @@ import {
   ISignInVerifyStateInvokeActivity,
   TokenExchangeInvokeResponse,
 } from '@microsoft/teams.api';
-import * as graph from '@microsoft/teams.graph';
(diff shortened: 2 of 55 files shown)


packages/apps/src/app.oauth.ts
```diff
@@ -6,7 +6,7 @@ import {
   ISignInVerifyStateInvokeActivity,
   TokenExchangeInvokeResponse,
 } from '@microsoft/teams.api';
-import * as graph from '@microsoft/teams.graph';
+import { Client as GraphClient} from '@microsoft/teams.graph';
```
corinagum:
```suggestion
import { Client as GraphClient } from '@microsoft/teams.graph';
```

packages/client/src/app.ts
```diff
@@ -1,9 +1,13 @@
...
+  type SilentRequest,
+} from '@azure/msal-browser';
 
-import * as teamsJs from '@microsoft/teams-js';
-import * as http from '@microsoft/teams.common/http';
-import { ILogger, ConsoleLogger } from '@microsoft/teams.common/logging';
-import * as graph from '@microsoft/teams.graph';
+import { app } from '@microsoft/teams-js';
+import { ConsoleLogger, Client as HttpClient, ILogger } from '@microsoft/teams.common';
+import  { Client as GraphClient } from '@microsoft/teams.graph';
```
corinagum:
```suggestion
import { Client as GraphClient } from '@microsoft/teams.graph';
```

packages/graph/src/index.spec.ts
```diff
@@ -1,19 +1,19 @@
 import { AxiosError } from 'axios';
 
-import * as http from '@microsoft/teams.common/http';
+import { Client as HttpClient} from '@microsoft/teams.common';
```
corinagum:
```suggestion
import { Client as HttpClient } from '@microsoft/teams.common';
```

examples/graph/src/index.ts
```diff
@@ -1,5 +1,5 @@
 import { App } from '@microsoft/teams.apps';
-import { ConsoleLogger } from '@microsoft/teams.common/logging';
+import { ConsoleLogger } from '@microsoft/teams.common';
 import * as endpoints from '@microsoft/teams.graph-endpoints';
```
corinagum, line 3: was this left intentionally?

heyitsaamir (approved):
Ty for cleaning up!

corinagum (approved):
Left a few nits. Thank you!

examples/graph/src/index.ts
Jesperholmbergmsft: Yeah the endpoints are a bit diffferent...  can't put all the myriad of functions in the root barrel file, but it's set up to tree shake well. Unless you're on webpack 4, in which case you gotta import from a subpath and set up an alias in the webpack config.

heyitsaamir (approved):

PR #571, opened 2026-05-10
fix(api): Activity.timestamp should be string instead of Date (type only)

heyitsaamir:
`Activity.timestamp` was lying about being a `Date`. It's a string. Always was. The doc comment literally says "expressed in ISO-8601 format" two lines above the `: Date` annotation 🙃

Inbound activities go through `express.json()` (no reviver), and `Activity.from()` is just `Object.assign` — nothing ever turns it into a `Date`. So `activity.timestamp.toISOString()` in a handler? Crash.

Fix: type `timestamp` and `localTimestamp` as `string`.

Also removing `withTimestamp()` and `withLocalTimestamp()` because timestamps set on outgoing bot activities are ignored by the chat service, so those builder helpers were a bit of a footgun/no-op.

### DIFF
--- a/packages/api/src/activities/activity.ts
+++ b/packages/api/src/activities/activity.ts
@@ -629,7 +629,7 @@ export class Activity<T extends string = string> implements IActivity<T> {
   /**
    * Sets the activity timestamp as an ISO-8601 string.
    *
-   * @deprecated Timestamps set on outgoing activities are ignored.
+   * @deprecated  Use ActivityInput/MessageActivityInput/TypingActivityInput when constructing outbound activities.
    */
   withTimestamp(value: Date | string) {
     this.timestamp = value instanceof Date ? value.toISOString() : value;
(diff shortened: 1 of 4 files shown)


packages/api/src/activities/activity.ts
```diff
@@ -302,8 +302,8 @@ export class Activity<T extends string = string> implements IActivity<T> {
     return this;
   }
 
-  withTimestamp(value: Date) {
-    this.timestamp = value;
+  withTimestamp(value: Date | string) {
```
rido-min, line 634: we should review if this method should be marked as obsolete, along with changes in #525. Timestamps set on outgoing activities are ignored

corinagum: I tested this out; although APX technically respects the bot's set timestamp in their code, the chat service ignores timestamp set by the bot on outgoing activities. So `withTimestamp()` is a no-op for send. 

This probably applies to `withLocalTimestamp` as well, though I didn't test it out.

heyitsaamir: Removing

corinagum (approved):

PR #579, opened 2026-05-15
fix(api): conditionally set citation encodingFormat based on text presence

chonlaphoom:
While following the citations example (https://microsoft.github.io/teams-sdk/typescript/in-depth-guides/ai/best-practices) (which calls addCitation without text), I encountered a BadSyntax error: citation.appearance.text and citation.appearance.encodingFormat must either both be empty or both have values. This happens because encodingFormat was unconditionally set to 'application/vnd.microsoft.card.adaptive', even when text was omitted

## Summary

Fixes `BadSyntax` error when `addCitation` is called without `text`.

## Problem

Calling `addCitation(position, appearance)` without `appearance.text` causes a 400 error:

```
code: "BadSyntax"
message: "Error while parsing citation entity with id: x"
Detail: "citation.appearance.text and citation.appearance.encodingFormat must either both be empty or both have values"
```

The root cause: `encodingFormat` was hardcoded to `'application/vnd.microsoft.card.adaptive'` unconditionally, but the Teams API requires it to be absent when `text` is absent.

## Fix

Changed `encodingFormat` to be set only when `appearance.text` is truthy:

```ts
encodingFormat: appearance.text ? 'application/vnd.microsoft.card.adaptive' : undefined,
```


corinagum (approved):
LGTM! Thank you for the contribution!

PR #589, opened 2026-05-21
TeamsSDK: Correct imports and return types in misc. packages

Jesperholmbergmsft:
This PR is in preparation of adding additional linter rules. There are no functional changes here, but I'm enabling isolatedDeclarations and isolatedModules to a bunch of smaller packages; rearranging imports to favor import as type whenever feasible, and adding return type annotations where missing. This helps reduce build time a smidge and provides better inputs for any tools consuming these libraries - e.g. for tree shaking or transpiling.

How tested:
 - The code builds; linter and unit tests pass
 - I built & packaged the libraries and used them in a separate playground app to make sure it still builds and runs fine.


packages/config/eslint.config.js
```diff
@@ -50,7 +50,7 @@ export default tseslint.config(
         'error',
         {
           'newlines-between': 'always-and-inside-groups',
-          'groups': ['builtin', 'external', 'internal', 'parent', 'sibling', 'index', 'object', 'type'],
+          'groups': ['builtin', 'external', 'internal', 'parent', 'sibling', 'index', 'object'],
```
Jesperholmbergmsft, line 53: Removing 'type' from the list to allow the type imports to be together with the value imports.

packages/config/eslint.config.js
```diff
@@ -60,8 +60,10 @@ export default tseslint.config(
           ],
           'alphabetize': {
             'order': 'asc',
-            'caseInsensitive': true
+            'caseInsensitive': true,
+            'orderImportKind': 'ignore',
```
Jesperholmbergmsft, line 64: Setting orderImportKind: ignore so the thing isn't too opinionated on whether 'import type' comes before 'import'. The imports are still sorted by module path, that should be enough.

packages/config/eslint.config.js
```diff
@@ -60,8 +60,10 @@ export default tseslint.config(
...
           },
+          'sortTypesGroup': false,
```
Jesperholmbergmsft, line 66: So types aren't imported as a separate subgroup.

packages/openai/src/audio.ts
```diff
@@ -1,9 +1,10 @@
 import OpenAI, { toFile } from 'openai';
 
-import { Fetch } from 'openai/core.mjs';
+import type { Fetch } from 'openai/core.mjs';
 
-import { IAudioModel, TextToAudioParams, AudioToTextParams } from '@microsoft/teams.ai';
-import { ILogger, ConsoleLogger } from '@microsoft/teams.common';
+import type { IAudioModel, TextToAudioParams, AudioToTextParams } from '@microsoft/teams.ai';
+import type { ILogger} from '@microsoft/teams.common';
```
corinagum:
```suggestion
import type { ILogger } from '@microsoft/teams.common';
```
nit

packages/botbuilder/src/plugin.ts
```diff
@@ -7,18 +7,18 @@ import {
...
 
-import { Credentials, IToken } from '@microsoft/teams.api';
+import type { Credentials, IToken } from '@microsoft/teams.api';
 import {
   Dependency,
   ExpressAdapter,
   HttpServer,
-  IHttpServer,
+  type IHttpServer,
   IPlugin,
```
corinagum, line 16: Should these (plus IPlugin &  ILogger below) be separated out?

packages/openai/src/chat.ts
```diff
@@ -1,16 +1,18 @@
...
+import type {
   ChatSendOptions,
   IChatModel,
-  LocalMemory,
   Message,
-  ModelMessage,
+  ModelMessage} from '@microsoft/teams.ai';
+import {
+  LocalMemory
 } from '@microsoft/teams.ai';
```
corinagum:
```suggestion
  ModelMessage
  } from '@microsoft/teams.ai';
import { LocalMemory } from '@microsoft/teams.ai';
```
nit

packages/botbuilder/src/plugin.ts
corinagum: or intentional?

corinagum (approved):
LGTM! just some nits & 1 question. lmk when ready, happy to re-stamp :)

packages/botbuilder/src/plugin.ts
Jesperholmbergmsft: Either is fine really. My preference would be to split this one since it's already on multiple lines, but the ILogger one with just two imports might as well stay on one line.

corinagum (approved):

PR #592, opened 2026-05-22
Default targeted replies for targeted inbound messages

heyitsaamir:
## Summary
- make `send` / `reply` default to targeted when replying to a targeted inbound message
- keep explicit public sends, different recipients, and different conversation refs public
- split the targeting logic into helpers so the rules are less spicy
- keep `targetedMessageInfo` limited to targeted inbound replies, so normal explicit targeted sends still work
- add targeted-messages example commands for `send private` and `send public`

Note: this also cleans up a pre-existing dead branch from when `isTargeted` moved onto `recipient` — `params.recipient?.isTargeted` already means `params.recipient` exists.

image

## Test plan
- `cd packages/apps && npx jest src/contexts/activity.test.ts --runInBand`
- `cd packages/apps && npm run build`
- `cd examples/targeted-messages && npm run build`

### DIFF
--- a/examples/targeted-messages/README.md
+++ b/examples/targeted-messages/README.md
@@ -52,7 +52,7 @@ To make a command behave like `send private` or `send public`:
 4. Only send the private response when that check passes.
 5. You do **not** need to manually set `withRecipient(activity.from, true)` in this example — `ActivityContext` will mark the response targeted automatically when the inbound message was targeted.
 
-That combo makes the bot treat the slash-command message as private and lets you choose whether the response should be private or public. If the response still looks wrong, restart the example so it picks up the latest `ActivityContext` change.
+That combo makes the bot treat the slash-command message as private and lets you choose whether the response should be private or public.
 
 ## Run
 
--- a/examples/targeted-messages/README.md
+++ b/examples/targeted-messages/README.md
@@ -12,6 +12,7 @@ Targeted messages are messages that only a specific recipient can see - other pa
 | `test update` | Sends a targeted message, then updates it after 3 seconds |
 | `test delete` | Sends a targeted message, then deletes it after 3 seconds |
 | `test public` | Sends a public reply (visible to everyone) |
+| `send private` | Only sends a private message if the incoming message is targeted |
 | `test inbound` | Reads `activity.recipient.isTargeted` and reports whether the inbound message was targeted at the bot |
 | `help` | Shows available commands |
 
@@ -24,15 +25,33 @@ The `appPackage/manifest.json` uses `manifestVersion: "devPreview"` because the 
 
 Slash commands arrive at the bot as regular `MessageActivity` events with `activity.recipient.isTargeted === true`, which the `test inbound` handler in this sample demonstrates.
 
+The `send private` command is useful for verifying whether the inbound message was targeted. If it isn't, the bot says `Send it to me privately first!`.
+
 ## Testing in a Group Chat
 
 To properly test targeted messages:
 
 1. Add the bot to a **group chat** with 2+ people
-2. Send `test send`
+2. Pick the command from the `/` slash menu, or type `send private` as a slash command
 3. **Expected result**: 
-   - You (the sender) should see the "🔒 Targeted message"
+   - You (the sender) should see the targeted message
    - Other participants should **NOT** see it
+
+If you type `send private` as a normal message in 1:1 chat, it will not come through as a targeted message, so the private branch won’t fire.
+
+You can also try `send private` to verify the bot only sends a private response when the inbound message is targeted.
+
+## Making a command private
+
+To make a command behave like `send private`:
+
+1. In `appPackage/manifest.json`, keep `supportsTargetedMessages: true` on the bot.
+2. Add the command under a slash-triggered `commandLists` entry for `team` / `groupChat`.
+3. In your handler, check `activity.recipient?.isTargeted === true`.
+4. Only send the private response when that check passes.
+5. You do **not** need to manually set `withRecipient(activity.from, true)` in this example — `ActivityContext` will mark the response targeted automatically when the inbound message was targeted.
+
+That combo makes the bot treat the slash-command message as private and keeps the response private too. If the response still looks public, restart the example so it picks up the latest `ActivityContext` change.
 
 ## Run
 
--- a/examples/targeted-messages/README.md
+++ b/examples/targeted-messages/README.md
@@ -12,6 +12,7 @@ Targeted messages are messages that only a specific recipient can see - other pa
 | `test update` | Sends a targeted message, then updates it after 3 seconds |
 | `test delete` | Sends a targeted message, then deletes it after 3 seconds |
 | `test public` | Sends a public reply (visible to everyone) |
+| `send public` | Only sends a public message if the incoming message is targeted |
 | `send private` | Only sends a private message if the incoming message is targeted |
 | `test inbound` | Reads `activity.recipient.isTargeted` and reports whether the inbound message was targeted at the bot |
 | `help` | Shows available commands |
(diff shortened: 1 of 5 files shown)


packages/apps/src/contexts/activity.ts
```diff
@@ -254,12 +254,18 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
...
-    // For targeted update (params.id exists), we don't update recipient since recipient cannot be changed.
-    if (params.type === 'message' && params.recipient?.isTargeted && !params.id) {
-      if (!params.recipient) {
-        params.recipient = this.activity.from;
-      }
+    // If the inbound message was targeted, default outgoing messages to targeted too.
+    // Callers can still opt out by explicitly supplying a non-targeted recipient.
+    if (
+      params.type === 'message' &&
+      this.activity.recipient?.isTargeted === true &&
```
heyitsaamir: hm this seems wrong.

heyitsaamir: nvm it's right

packages/apps/src/contexts/activity.ts
```diff
@@ -254,20 +254,23 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
   async send(activity: ActivityLike, conversationRef?: ConversationReference) {
     const params = toActivityParams(activity);
 
-    // For targeted send, set the recipient if not already set.
-    // For targeted update (params.id exists), we don't update recipient since recipient cannot be changed.
-    if (params.type === 'message' && params.recipient?.isTargeted && !params.id) {
-      if (!params.recipient) {
```
heyitsaamir, line 260: This actually seems like a bug from before. How can params.recippient.isTargeted be true, and params.recipient not exist?

examples/targeted-messages/src/index.ts
```diff
@@ -50,11 +55,34 @@ app.on('message', async ({ send, activity, api }) => {
...
+    const isTargeted = activity.recipient?.isTargeted === true;
+    console.log('[send public]', { isTargeted });
+
+    if (!isTargeted) {
+      await send('Send it to me privately first!');
+    } else {
+      await send(
+        new MessageActivity('🌍 This is a **public message** — everyone can see this!')
+          .withRecipient(activity.from)
+      );
```
corinagum, line 63: Could we get further documentation underlining that for auto-targeting, caller must pass any recipient to work? (Maybe you had planned this already)

corinagum: (Inline comment indicating recipient = user here would also be good, I think)

packages/apps/src/contexts/activity.ts
```diff
@@ -254,20 +254,23 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
...
+    // Callers can still opt out by explicitly supplying a non-targeted recipient.
+    if (params.type === 'message' && isIncomingTargeted && !params.id && !params.recipient) {
+      params.recipient = {
+        ...this.activity.from,
+        isTargeted: true,
+      };
     }
 
+    const isOutgoingTargeted = params.type === 'message' && params.recipient?.isTargeted === true;
+
```
corinagum: What is the behavior when `params.recipient` is set with auto-quoting/targetedMessageInfo? Quoted replies and targeted messages are mutually exclusive, so does it need to be stripped?

packages/apps/src/contexts/activity.ts
corinagum: Confirmed, pre-existing bug. Originally, `isTargeted` was on `Activity` so `params.isTargeted` and `!params.recipient` checks were indepdent of each other. Now that `isTargeted` is on recipient, `params.recipient` must exist. Could you add a comment to the PR summary as confirmation?

examples/targeted-messages/src/index.ts
```diff
@@ -10,6 +10,11 @@ const app = new App({
 
 app.on('message', async ({ send, activity, api }) => {
   const text = activity.text?.toLowerCase() || '';
+  console.log('[message received]', {
+    text: activity.text,
+    isTargeted: activity.recipient?.isTargeted === true,
+    from: activity.from?.id,
+  });
```
corinagum: Logger? Or are we moving away from it?

examples/targeted-messages/README.md
```diff
@@ -24,16 +26,34 @@ The `appPackage/manifest.json` uses `manifestVersion: "devPreview"` because the
...
+
+To make a command behave like `send private` or `send public`:
+
+1. In `appPackage/manifest.json`, keep `supportsTargetedMessages: true` on the bot.
+2. Add the command under a slash-triggered `commandLists` entry for `team` / `groupChat`.
+3. In your handler, check `activity.recipient?.isTargeted === true`.
+4. Only send the private response when that check passes.
+5. You do **not** need to manually set `withRecipient(activity.from, true)` in this example — `ActivityContext` will mark the response targeted automatically when the inbound message was targeted.
+
+That combo makes the bot treat the slash-command message as private and lets you choose whether the response should be private or public. If the response still looks wrong, restart the example so it picks up the latest `ActivityContext` change.
```
corinagum: This seems like a debugging sentence, maybe remove?

packages/apps/src/contexts/activity.ts
```diff
@@ -254,12 +254,18 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
...
+      params.type === 'message' &&
+      this.activity.recipient?.isTargeted === true &&
+      !params.id &&
+      !params.recipient
+    ) {
+      params.recipient = {
+        ...this.activity.from,
+        isTargeted: true,
+      };
     }
```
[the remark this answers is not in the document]
heyitsaamir: good point, if the conversation id is provided, and is different, we should do a reuglar send.

packages/apps/src/contexts/activity.test.ts
```diff
@@ -279,6 +309,28 @@ describe('ActivityContext', () => {
...
+
+        expect(mockSender.send).toHaveBeenCalledTimes(1);
+        expect(mockSender.send).toHaveBeenCalledWith(
+          expect.objectContaining({
+            text: 'Public message',
+            type: 'message',
+            recipient: expect.objectContaining({ id: 'test-user', name: 'Test User', role: 'user' }),
+          }),
+          mockRef
+        );
```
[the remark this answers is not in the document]
heyitsaamir: good call

examples/targeted-messages/src/index.ts
heyitsaamir: actually the intent here is that the default behavior is to auto-target. But you bring up a really good point that if recipient that is not the user is set, then we shouldn't auto-target

examples/targeted-messages/src/index.ts
heyitsaamir: Yup will do. Just an example so didn't prioritize.

examples/targeted-messages/README.md
heyitsaamir: huh weird ,why did it add this, lol.

packages/apps/src/contexts/activity.ts
```diff
@@ -252,45 +254,80 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
...
+
+    if (this.shouldOutboundBeAutoTargeted(params, conversationRef)) {
+      this.applyTargetedRecipient(params);
     }
 
-    // Auto-populate targetedMessageInfo entity for prompt preview
-    // when replying to a targeted message in the reactive flow.
-    if (
-      params.type === 'message' &&
-      this.activity.recipient?.isTargeted === true
```
heyitsaamir, line 269: Looks like this branch actually tried to do this very behavior we're adding here, but it never worked originally because it tried to do:
```
params.entities.push({
          type: 'targetedMessageInfo',
          messageId: this.activity.id,
        });
```

which is not the way this feature works

heyitsaamir: Yep, good call. `targetedMessageInfo` is not what makes the message private — `recipient.isTargeted` is the delivery mechanism. The PR now only adds `targetedMessageInfo` when replying from a targeted inbound activity, where it can point back to the original targeted prompt for preview/context. Generic targeted sends only use `recipient.isTargeted`, which avoids the 400 we saw when adding preview metadata to non-targeted inbound sends.

packages/apps/src/contexts/activity.ts
heyitsaamir: yup it should be stripped. We just check for isTargetedOutbound, and if it is, we strip the quoted replies.

heyitsaamir: (made it more readable)

corinagum (approved):

PR #609, opened 2026-06-06
Split apps Core around authenticated transports

heyitsaamir:
One liner: add an internal Core runtime for apps and move auth-aware transport setup behind it.

Stacking note: this is stacked on the BotBuilder HTTP adapter PR. That lower PR removes the BotBuilder plugin/server-DI pressure so this Core split can stay focused on App/Core responsibilities.

Why: App was doing runtime setup, token wiring, server setup, and typed client construction all in one place. This splits the boring plumbing out so App can stay focused on Teams behavior like activities, OAuth, ApiClient, and GraphClient.

Core's job, in principle:
- Inbound transport: own the HTTP entrypoint for incoming Teams activities.
- Inbound auth: validate incoming activity requests before App handles them.
- App dispatch: call the App-provided activity handler once the inbound request is valid.
- Outbound transport: provide HttpClient instances for outbound calls.
- Outbound auth: attach the right bot/app Graph auth behavior to those outbound clients.
- Host lifecycle: initialize/start/stop the local server when this runtime owns one.

App's job stays:
- Own Teams semantics: activities, send/reply, OAuth, routing, tabs/functions, plugins, storage, events, and manifest behavior.
- Own typed protocol clients: construct ApiClient and GraphClient using Core-provided transports.
- Own user delegated auth/user Graph for now.

```mermaid
flowchart LR
  Teams[Teams/Bot Framework] -->|incoming HTTP activity| Core

  subgraph Core[Core: runtime + auth]
    Inbound[Inbound HTTP transport]
    InboundAuth[Inbound auth validation]
    Dispatch[Dispatch to App handler]
    Outbound[Authenticated HttpClient factory]
    Token[TokenManager / authorize hook]
    Inbound --> InboundAuth --> Dispatch
    Token --> Outbound
  end

  subgraph App[App: Teams semantics]
    Handler[Activity processing/routing]
    Sender[Send/reply/stream]
    OAuth[OAuth + user token flow]
    TypedClients[ApiClient / GraphClient construction]
  end

  Dispatch --> Handler
  Handler --> Sender
  Handler --> OAuth
  Handler --> TypedClients
  Sender --> TypedClients
  TypedClients -->|bot/app Graph transport| Outbound
  Outbound -->|outbound HTTP calls| Teams
  Outbound -->|outbound Graph calls| Graph[Microsoft Graph]
```

Interesting bits:
- Activity sending now gets a serviceUrl-bound ApiClient from App instead of building from a shared raw HTTP client.
- Core provides authenticated HttpClient instances; App keeps typed client construction.
- Added an authorize hook for custom outbound bot/app Graph token resolution. The older token factory remains, but cannot be combined with authorize.

Reviewer tips:
- Start with packages/apps/src/core/core.ts, then packages/apps/src/app.ts.
- The key thing to check is whether the Core/App boundary feels right.

Testing:
- npm run build --workspace @microsoft/teams.apps
- npm run build --workspace @microsoft/teams.botbuilder


heyitsaamir:
Will revisit later.

PR #610, opened 2026-06-07
Replace BotBuilder plugin with HTTP adapter

heyitsaamir:
Make BotBuilder an HTTP server adapter instead of an App plugin.

Why: BotBuilder participates at the inbound HTTP layer. Modeling it as an adapter keeps App plugin DI cleaner and lets Core/App treat BotBuilder as host integration rather than a privileged plugin that reaches back into the server. This also allows us to clean up our plugin DI that is unnecessarily fairly privileged (eg. why do plugins get access to creds, tokens etc). Doing this allows us to then move toward a more layered architecture (similar to .net), where the "core" layer will be the only "more" priviledged layer (with creds, tokens etc), instead of this information being passed around everywhere.

Migration path:
```ts
// Before
new App({
  plugins: [new BotBuilderPlugin({ handler })],
});

// After
new App({
  httpServerAdapter: new BotBuilderAdapter({ handler }),
});
```

If you already construct a BotBuilder CloudAdapter:
```ts
new App({
  httpServerAdapter: new BotBuilderAdapter({ cloudAdapter, handler }),
});
```

If no cloudAdapter is supplied, BotBuilderAdapter constructs one from explicit credentials, MicrosoftApp* environment variables, or the same CLIENT_ID / CLIENT_SECRET / TENANT_ID env vars used by teams.apps.

Interesting bits:
- Adds BotBuilderAdapter implementing IHttpServerAdapter.
- Removes BotBuilderPlugin and its DI dependencies.
- Cleans up now-unused App plugin DI entries for raw tokens, raw client, metadata, and credentials.

Reviewer tips:
- Start with packages/botbuilder/src/adapter.ts.
- The key thing to check is whether the adapter wrapper preserves the previous BotBuilder flow before forwarding to the Teams handler.

Testing:

image


packages/botbuilder/src/adapter.ts
```diff
@@ -0,0 +1,223 @@
...
+
+      if (res.headersSent) {
+        return next();
+      }
+
+      await this.forwardToTeams(req, res, teamsHandler);
+    } catch (err) {
+      if (!res.headersSent) {
+        res.status(500).send('internal server error');
+      }
```
corinagum, line 199: Old code had a log here. I think we need here as well, otherwise the error is silent.

packages/botbuilder/src/adapter.ts
```diff
@@ -0,0 +1,223 @@
...
+          return;
+        }
+
+        await this.forwardToTeams(req, res, teamsHandler);
+        return;
+      }
+
+      if (res.headersSent) {
+        return next();
+      }
```
corinagum: I believe processActivity returns `InvokeResponse`, no res, so this is dead code. (Also no next handler)

packages/botbuilder/src/adapter.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+    process.env.CLIENT_ID = 'app-id';
+    process.env.CLIENT_SECRET = 'secret';
+    process.env.TENANT_ID = 'tenant-id';
+
+    expect(() => new BotBuilderAdapter()).not.toThrow();
+  });
+
+  it('throws when neither cloudAdapter nor credentials are provided', () => {
+    expect(() => new BotBuilderAdapter()).toThrow('BotBuilderAdapter requires credentials');
+  });
```
corinagum, line 190: Might be good to have test cases around `processBotBuilder` (including failing) as well.

packages/botbuilder/src/adapter.ts
```diff
@@ -0,0 +1,223 @@
...
+
+      await this.forwardToTeams(req, res, teamsHandler);
+    } catch (err) {
+      if (!res.headersSent) {
+        res.status(500).send('internal server error');
```
lilyydu, line 198: should we log here?

packages/botbuilder/src/adapter.ts
```diff
@@ -0,0 +1,223 @@
...
+        if (invokeResponse && invokeResponse.status !== NOT_IMPLEMENTED) {
+          res.status(invokeResponse.status ?? 200).send(invokeResponse.body);
+          return;
+        }
+
+        await this.forwardToTeams(req, res, teamsHandler);
+        return;
+      }
+
+      if (res.headersSent) {
```
lilyydu: I think this code is dead now that we call processActivity

corinagum (approved):

PR #611, opened 2026-06-08
Remove app manifest handling from teams.apps

heyitsaamir:
Removing the manifest concept from `@microsoft/teams.apps`.

Why:
The Teams app manifest is deployment/configuration metadata, not runtime state. The server does not need the manifest to exist when it runs, and in some setups it may not exist at runtime at all.

`teams.apps` currently assumes manifest data can be passed into the runtime, stored on `App`, mutated by `tab()`/`configTab()`, and used for proactive bot identity. That couples runtime behavior to one specific configuration mechanism. If someone configures or packages their app another way, those runtime dependencies are weird and not very useful. Additionally, we are not promoting keeping manifests in `appPackage/` now either (the new CLI doesn't promote it for eg).

It also does not buy us much right now:
- `app.manifest` synthesizes metadata, but nothing in the runtime depends on it.
- `tab()`'s useful runtime behavior is serving static files; the manifest mutation is packaging metadata.
- `configTab()` only mutates manifest metadata.
- `app.name` only reads `manifest.name.full` to set proactive `bot.name`, but bot name is not required.

Interesting bits:
- Removed `manifest`, `_manifest`, and `name` from `App`.
- `tab()` now matches teams.py behavior: it just hosts static content.
- Removed `configTab()` since it only existed to mutate manifest metadata.
- Dropped generated manifest types/schema from the apps package.
- Made `Account.name` optional because proactive bot references don't need to invent a name.
- Devtools falls back to `App` when rendering unnamed senders.

Tips for reviewers:
Start with `packages/apps/src/app.ts` and `packages/apps/src/app.embed.ts`. The giant deletions are generated manifest types/schema.

Testing:
- `npm run build --workspace @microsoft/teams.api`
- `npm test --workspace @microsoft/teams.apps`
- `npm run clean --workspace @microsoft/teams.apps && npm run build --workspace @microsoft/teams.apps`
- `npm run build --workspace @microsoft/teams.devtools`


packages/devtools/src/components/ActivitiesGrid/ActivityGridColumns.tsx
```diff
@@ -117,7 +117,7 @@ const useActivityGridColumns = ({
             );
           case 'from':
             return item.body.from ? (
-              <OverflowCell content={item.body.from.name} className={classes.cell} />
+              <OverflowCell content={item.body.from.name || 'App'} className={classes.cell} />
```
corinagum, line 120: Not sure these changes matter; I don't think we will be doing a new release to devtools, right?

heyitsaamir: name is optional now, so just to make TS happy

corinagum (approved):

PR #622, opened 2026-06-16
test: add integration tests for Teams TypeScript SDK

corinagum:
Adds 12 integration tests that make real API calls against the Teams Bot Framework service.

**Tests:**
- Activities: create, update, reply, delete (4)
- Members: get all, get by ID, get paged (3)
- Conversations: create 1:1, create group (2)
- Teams: get details, get channels (2)
- Reactions: add and delete (1)

**Setup:**
```bash
npm install
cp test/integration/.env.example test/integration/.env.botid-prod
# Fill in credentials
npm run test:integration --workspace=@microsoft/teams.integration-tests
```

Integration tests are isolated from normal `npm test` (no `test` script in the workspace, excluded from turbo).

See [cross-SDK runbook](https://github.com/microsoft/teams-sdk/blob/main/INTEGRATION-TESTS.md) for provisioning and troubleshooting.


test/integration/src/fixture.ts
```diff
@@ -0,0 +1,120 @@
...
+      clientId: config.clientId,
+      clientSecret: config.clientSecret,
+      authority: `https://login.microsoftonline.com/${config.tenantId}`,
+    },
+  });
+
+  const scope = config.scope ?? 'https://api.botframework.com/.default';
+  const tokenResponse = await msalApp.acquireTokenByClientCredential({
+    scopes: [scope],
+  });
```
heyitsaamir, line 80: Hm, why not use App directly here? it has all this already

corinagum: The intent is to test the API layer in isolation without constructing the full App. Theoretically, App adds abstraction that might make issues with HTTP/auth harder to determine. There's no harm in using App, but it wouldn't be consistent with C# and PY using raw API client, and it keeps it explicit on what we're testing.

heyitsaamir: Fair, but we're basically doing exactly what App does :)

heyitsaamir (approved):

PR #627, opened 2026-06-26
Agent 365 support

heyitsaamir:
This is the TypeScript feature branch for Agent 365 and ACF support in 2.1. It brings the stacked identity, lifecycle, proactive messaging, and observability work together as one integration point.

## Included commits

- [`fa31741`](https://github.com/microsoft/teams.ts/commit/fa31741f02cd4392de91a45da1992b78daa1015e) Add agent-365 feature branch placeholder
- [`27b610d`](https://github.com/microsoft/teams.ts/commit/27b610db7260b08d45e0c8f47fefb757ded85fb0) Add AgenticIdentity API surface (#631)
- [`fba29da`](https://github.com/microsoft/teams.ts/commit/fba29da88e8d25012366525628ac03c5deafdb26) Add agentic send helpers (#632)
- [`7851e95`](https://github.com/microsoft/teams.ts/commit/7851e958aa11db9e17400e931de3cda43a428fed) Use flattened conversation reaction helpers
- [`ea22572`](https://github.com/microsoft/teams.ts/commit/ea2257286614bd761c3db82e114068a359ea5d01) Simplify API client scoping (#642)
- [`5ba95ae`](https://github.com/microsoft/teams.ts/commit/5ba95aec417cfbaefbdd72e38fad3314dc7fe4c5) Remove service URL from app send options
- [`3a63a4d`](https://github.com/microsoft/teams.ts/commit/3a63a4d36428428d99fd51665a45cfb9a8bbab2b) Add Agentic User agentLifecycle event support (#643)
- [`0379699`](https://github.com/microsoft/teams.ts/commit/037969903b1087305e0f2905fbaa4fe5155f2532) Add OpenTelemetry API middleware (#655)
- [`da2e41a`](https://github.com/microsoft/teams.ts/commit/da2e41a5ebf7c465a89813c659923e7301a759f9) Add OpenTelemetry app diagnostics (#656)
- [`ed0568b`](https://github.com/microsoft/teams.ts/commit/ed0568ba148fab951e080ac24005822c84508455) Remove unused Agent 365 example dependency
- [`2bd78cc`](https://github.com/microsoft/teams.ts/commit/2bd78cca061ba5e40945c8b76ca2bd0ede4aa8f6) Rename Agent 365 SDK surfaces to Agentic terminology (#668)
[...]

PR #630, opened 2026-06-26
Add agentic auth provider infrastructure

heyitsaamir:
Adds the auth provider abstraction layer that sits between the app and the API client:

- `AuthProvider` type — `{ token(options: { scope?, agenticIdentity? }) }`
- `AuthProviderInterceptor` — HTTP interceptor that resolves tokens via AuthProvider, reads per-request agentic identity from extensions
- `AppAuthProvider` — default implementation backed by `TokenManager`, handles bot/agentic token routing
- `Client` (ApiClient) constructor extended to accept `cloud`, `authProvider`, `agenticIdentity`
- HttpClient interceptor interface updated to receive `log` in request context
- `App.api` wired to use provider-based auth


heyitsaamir:

- #627
    - #628
        - #629
            - #630 ◀
                - #631
                    - #632

packages/api/src/clients/index.ts
```diff
@@ -26,43 +29,71 @@ export class Client {
...
   protected _http: HttpClient;
   protected _apiClientSettings: Partial<ApiClientSettings>;
+  protected _authProvider?: AuthProvider;
+  protected _cloud?: CloudEnvironment;
+  protected _defaultAgenticIdentity?: AgenticIdentity;
 
-  constructor(serviceUrl: string, options?: HttpClient | HttpClientOptions, apiClientSettings?: Partial<ApiClientSettings>, cloud?: CloudEnvironment) {
+  constructor(
+    serviceUrl: string,
+    options?: HttpClient | HttpClientOptions,
```
heyitsaamir: rename this top `httpOptions`

packages/api/src/clients/index.ts
```diff
@@ -26,43 +29,71 @@ export class Client {
...
+    apiClientSettings?: Partial<ApiClientSettings>,
+    cloud?: CloudEnvironment,
+    authProvider?: AuthProvider,
+    agenticIdentity?: AgenticIdentity
+  ) {
     this.serviceUrl = serviceUrl;
+    this._cloud = cloud;
+    this._authProvider = authProvider;
+    this._defaultAgenticIdentity = agenticIdentity;
```
heyitsaamir: I think we can put these in `ApiClientSettings`. Thoughts?

packages/apps/src/app.ts
```diff
@@ -304,6 +297,17 @@ export class App<TPlugin extends IPlugin = IPlugin> {
...
       cloud: this.cloud,
     }, this.log);
+    this.authProvider = new AppAuthProvider(this.tokenManager, this.cloud);
+
+    const serviceUrl = (this.options.serviceUrl ?? process.env.SERVICE_URL ??
+      'https://smba.trafficmanager.net/teams').replace(/\/+$/, '');
+    this.api = new ApiClient(
+      serviceUrl,
+      this.client.clone(),
+      this.options.apiClientSettings,
```
heyitsaamir: the apiclientsettings should Omit authprovider and agentic identity from being passed in.

packages/api/src/clients/index.ts
heyitsaamir: Done — renamed to `httpOptions`.

packages/api/src/clients/index.ts
heyitsaamir: Done — `cloud`, `authProvider`, and `agenticIdentity` are now fields on `ApiClientSettings`. The constructor is now `(serviceUrl, httpOptions?, apiClientSettings?)`.

packages/apps/src/app.ts
heyitsaamir: Done — the app-facing type is now `Pick<ApiClientSettings, 'oauthUrl'>`, so only `oauthUrl` is exposed. `cloud`, `authProvider`, and `agenticIdentity` are managed internally.

packages/api/src/clients/auth-provider-interceptor.ts
```diff
@@ -0,0 +1,41 @@
...
+      if (resolvedToken != null) {
+        log.warn('Auth provider returned an empty token; Authorization header was not added.');
+      }
+      return config;
+    }
+
+    config.headers ??= new AxiosHeaders();
+    config.headers.Authorization = `Bearer ${resolvedToken}`;
+    return config;
+  };
```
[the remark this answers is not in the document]
heyitsaamir: We should remove cloud as a dependency.

packages/api/src/clients/api-client-settings.ts
```diff
@@ -8,6 +11,21 @@ export type ApiClientSettings = {
...
+  /**
+   * Auth provider for resolving tokens per-request.
+   */
+  readonly authProvider?: AuthProvider;
+
+  /**
+   * Default agentic identity for this client instance.
+   */
+  readonly agenticIdentity?: AgenticIdentity;
 };
```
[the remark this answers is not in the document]
heyitsaamir: Hmm, agree. Check what approach we took in Python. I dn't want to add another options thing in this ApiClient.

packages/api/src/clients/auth-provider-interceptor.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+    const client = new HttpClient({ interceptors: [new AuthProviderInterceptor(authProvider)] });
+    const requests = mockAdapter(client);
+
+    await client.get('/test', { headers: { Authorization: 'Bearer explicit-token' } });
+
+    expect(authProvider.token).not.toHaveBeenCalled();
+    expect(requests[0].headers.Authorization).toBe('Bearer explicit-token');
+  });
+
+  it('passes agentic identity extension and uses agentic scope', async () => {
```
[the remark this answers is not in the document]
heyitsaamir: agree

packages/api/src/clients/auth-provider-interceptor.ts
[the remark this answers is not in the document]
heyitsaamir: Done — removed `cloud` from `AuthProviderInterceptor`. It was declared but never used. Constructor is now just `(authProvider, defaultAgenticIdentity?)`.

packages/api/src/clients/api-client-settings.ts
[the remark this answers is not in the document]
heyitsaamir: Fixed — `mergeApiClientSettings` now passes through `cloud`, `authProvider`, and `agenticIdentity` in the returned object. Also resolves `cloud` from `apiClientSettings.cloud` if the separate param is not provided.

packages/api/src/clients/auth-provider-interceptor.spec.ts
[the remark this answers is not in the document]
heyitsaamir: Done — renamed to "forwards agentic identity to auth provider and sets token".

packages/api/src/clients/auth-provider-interceptor.ts
```diff
@@ -0,0 +1,39 @@
...
+export class AuthProviderInterceptor implements Interceptor {
+  constructor(
+    readonly authProvider: AuthProvider,
+    readonly defaultAgenticIdentity?: AgenticIdentity
+  ) { }
+
+  request: Interceptor['request'] = async ({ config, log }) => {
+    if (config.headers?.Authorization || config.headers?.authorization) {
+      return config;
+    }
```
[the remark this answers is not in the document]
heyitsaamir: The interceptor uses `config.headers?.Authorization || config.headers?.authorization` which works correctly — Axios normalizes headers via `AxiosHeaders` but property access still works for reading. Using `.get()/.set()` would be slightly more defensive but isn't needed here since we control header creation. Keeping as-is.

packages/api/src/clients/index.ts
```diff
@@ -26,43 +29,68 @@ export class Client {
...
-  constructor(serviceUrl: string, options?: HttpClient | HttpClientOptions, apiClientSettings?: Partial<ApiClientSettings>, cloud?: CloudEnvironment) {
+  constructor(
+    serviceUrl: string,
+    httpOptions?: HttpClient | HttpClientOptions,
+    apiClientSettings?: Partial<ApiClientSettings>,
+  ) {
     this.serviceUrl = serviceUrl;
+    this._cloud = apiClientSettings?.cloud;
+    this._authProvider = apiClientSettings?.authProvider;
+    this._defaultAgenticIdentity = apiClientSettings?.agenticIdentity;
```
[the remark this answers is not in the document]
heyitsaamir: This likelihood of this breaking someone is very low right now so I'm proposing this is fine.

packages/api/src/clients/auth-provider-interceptor.ts
```diff
@@ -0,0 +1,39 @@
...
+import { AgenticIdentity } from '../models';
+
+import type { AuthProvider } from './auth';
+
+export const AGENTIC_IDENTITY_EXTENSION = 'microsoft.teams.agenticIdentity';
+
+export class AuthProviderInterceptor implements Interceptor {
+  constructor(
+    readonly authProvider: AuthProvider,
+    readonly defaultAgenticIdentity?: AgenticIdentity
```
corinagum, line 15: This interceptor-level `defaultAgenticIdentity`  is TS only concept, meaning default identity applies to all requests on the cflient. Can you confirm this is intentional vs PY's client-level threading?

packages/api/src/clients/auth.ts
```diff
@@ -0,0 +1,10 @@
+import type { AgenticIdentity } from '../models';
+
+export type AuthProviderToken = string | { toString(): string };
+
+export type AuthProvider = {
+  readonly token: (options: {
+    readonly scope?: string;
```
corinagum, line 7: api nit: PY's token requires scope, whereas here it is optional. Should be consistent.

corinagum (commented):
1. Default agentic identity lives in a different layer than PY. Here the interceptor holds `defaultAgenticIdentity` and falls back to it when no per-request extension is set. In PY the interceptor has no default: `agentic_identity` is threaded into the sub-clients, and the interceptor only reads the per-request extension. Is the interceptor-level default the intended TS design, or should it match PY's client-level threading?
2. The summary says the constructor was "extended", but it's actually a positional removal.  `Client` / `ApiClient` 's 4th `cloud`?  param was removed; That's a breaking change for `new ApiClient(url, http, settings, cloud)` (loud compile error for typed consumers, which is good). Should word as a breaking change and a quick check that no sample/doc constructs `ApiClient` with cloud positionally.

packages/api/src/clients/auth-provider-interceptor.ts
heyitsaamir: https://github.com/microsoft/teams.py/pull/464/changes#diff-5cfc007331f6ea3521bbbb8b299f112ef73665fd30d3f324a517e7335606abf7R34 it's also here in py

packages/api/src/clients/auth.ts
heyitsaamir: Token does, but AuthProvider.token doesn't - https://github.com/microsoft/teams.py/pull/464/changes#diff-5cfc007331f6ea3521bbbb8b299f112ef73665fd30d3f324a517e7335606abf7R19

heyitsaamir:
> 1. Default agentic identity lives in a different layer than PY. Here the interceptor holds `defaultAgenticIdentity` and falls back to it when no per-request extension is set. In PY the interceptor has no default: `agentic_identity` is threaded into the sub-clients, and the interceptor only reads the per-request extension. Is the interceptor-level default the intended TS design, or should it match PY's client-level threading?
> 2. The summary says the constructor was "extended", but it's actually a positional removal.  `Client` / `ApiClient` 's 4th `cloud`?  param was removed; That's a breaking change for `new ApiClient(url, http, settings, cloud)` (loud compile error for typed consumers, which is good). Should word as a breaking change and a quick check that no sample/doc constructs `ApiClient` with cloud positionally.

1. In Py, the interceptor also holds a default agentic identity (that's passed in from ApiClient). Lmk if you disagree with this design :).
2. Yeah, agree that cloud was removed. I moved it to apiclientsettings instead of keeping it as the 4th object. Agree it's breaking, but APIClient should be mostly an internal class, so there's a low chance of this being created independently. I'm proposing we move it to settings instead of keeping it an option (this way it can be included in the apisettings plumbing that already exists). Lmk if you're uncomfortable with this and I can move it back.

packages/api/src/clients/api-client-settings.ts
```diff
@@ -8,6 +11,21 @@ export type ApiClientSettings = {
...
+
+  /**
+   * Cloud environment for token scopes and endpoints.
+   */
+  readonly cloud?: CloudEnvironment;
+
+  /**
+   * Auth provider for resolving tokens per-request.
+   */
+  readonly authProvider?: AuthProvider;
```
singhk97, line 23: how would this work with the existing token factory approach?

corinagum:
> > 1. Default agentic identity lives in a different layer than PY. Here the interceptor holds `defaultAgenticIdentity` and falls back to it when no per-request extension is set. In PY the interceptor has no default: `agentic_identity` is threaded into the sub-clients, and the interceptor only reads the per-request extension. Is the interceptor-level default the intended TS design, or should it match PY's client-level threading?
> > 2. The summary says the constructor was "extended", but it's actually a positional removal.  `Client` / `ApiClient` 's 4th `cloud`?  param was removed; That's a breaking change for `new ApiClient(url, http, settings, cloud)` (loud compile error for typed consumers, which is good). Should word as a breaking change and a quick check that no sample/doc constructs `ApiClient` with cloud positionally.
> 
> 1. In Py, the interceptor also holds a default agentic identity (that's passed in from ApiClient). Lmk if you disagree with this design :).
> 2. Yeah, agree that cloud was removed. I moved it to apiclientsettings instead of keeping it as the 4th object. Agree it's breaking, but APIClient should be mostly an internal class, so there's a low chance of this being created independently. I'm proposing we move it to settings instead of keeping it an option (this way it can be included in the apisettings plumbing that already exists). Lmk if you're uncomfortable with this and I can move it back.

Overall I'm fine with it, just want to be super clear in the PRs/documentation. If there are samples/docs that construct `ApiClient` it's more likely that it will be breaking for customers.

corinagum (approved):

packages/api/src/clients/api-client-settings.ts
heyitsaamir: AuthProvider provided by app uses tokenmanager -> token manager uses the token factory if provided

PR #632, opened 2026-06-26
Add agentic send helpers

heyitsaamir:
Wires agentic identity into the app-layer send/reply helpers (mirrors Python PR #480):

- `app.send(conversationId, activity, options?)` accepts `RequestOptions`
- `app.reply()` supports options via `isRequestOptions` type guard
- `app.getAgenticIdentity(appId, userId, opts?)` factory method
- `ActivitySender` refactored to accept pre-built `Client`; merges `agenticIdentity` and `serviceUrl` into all API calls
- Per-turn `ActivitySender` in `$process` with identity from `activity.recipient`
- `IActivitySender` interface extended with `readonly agenticIdentity?`
- `ActivityContext.send()` simplified — delegates to sender directly
- `examples/agent365/` — reactive echo + proactive send sample


heyitsaamir:

- #627
    - #628
        - #629
            - #630
                - #631
                    - #632 ◀

packages/apps/src/contexts/activity.ts
```diff
@@ -279,7 +279,8 @@ export class ActivityContext<T extends Activity = Activity, TExtraCtx extends {}
       }
     }
 
-    return await this.activitySender.send(params, conversationRef ?? this.ref);
+    const ref = conversationRef ?? this.ref;
+    return await this.activitySender.send(params, ref);
```
heyitsaamir: unnecessary

packages/apps/src/activity-sender.ts
```diff
@@ -32,23 +37,28 @@ export class ActivitySender implements IActivitySender {
       throw new Error('Targeted messages are not supported in 1:1 (personal) chats.');
     }
 
+    // Build options with ref's serviceUrl override and sender's agentic identity
+    const reqOptions: RequestOptions = {
+      ...options,
+      serviceUrl: ref.serviceUrl,
+      ...(this.agenticIdentity && { agenticIdentity: this.agenticIdentity }),
```
heyitsaamir: I don't believe this is necessary anymore, since api is set with a default agentic identity for this context

packages/apps/src/app.process.ts
```diff
@@ -43,7 +44,14 @@ export async function $process<TPlugin extends IPlugin>(
...
 
   const client = this.client.clone();
-  const apiClient = new ApiClient(serviceUrl, this.client.clone({ token: () => this.getBotToken() }), this.options.apiClientSettings);
+  const apiClient = new ApiClient(
+    serviceUrl,
+    this.client.clone(),
+    this.options.apiClientSettings,
+    this.cloud,
+    this.authProvider,
+    getAgenticIdentity(activity.recipient)
```
heyitsaamir: we're getting agentic identities twice

packages/apps/src/contexts/activity.ts
heyitsaamir: Removed — no more `return await` (unnecessary without try/catch).

packages/apps/src/activity-sender.ts
heyitsaamir: Agreed — removed. The API client now carries the default agenticIdentity via `ApiClientSettings`, so the sender doesn't need to merge it per-request.

packages/apps/src/app.process.ts
heyitsaamir: Fixed — `getAgenticIdentity()` is now called once and passed into the per-turn `ApiClient` via `apiClientSettings.agenticIdentity`. The `ActivitySender` no longer carries or merges it.

packages/apps/src/app.ts
```diff
@@ -552,13 +559,29 @@ export class App<TPlugin extends IPlugin = IPlugin> {
...
+   * @param agenticAppId the agentic app's ID
+   * @param agenticUserId the agentic user's ID
+   * @param opts optional overrides (tenantId defaults to this app's configured tenant)
+   */
+  getAgenticIdentity(agenticAppId: string, agenticUserId: string, opts?: { tenantId?: string }): AgenticIdentity {
+    return {
+      agenticAppId,
+      agenticUserId,
+      tenantId: opts?.tenantId ?? this.options.tenantId ?? '',
+    };
```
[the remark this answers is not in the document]
heyitsaamir: Fixed — `getAgenticIdentity()` now uses `opts?.tenantId ?? this.options.tenantId` (no empty string fallback), so nullish coalescing works correctly downstream in `TokenManager.getAgenticToken`.

packages/apps/src/activity-sender.ts
```diff
@@ -32,23 +35,27 @@ export class ActivitySender implements IActivitySender {
...
+      : await this.api.conversations.activities(ref.conversation.id).create(activity, reqOptions);
     return { ...activity, ...res };
   }
 
   createStream(ref: ConversationReference): IStreamer {
-    // Create API client for this conversation's service URL
-    const api = new Client(ref.serviceUrl, this.client);
-    return new HttpStream(api, ref, this.logger);
+    return new HttpStream(this.api, ref, this.logger);
   }
```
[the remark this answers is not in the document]
heyitsaamir: Good catch. The `createStream()` path uses the shared Client which was already constructed with the correct serviceUrl for this conversation turn (set in `app.process.ts`). The serviceUrl in `ref` matches the Client's serviceUrl since both come from the incoming activity. No override needed in practice, but will note it as a potential issue if we support cross-tenant streaming in the future.

packages/apps/src/activity-sender.ts
```diff
@@ -32,23 +35,27 @@ export class ActivitySender implements IActivitySender {
...
+        ? await this.api.conversations.activities(ref.conversation.id).updateTargeted(activity.id, activity, reqOptions)
+        : await this.api.conversations.activities(ref.conversation.id).update(activity.id, activity, reqOptions);
       return { ...activity, ...res };
     }
 
     const res = isTargeted
-      ? await api.conversations.activities(ref.conversation.id).createTargeted(activity)
-      : await api.conversations.activities(ref.conversation.id).create(activity);
+      ? await this.api.conversations.activities(ref.conversation.id).createTargeted(activity, reqOptions)
+      : await this.api.conversations.activities(ref.conversation.id).create(activity, reqOptions);
```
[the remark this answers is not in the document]
heyitsaamir: Correct — targeted sends use `RequestOptions<'serviceUrl'>` (no agenticIdentity). However, the per-turn Client is now constructed with `agenticIdentity` in its `ApiClientSettings`, and `AuthProviderInterceptor` uses `defaultAgenticIdentity` as fallback. So the auth header will still include the agentic token for targeted sends.

packages/apps/src/app.process.spec.ts
```diff
@@ -152,10 +153,9 @@ describe('App', () => {
 
       // Track what serviceUrl is used when sending
       let capturedServiceUrl: string | undefined;
-      const originalSend = app['activitySender'].send.bind(app['activitySender']);
-      jest.spyOn(app['activitySender'], 'send').mockImplementation((activity, ref) => {
+      jest.spyOn(ActivitySender.prototype, 'send').mockImplementation(async (activity, ref) => {
         capturedServiceUrl = ref.serviceUrl;
-        return originalSend(activity, ref);
+        return { ...activity, id: 'sent-1' };
       });
```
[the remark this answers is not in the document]
heyitsaamir: Fixed — added `jest.restoreAllMocks()` in `afterEach` to prevent spy leaks.

packages/apps/src/app.process.spec.ts
```diff
@@ -174,10 +174,9 @@ describe('App', () => {
       const serviceUrl2 = 'https://service-2.botframework.com';
 
       const capturedServiceUrls: string[] = [];
-      const originalSend = app['activitySender'].send.bind(app['activitySender']);
-      jest.spyOn(app['activitySender'], 'send').mockImplementation((activity, ref) => {
+      jest.spyOn(ActivitySender.prototype, 'send').mockImplementation(async (activity, ref) => {
         capturedServiceUrls.push(ref.serviceUrl);
-        return originalSend(activity, ref);
+        return { ...activity, id: 'sent-1' };
       });
```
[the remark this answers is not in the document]
heyitsaamir: Fixed — same `jest.restoreAllMocks()` in `afterEach`.

packages/apps/src/activity-sender.ts
```diff
@@ -32,23 +35,27 @@ export class ActivitySender implements IActivitySender {
...
+    };
+
     // Decide create vs update, with targeted variants
     if (activity.id) {
       const res = isTargeted
-        ? await api.conversations.activities(ref.conversation.id).updateTargeted(activity.id, activity)
-        : await api.conversations.activities(ref.conversation.id).update(activity.id, activity);
+        ? await this.api.conversations.activities(ref.conversation.id).updateTargeted(activity.id, activity, reqOptions)
+        : await this.api.conversations.activities(ref.conversation.id).update(activity.id, activity, reqOptions);
       return { ...activity, ...res };
```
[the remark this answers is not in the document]
heyitsaamir: Known limitation — targeted sends use `RequestOptions<'serviceUrl'>` (no agenticIdentity). However, the per-turn `Client` is constructed with `agenticIdentity` in its `ApiClientSettings`, and `AuthProviderInterceptor` uses `defaultAgenticIdentity` as fallback for all requests including targeted ones. Auth is handled correctly.

packages/apps/src/activity-sender.ts
```diff
@@ -32,23 +35,27 @@ export class ActivitySender implements IActivitySender {
...
     }
 
     const res = isTargeted
-      ? await api.conversations.activities(ref.conversation.id).createTargeted(activity)
-      : await api.conversations.activities(ref.conversation.id).create(activity);
+      ? await this.api.conversations.activities(ref.conversation.id).createTargeted(activity, reqOptions)
+      : await this.api.conversations.activities(ref.conversation.id).create(activity, reqOptions);
     return { ...activity, ...res };
```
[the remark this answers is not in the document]
heyitsaamir: Same as above — the Client's default agenticIdentity (via AuthProviderInterceptor) handles targeted creates. Per-request agenticIdentity override isn't needed since the Client is already scoped to this conversation turn's identity.

examples/agent365/src/main.ts
```diff
@@ -0,0 +1,44 @@
...
+import { ConsoleLogger } from '@microsoft/teams.common';
+
+const app = new App({
+  logger: new ConsoleLogger('@examples/agent365', { level: 'debug' }),
+});
+
+app.on('message', async ({ send, reply, activity, api, log }) => {
+  log.info(`[Agent365 reactive] Message received: ${activity.text}`);
+  log.info(`[Agent365 reactive] From: ${JSON.stringify(activity.from)}`);
+  log.info(`[Agent365 reactive] Recipient agentic identity: ${JSON.stringify(activity.recipient)}`);
```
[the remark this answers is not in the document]
heyitsaamir: Fixed — now logs only `activity.from?.id` and `activity.recipient?.id` instead of full objects.

packages/apps/src/app.ts
```diff
@@ -552,13 +559,29 @@ export class App<TPlugin extends IPlugin = IPlugin> {
...
+  }
+
+  /**
+   * Create an AgenticIdentity for use with agentic send helpers.
+   *
+   * @param agenticAppId the agentic app's ID
+   * @param agenticUserId the agentic user's ID
+   * @param opts optional overrides (tenantId defaults to this app's configured tenant)
+   */
+  getAgenticIdentity(agenticAppId: string, agenticUserId: string, opts?: { tenantId?: string }): AgenticIdentity {
```
corinagum: `getAgenticIdentity` diverges from PY.  `get_agentic_identity` raises if it can't resolve a tenant, and accepts `agentic_app_blueprint_id` . The TS version returns `{ tenantId: undefined }` silently, and there's no way to set the blueprint id. Worth noting the TS extractor `getAgenticIdentity(account)` in `account.ts` does carry `agenticAppBlueprintId` , so the factory is varied even within TS. Should a tenant guard + a `blueprintId` option be added to match PY?

corinagum: Also: PY resolves tenant from `credentials.tenantId`  vs TS  `options.tenantId` , are those equivalent?)

packages/apps/src/app.process.ts
```diff
@@ -43,7 +44,17 @@ export async function $process<TPlugin extends IPlugin>(
   }
 
   const client = this.client.clone();
-  const apiClient = new ApiClient(serviceUrl, this.client.clone({ token: () => this.getBotToken() }), this.options.apiClientSettings);
+  const agenticIdentity = getAgenticIdentity(activity.recipient);
```
corinagum: This collides with `app.ts`, will this cause confusion?

packages/apps/src/types/plugin/sender.ts
```diff
@@ -7,10 +7,15 @@ import { IStreamer } from '../streamer';
  * Separates sending concerns from transport concerns
  */
 export interface IActivitySender {
+  /**
+   * The agentic identity used for outbound requests (if any)
+   */
+  readonly agenticIdentity?: AgenticIdentity;
```
corinagum: Is anything meant to read this? Looks like this is not set anywhere.

packages/apps/src/app.ts
heyitsaamir: Good catch - 
1. https://github.com/microsoft/teams.py/pull/494 - updated Py to default with self.id if blueprint id is not provided
2. Updated this to accept blueprint id optionally and default to this.id if not provided.

packages/apps/src/types/plugin/sender.ts
heyitsaamir: You're right - no longer necessary (and in fact we should just remove this sender concept, it feels legacy)

corinagum (approved):

PR #634, opened 2026-07-06
feat(api): flatten client method chains to reduce call hops

lilyydu:
## Summary

Reduces the number of chained member accesses ("hops") needed to call the API by promoting grouped sub-client methods onto their parent clients. Common calls drop from **3 hops to 2**. The change is **additive and backward compatible** — the existing chained accessors remain as `@deprecated` aliases until officially removed.

## Flattening

| Old (deprecated, kept) | New |
| --- | --- |
| `users.token.{get,getAad,getStatus,signOut,exchange}` | `users.{getToken,getAadTokens,getTokenStatus,signOut,exchangeToken}` |
| `conversations.activities(id).{create,update,reply,delete,members,createTargeted,updateTargeted,deleteTargeted}` | `conversations.{createActivity,updateActivity,replyToActivity,deleteActivity,getActivityMembers,createTargetedActivity,updateTargetedActivity,deleteTargetedActivity}` |
| `conversations.members(id).{get,getById,getPaged,delete}` | `conversations.{getMembers,getMemberById,getPagedMembers}` |
| `client.reactions.{add,delete}` | `conversations.{addReaction,deleteReaction}` |

`teams.*` and `meetings.*` were already 2 hops — unchanged.

## Other changes

- **BotClient deprecated** — no longer used; the whole class and the `client.bots` accessor are marked `@deprecated` (not flattened).
- **Reactions moved** into the conversation/activity context (`conversations.addReaction(conversationId, activityId, type)`); `client.reactions` kept as a deprecated alias.
- **No internal use of deprecated members** — deprecated sub-client instances are held privately (`protected _token`/`_bots`/`_reactions`) and exposed via deprecated public accessors, so the flattened code and `Client.set http` never touch a deprecated member.
- **Internal callers migrated** — `@microsoft/teams.apps` now uses the flattened API.

## Tests

- New specs for the flattened methods on `UserClient` and `ConversationClient`.
- Full coverage of the deprecated chained aliases, co-located with the relevant sub-client specs.

## Verification

- `@microsoft/teams.api`: 20 suites / 194 tests ✓, lint ✓
- `@microsoft/teams.apps`: 19 suites / 284 tests ✓
- All essential (non-graph) packages build ✓

## Notes

- Deprecation notices intentionally omit a specific removal date ("in a future release"). Pre-existing deprecations elsewhere were left untouched.

### DIFF
--- a/packages/api/src/activities/message/message-reaction.ts
+++ b/packages/api/src/activities/message/message-reaction.ts
@@ -62,7 +62,7 @@ export class MessageReactionActivity
 
   /**
    * Add a message reaction.
-   * @deprecated Use the api.reactions.add instead.
+   * @deprecated Use api.conversations.addReaction instead.
    */
   addReaction(reaction: MessageReaction) {
     if (!this.reactionsAdded) {
@@ -75,7 +75,7 @@ export class MessageReactionActivity
 
   /**
    * Remove a message reaction.
-   * @deprecated Use the api.reactions.delete instead.
+   * @deprecated Use api.conversations.deleteReaction instead.
    */
   removeReaction(reaction: MessageReaction) {
     if (!this.reactionsRemoved) {
--- a/packages/api/src/clients/conversation/index.ts
+++ b/packages/api/src/clients/conversation/index.ts
@@ -5,9 +5,10 @@ import {
   type ClientOptions as HttpClientOptions
 } from '@microsoft/teams.common';
 
-import { Account, Conversation, ConversationResource } from '../../models';
+import { Account, Conversation, ConversationResource, MessageReactionType } from '../../models';
 
 import { ApiClientSettings, mergeApiClientSettings } from '../api-client-settings';
+import { ReactionClient } from '../reaction';
 
 import { ActivityParams, ConversationActivityClient } from './activity';
 import { ConversationMemberClient } from './member';
@@ -65,6 +66,7 @@ export class ConversationClient {
   protected _http: HttpClient;
   protected _activities: ConversationActivityClient;
   protected _members: ConversationMemberClient;
+  protected _reactions: ReactionClient;
   protected _apiClientSettings: Partial<ApiClientSettings>;
 
   constructor(serviceUrl: string, options?: HttpClient | HttpClientOptions, apiClientSettings?: Partial<ApiClientSettings>) {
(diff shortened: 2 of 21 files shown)


packages/api/src/activities/message/message-reaction.ts
heyitsaamir, line 74: does this do anything? should we just remove this since this is probably not used as it doesn't work

packages/api/src/clients/conversation/index.ts
```diff
@@ -65,6 +66,7 @@ export class ConversationClient {
   protected _http: HttpClient;
   protected _activities: ConversationActivityClient;
   protected _members: ConversationMemberClient;
+  protected _reactions: ReactionClient;
   protected _apiClientSettings: Partial<ApiClientSettings>;
```
heyitsaamir: let's make all these private since they're just implementation details.

packages/api/src/clients/conversation/index.ts
```diff
@@ -112,6 +125,104 @@ export class ConversationClient {
...
+
+  /**
+   * Get paged members of a conversation.
+   */
+  getPagedMembers(conversationId: string, pageSize?: number, continuationToken?: string) {
+    return this._members.getPaged(conversationId, pageSize, continuationToken);
+  }
+
+  /**
+   * @deprecated This will be removed in a future release.
```
heyitsaamir: is this intentional?

packages/api/src/clients/reaction/reaction.spec.ts
```diff
@@ -1,5 +1,7 @@
 import { Client } from '@microsoft/teams.common';
 
+import { Client as ApiClient } from '../index';
+
 import { ReactionClient } from './reaction';
```
heyitsaamir, line 5: Mark all the methods deprecated in this? And point to conversations.* methods?

heyitsaamir: Do similar stuff for activity and other clients as well?

heyitsaamir (commented):
Liking this direction!

packages/api/src/clients/reaction/reaction.spec.ts
lilyydu: woops! I thought I did that - ty for catching

packages/api/src/clients/index.ts
```diff
@@ -15,26 +15,41 @@ import { UserClient } from './user';
...
+   * `conversations.deleteReaction(...)` instead. This will be removed in a
+   * future release.
+   */
+  get reactions() {
+    return this._reactions;
+  }
   protected _http: HttpClient;
+  protected _bots: BotClient;
+  protected _reactions: ReactionClient;
   protected _apiClientSettings: Partial<ApiClientSettings>;
```
heyitsaamir, line 53: want to mark all these as deprecated as well similar to above

heyitsaamir (approved):

packages/api/src/activities/message/message-reaction.ts
lilyydu: I don't think so but tagging @corinagum to make sure - would it be breaking if we removed?

heyitsaamir (approved):

MehakBindra (approved):

PR #639, opened 2026-07-12
Split up Outbound and Inbound Activities

heyitsaamir:
Currently, `MessageActivity` / `TypingActivity` are used to represent both _outbound_ AND _inbound_ activities. These classes model the full activity shape, including fields that have no meaning on an outbound call. (`from`, `conversation`, `channelId`, `serviceUrl`, `timestamp`, etc.).

That creates confusing devex: callers can set fields that are ignored or overwritten by the send path, and it blurs the distinction between activities the app receives and activities the app sends.

In practice, apps only send two outbound activity types:

1. Message
2. Typing

This PR prototypes an outbound input model for those sendable activities:

- `ActivityInput` as the outbound base class
- `MessageActivityInput`
- `TypingActivityInput`
- `ActivityParams = IMessageActivityInput | ITypingActivityInput`

The inbound activity/router model stays unchanged, but app-facing send/create surfaces are now centered around outbound activity inputs.

## Backward compatibility

Now, existing code often does:

```ts
ctx.send(new MessageActivity().withText("hi"))
```

To keep that working, this PR keeps accepting `MessageActivity` and `TypingActivity` on send surfaces through deprecated overloads. When those legacy builders are passed to `send`, they are converted into `MessageActivityInput` / `TypingActivityInput` before sending.

That gives users migration guidance without breaking existing code:

So now, this is the more correct code:
```ts
ctx.send(new MessageActivityInput().withText("hi"))
```
But this continues to work:
image

The conversion intentionally drops server-populated fields like `from`, `conversation`, `channelId`, and `serviceUrl` so the wire payload matches what the app is actually allowed to send.

## Non-sendable activity cleanup

`ActivityParams` remains the outbound input union. Non-sendable activity types such as `messageUpdate`, `messageDelete`, `messageReaction`, and invoke feedback are not accepted by activity create/send APIs, so we removed those.

Call sites that previously tried to create those activity types were updated to use the appropriate APIs instead, such as update, delete, and reaction endpoints.


packages/api/src/activities/activity.ts
```diff
@@ -119,6 +119,243 @@ export interface IActivity<T extends string = string> {
...
+  withRecipient(value: Account, isTargeted: boolean = false) {
+    this.recipient = { ...value, isTargeted: isTargeted ? true : undefined };
+    return this;
+  }
+
+  withReplyToId(value: string) {
+    this.replyToId = value;
+    return this;
+  }
+
```
heyitsaamir, line 233: These convenient methods that only apply to outbound activities should be deprecated on Activity

packages/api/src/activities/typing.ts
```diff
@@ -7,6 +7,94 @@ export interface ITypingActivity extends IActivity<'typing'> {
...
+      entities: activity.entities,
+      channelData: activity.channelData,
+      text: activity.text,
+    });
+  }
+
+  /**
+   * The text content of the message.
+   */
+  withText(value: string) {
```
heyitsaamir, line 86: Can typing have text? For streaming?

packages/apps/src/activity-sender.ts
```diff
@@ -14,36 +21,42 @@ export class ActivitySender implements IActivitySender {
...
+  async send(activity: ActivityParams | DeprecatedInputActivity, ref: ConversationReference): Promise<SentActivity> {
     // Create API client for this conversation's service URL
     const api = new Client(ref.serviceUrl, this.client);
+    const params = toActivityParams(activity);
 
-    // Merge activity with conversation reference
-    activity = {
-      ...activity,
+    // Merge activity with conversation reference for the wire payload.
+    const payload = {
```
heyitsaamir, line 36: Can payload be Imessageactivityinput?

packages/apps/src/http/http-stream.ts
```diff
@@ -408,29 +409,30 @@ export class HttpStream implements IStreamer {
       throw new StreamCancelledError();
     }
 
-    activity = {
-      ...activity,
+    const params = toActivityParams(activity);
+    const payload = {
```
heyitsaamir, line 458: Same

packages/devtools/src/components/Feedback/Feedback.tsx
```diff
@@ -111,26 +107,8 @@ const Feedback: FC<FeedbackProps> = ({
         childLog.error('Missing message id');
         return;
       }
-      const originalActivity = findByMessageId(value.id);
```
heyitsaamir, line 114: What would be the impact of not changing anything in devtools package?

packages/api/src/activities/activity.ts
```diff
@@ -119,6 +119,243 @@ export interface IActivity<T extends string = string> {
...
+ * fields such as `from`, `conversation`, `channelId`, `serviceUrl`, and timestamps are
+ * filled by the sender or service, not by callers constructing outbound input.
+ */
+export interface IActivityInput<T extends string = string> {
+  readonly type: T;
+  id?: string;
+  recipient?: Account;
+  replyToId?: string;
+  entities?: Entity[];
+  channelData?: ChannelData;
```
heyitsaamir, line 159: include js docs (for this and below, and other new public facing classes / interfcaes)

packages/api/src/activities/message/message.ts
```diff
@@ -107,6 +107,305 @@ export interface IMessageActivity extends IActivity<'message'> {
...
+
+  /**
+   * Suggested actions presented with the message.
+   */
+  suggestedActions?: SuggestedActions;
+
+  /**
+   * Delivery hint for the message.
+   */
+  deliveryMode?: DeliveryMode;
```
MehakBindra: We have removed this from .net , it has no effect

packages/api/src/activities/message/message.ts
```diff
@@ -107,6 +107,305 @@ export interface IMessageActivity extends IActivity<'message'> {
...
+export interface IMessageActivityInput extends IActivityInput<'message'> {
+  /**
+   * Message text.
+   */
+  text?: string;
+
+  /**
+   * Summary text displayed when the channel cannot render the full activity.
+   */
+  summary?: string;
```
MehakBindra: summary for input ?

packages/api/src/activities/message/message.ts
```diff
@@ -107,6 +107,305 @@ export interface IMessageActivity extends IActivity<'message'> {
...
+
+  /**
+   * Delivery hint for the message.
+   */
+  deliveryMode?: DeliveryMode;
+
+  /**
+   * Channel-specific or caller-defined value associated with the message.
+   */
+  value?: any;
```
MehakBindra: What is this field for ?

heyitsaamir: "Message-back" actions (it's legacy, but still works)

packages/api/src/activities/typing.ts
```diff
@@ -7,6 +7,106 @@ export interface ITypingActivity extends IActivity<'typing'> {
...
+    if (!this.channelData.streamId) {
+      this.channelData.streamId = this.id;
+    }
+
+    if (!this.channelData.streamType) {
+      this.channelData.streamType = 'streaming';
+    }
+    if (!this.channelData.streamSequence) {
+      this.channelData.streamSequence = sequence;
+    }
```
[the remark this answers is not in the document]
heyitsaamir: I'm not changing any behavior here right? Doesn't this bug already exist then? Or is this even a bug?

packages/api/src/activities/message/message.ts
```diff
@@ -107,6 +107,305 @@ export interface IMessageActivity extends IActivity<'message'> {
...
+   * Mark the message as the final activity in a stream.
+   */
+  addStreamFinal() {
+    const { streamId } = this.channelData || {};
+
+    this.addEntity({
+      type: 'streaminfo',
+      streamId: streamId || this.id || '',
+      streamType: 'final',
+      streamSequence: this.channelData?.streamSequence || 1,
```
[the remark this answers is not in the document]
heyitsaamir: they don't start at 0. they start at 1.

packages/devtools/src/screens/ChatScreen/ChatScreen.tsx
```diff
@@ -76,30 +66,22 @@ const ChatScreen: FC<ChatScreenProps> = ({ isConnected }) => {
...
-                  : undefined,
-              });
-              await teamsApi.conversations.activities(chat.id).create(updateActivity);
+              await teamsApi.conversations.updateActivity(
+                chat.id,
+                action.id,
+                new MessageActivityInput(originalMessage.body?.content || '', {
+                  attachments: originalMessage.attachments || [],
+                  value: messageBody,
+                })
```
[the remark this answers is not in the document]
heyitsaamir: devtools is deprecated.

MehakBindra (approved):

PR #653, opened 2026-07-17
fix(apps): add opt-out for per-activity user token lookup

### LINKED ISSUE #652: [Bug]: user token is fetched on every inbound activity even when `ctx.userGraph` is never used
### Summary

`app.process.ts` calls the Bot Framework user-token service on **every inbound activity**, before the context is built, even when the app never reads `ctx.userGraph`:

https://github.com/microsoft/teams.ts/blob/main/packages/apps/src/app.process.ts#L81-L87

The fetched `userToken` has effectively one consumer (the `userGraph` client / `isSignedIn`). For apps that don't call Graph *as the user* — especially those with no OAuth connection provisioned for the default `graph` connection — this is a wasted round-trip on the hot path of every turn that returns `404` every time and marks each httpx/OTel span as an error.

This is the TypeScript counterpart of the Python bug: microsoft/teams.py#506 (full analysis there).

### Expected

Don't issue the request when its result is unused — either fetch lazily on first access of `ctx.userGraph` / `ctx.userToken`, or add an opt-out. TS is already close: `userGraph` receives the token as a lazy thunk (`token: () => userToken`); it just resolves it eagerly first.

### Notes

- No opt-out today: `skipAuth` governs only inbound JWT validation.
- The Python SDK is addressing this with an auto-detected `fetch_user_token` flag (skips the lookup unless an OAuth connection is explicitly configured); a parallel approach fits TS.

corinagum:
## Problem

Fixes #652

Every inbound activity triggers an eager Bot Framework `getUserToken` call, even when the app never uses `ctx.userGraph` / `ctx.isSignedIn`. For apps with no OAuth connection this adds latency per turn and emits ERROR spans (404s from the token service).

## Change (non-breaking)

Adds a `fetchUserToken` option to `OAuthSettings` that gates the eager token lookup.

- **Auto-detect default:** when unset, the lookup is skipped unless OAuth is explicitly configured (i.e. `oauth.defaultConnectionName` was provided by the user). Apps with no OAuth connection stop paying the cost with no config change.
- **Explicit override:** set `oauth.fetchUserToken: true` to force the lookup, or `false` to always skip it.

OAuth-configured apps are unchanged.

## Edge case

Apps that rely on the default `"graph"` connection *without* explicitly passing `oauth.defaultConnectionName`, and read `ctx.isSignedIn` / `ctx.userGraph`, will now see the fetch skipped. They must set `oauth.fetchUserToken: true`.

## Verification

- `eslint` clean; `tsc` build ok; all apps tests pass (added a `user token lookup` describe block).
- In-process transport check (no network): default app = **0** GetToken calls, OAuth app = **1** (unchanged), explicit opt-out = **0**.

The breaking full-laziness change (fetch on first access of `userGraph`, deprecate `isSignedIn`/`userToken`) is deferred to 2.1.

### DIFF
--- a/packages/apps/src/app.process.spec.ts
+++ b/packages/apps/src/app.process.spec.ts
@@ -349,4 +349,62 @@ describe('App', () => {
       expect(errors[0].activity).toBeDefined();
     });
   });
+
+  describe('user token lookup', () => {
+    const userActivity = new MessageActivity('hi', {
+      id: 'a1',
+      channelId: 'msteams',
+      from: { id: 'user-1', name: 'User' },
+      conversation: { id: 'c1' },
+      recipient: { id: 'bot' },
+    } as Partial<IMessageActivity>);
+
+    it('does not fetch the user token when no OAuth connection is configured', async () => {
+      const testApp = createTestApp();
+      testApp.start();
+      const spy = jest.spyOn(testApp.api.users, 'getToken');
+
+      await testApp.process({ token, body: userActivity });
+
+      expect(spy).not.toHaveBeenCalled();
+      testApp.stop();
+    });
+
+    it('fetches the user token when an OAuth connection is configured', async () => {
+      const testApp = createTestApp({ oauth: { defaultConnectionName: 'graph' } });
+      testApp.start();
+      const spy = jest
+        .spyOn(testApp.api.users, 'getToken')
+        .mockResolvedValue({ token: 'user-token' } as any);
+
+      await testApp.process({ token, body: userActivity });
+
+      expect(spy).toHaveBeenCalledTimes(1);
+      testApp.stop();
+    });
+
+    it('honors an explicit fetchUserToken=false override even when OAuth is configured', async () => {
+      const testApp = createTestApp({ oauth: { defaultConnectionName: 'graph', fetchUserToken: false } });
+      testApp.start();
+      const spy = jest.spyOn(testApp.api.users, 'getToken');
+
+      await testApp.process({ token, body: userActivity });
+
+      expect(spy).not.toHaveBeenCalled();
+      testApp.stop();
+    });
+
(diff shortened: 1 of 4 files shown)


heyitsaamir:
* `fetchUserToken` -> `automaticallyFetchUserToken`?
* Ideally, I think we should try and deprecate this behavior. It's a poor default we have.
* I think a workaround we can have by default is to try once (the first first time a request comes in), if the connection isn't found, keep that in state, and never try again. This way we pay the price once at-most without breaking everyone.

corinagum:
@heyitsaamir 
1. Leaning towards keeping `fetchUsertoken`. It's a simple on/off for the fetch, and "automatic" is regarding the default behavior (whether an OAuth connection is configured), not the flag's meaning to force a fetch. 
2. Yes, hence breaking change deferred to 2.1
3. We can't tell "connection not fround" apart from "user not signed in"
  - teams.ts  app.oauth.ts  and teams.net /core `OAuthFlow.cs` both bucket 404/400/412 together as "no token / show sign-in card," and only treat other codes as distinguishable failures. Neither has any notion of "connection missing" as a separate case.
  - Did a live probe against `token.botframework.com`, three cases:
      - connection not configured > 404, empty body
      - connection exists (real Aadv2 connection) + user not signed in > 404, empty body
      - bogus name > 404, empty body
  - All three are identical: bare 404, empty body, no error code or message. There's no `TokenResponse` payload to inspect.
 - Caching on the 404 would inevitably cache legitimate not-signed-in users and permanently break their sign-in after their first pre-auth message. The condition of app-level "connection missing" is indistinguishable from a transient user-level state at this boundary.

Proposed: land this PR as-is, keep to 2.1 for the deprecation, and park the negative-cache idea pending the connection-level approach.

MehakBindra (approved):

PR #654, opened 2026-07-20
Add OpenTelemetry HTTP middleware

heyitsaamir:
Adds the first cut of OpenTelemetry observability across the TS Teams SDK runtime.

## Why

This gives developers a real trace through the SDK: inbound activity → handler → outbound Teams API call → auth. Makes it way easier to debug whether Teams called us, whether the handler ran, whether we sent the reply, and which auth flow was used.

It also sets up the Agent 365 story. Teams SDK spans are the useful diagnostic detail; Agent 365 product visibility still comes from explicit A365 scopes like invoke_agent / execute_tool / chat.

## Interesting bits

- Adds public common HTTP middleware support so API telemetry can wrap auth + transport cleanly.
- Adds API outbound telemetry middleware for `microsoft.teams.api.client`.
- Emits `microsoft.teams.auth.outbound` from the API auth token callback.
- Adds Apps spans/metrics for `microsoft.teams.activity.process` and `microsoft.teams.handler`.
- Adds Agent365-compatible baggage from inbound Teams activities.
- Adds package version as the OTel instrumentation scope version for `Microsoft.Teams.Api` and `Microsoft.Teams.Apps`.
- Keeps the generic HTTP middleware mechanics-only. API methods own semantic attrs and response hooks.
- Restores legacy `RequestOptions`, `AuthProviderInterceptor`, and `ApiClient.http` setter compatibility after the middleware refactor.

TS/Python intentionally differ from .NET right now: .NET still has Core-shaped names like `Microsoft.Teams.Core`, `turn`, and `conversation_client`. This PR uses the newer API/Apps split and `microsoft.teams.api.client` middleware model.

## Reviewer tips

Start with common HTTP middleware, then API outbound middleware, then conversation activity methods that attach semantic attrs/hooks. The compatibility commit is worth reviewing too: it preserves per-request serviceUrl/agenticIdentity and custom HTTP client behavior.

## Testing

- `@microsoft/teams.common` tests passed.
- `@microsoft/teams.api` targeted tests and build passed.
- `@microsoft/teams.apps` targeted tests and build passed.
- Live Teams smokes passed for Agent365 / ACF and Echo / PokemonCatcher.
- Aspire inspection confirmed traces + metrics, expected span nesting, API/auth spans, and response-derived activity IDs.
- Fixed issues found during smoke/review: Agent365 pre-handler auth now uses the scoped agentic API client, and legacy API client compatibility is restored.


heyitsaamir:
Superseded by the smaller Git Spice stack: #655 (API), #656 (Apps), and #657 (Agent365 docs/examples).

PR #655, opened 2026-07-20
Add OpenTelemetry API middleware

heyitsaamir:
Adds the API-side plumbing for Teams SDK OpenTelemetry: common HTTP middleware, outbound API spans, and auth spans.

## Why

This is the foundation layer. It lets us see what the SDK is doing when it calls Teams APIs: which operation ran, which auth flow was used, whether the call failed, and how it lines up under the app’s trace.

Keeping this in the API/common layer also means Apps can stay focused on inbound activity processing instead of knowing HTTP client internals. Tiny miracle, big payoff.

## Interesting bits

- Adds public common HTTP middleware support.
- Adds `microsoft.teams.api.client` for outbound Teams API calls.
- Adds `microsoft.teams.auth.outbound` from the API auth token callback.
- Uses `Microsoft.Teams.Api` as the instrumentation scope, with the package version attached as the scope version.
- Keeps the outbound middleware mechanics-only: start/end span, record calls/errors, record exceptions, and run the optional response hook.
- Client methods own semantic attributes and response-derived tags, including response `activity.id`.
- Internal auth should use the token callback path, not `AuthProviderInterceptor`; interceptor can remain only as public compatibility.
- Keeps scoped service URL / agentic identity and custom `ApiClient.http` behavior working.

## Reviewer tips

Start with common HTTP client middleware, then the API outbound middleware, then conversation activity methods. The scoped client and custom HTTP bits are worth a close look because they keep current scoping behavior working while routing auth through the new token callback path.

## Testing

CI is expected to cover this layer. The full stack was also live-smoked against Agent365/ACF and Echo/PokemonCatcher before splitting.


heyitsaamir:

- #627
    - #655 ◀

packages/common/src/http/client.ts
```diff
@@ -290,9 +376,141 @@ export class Client {
...
+      return this.http.request<T, R, D>(resolvedConfig);
+    }
+
+    const transportConfig: RequestConfig<D> = { ...resolvedConfig };
+    delete transportConfig.method;
+    delete transportConfig.url;
+    delete transportConfig.data;
+
+    switch (method) {
+      case 'get':
```
corinagum, line 391: This drops any body passed via `client.get(url, {data})` / `delete(url, {data})`. axios only reads GET/DELETE bodies from  `config.data`. The old path preserved it. Since `@microsoft/teams.common` is public, this silently changes the contract. Keep data on `transportConfig`  for get/delete, or fall through to `http.request(resolvedConfig)` when data is present

packages/api/src/clients/api-outbound-middleware.ts
```diff
@@ -0,0 +1,107 @@
...
+      {
+        kind: SpanKind.CLIENT,
+        attributes: metadata.attributes,
+      },
+      async (span) => {
+        recordTeamsApiOutboundCall(metadata.operation);
+
+        try {
+          const res = await next();
+          await metadata.onResponse?.(span, res as AxiosResponse);
```
corinagum: This runs inside the same try as `next()`, so a throwing response hook discards the successful response, increments the error counter, and propagates to the caller. Wrap `onResponse` in its own try/catch that logs and swallows so telemetry failures can't fail the API call.

packages/common/src/http/client.ts
heyitsaamir: Fair, kept it the same as before to reduce churn

packages/api/src/clients/api-outbound-middleware.ts
heyitsaamir: Good catch — split these apart so the hook no longer shares `next()`'s try. `next()` now has its own try that records the outbound error + exception on the span; `onResponse` runs in a separate try that just logs and swallows any failure. So a throwing/rejecting hook can't discard the successful response or get counted as an outbound error anymore, and the outer try is only there for `finally { span.end() }`. Added a regression test that asserts a throwing hook doesn't reject the call, doesn't record an error/exception, and still ends the span once.

corinagum (approved):

PR #656, opened 2026-07-20
Add OpenTelemetry app diagnostics

heyitsaamir:
Adds the Apps-side OpenTelemetry layer: inbound activity processing and handler telemetry.

## Why

This gives developers the other half of the story: Teams called the app, the SDK processed the activity, and the right handler ran. When paired with the API layer below it, the trace reads like: inbound activity → handler → outbound Teams API → auth. Much nicer than guessing from scattered logs.

## Interesting bits

- Adds `microsoft.teams.activity.process` around inbound activity processing.
- Adds `microsoft.teams.handler` around handler dispatch.
- Adds Apps metrics for activity processing and handler dispatch/duration.
- Uses `Microsoft.Teams.Apps` as the instrumentation scope, with the package version attached as the scope version.
- Restores proactive send/reply scoping compatibility where the middleware refactor touched it.
- Leaves Agent365-specific integration for a separate follow-up.

## Reviewer tips

Start with the app processing path, then check handler telemetry and compatibility. This PR stacks on the API layer, so API outbound spans live in the parent PR.

## Testing

CI is expected to cover this layer. The full stack was also live-smoked against Agent365/ACF and Echo/PokemonCatcher before splitting, with Aspire showing the expected Apps/API span nesting.


heyitsaamir:

- #627
    - #655
        - #656 ◀

packages/apps/src/diagnostics/baggage.ts
```diff
@@ -0,0 +1,265 @@
...
+      .userId(activity.from?.aadObjectId)
+      .userName(activity.from?.name)
+      .userEmail(activity.from?.email)
+      .agentId(agentId)
+      .agentName(activity.recipient?.name)
+      .agenticUserId(activity.recipient?.agenticUserId)
+      .agentBlueprintId(activity.recipient?.agenticAppBlueprintId)
+      .agenticUserEmail(activity.recipient?.email)
+      .agentDescription(activity.recipient?.userRole);
+  }
```
corinagum: fromActivity  puts raw caller PII into baggage and `app.process.ts` scopes that baggage over the whole handler pipeline (`next()` , `context.send`, Graph, and any third-party HTTP a handler makes).

I know OTel is fully opt-in and the SDK configures no propagators, so nothing leaves the process unless a consumer stands up an OTel SDK. My concern is the coupling once they do:  `@opentelemetry/sdk-node`  defaults to a `CompositePropagator` of `W3CTraceContextPropagator` + `W3CBaggagePropagator`, and `instrumentation-http` injects the active context into outbound headers. So, the common copy-paste setup turns "I enabled tracing" into "user email/name/id now ride a `baggage:`  header to every downstream call, including non-Microsoft endpoints my handler hits." **The consumer opted into observability, not PII-over-the-wire.**

**span attributes are local (exporter-only, never propagated); baggage is a wire channel**. Correlation IDs (tenant / conversation / agent id) are reasonable in baggage; name/email/user-id probably shouldn't be, since that's the propagating channel.

corinagum (changes requested):

packages/apps/src/diagnostics/baggage.ts
heyitsaamir: Totally fair callout. I was following what's done in .net and figured this happened there already. In any case, I've removed this logic from this PR and kept it simple Otel for Teams SDK for now. I'm adding Agent 365 support in the next PR to keep it all together.

heyitsaamir: Moved this up the stack into a separate PR now.

Agent365 has certain requirements for what needs to be logged (user.id, etc.), which will be on by default — a developer opting into Agent365 observability is an indication of accepting the required fields. Anything extra (like email, which is clearly PII) will be explicit opt-in.

corinagum: Sounds good. Yeah, user id without associated PII is fine.

corinagum (approved):

PR #679, opened 2026-07-30
Wire the agent365 example onto the OpenTelemetry SDK surface

heyitsaamir:
## First Why is this diff MASSIVE?

It's because [microsoft-otel distro](https://www.npmjs.com/package/@microsoft/opentelemetry?activeTab=readme) introduces a massive chain of dependencies for some reason. I wanted to split this PR in its own to demonstrate that it's just 1 dep that's causing this massive diff.

## Summary

Rebuilds `examples/agent365` on top of the SDK's OpenTelemetry surface (added in the base PR) and pulls in the one dependency that wiring needs.

- Bootstraps the **Microsoft OpenTelemetry distro** as the app's exporter/OTel SDK — the SDK itself stays generic, so the distro lives here in the example.
- Rebuilds the example around the three real flows: **exporter auth**, **reactive turns**, and **proactive sends**, using `App.tokenProvider` and `createAgent365Scope`.

## Reviewer tips

- `src/observability.ts` is the exporter/auth bootstrap.
- `src/main.ts` is the reactive path; `src/proactive.ts` the proactive one, including the final `flushAgent365Spans()` so the batch processor drains before exit.
- The reactive path shows that the Teams App automatically fills the baggage for you per-turn
- The proactive path shows how you can use the helpers and the app token minting to include manual governance into your own app.

## Validation

Run live against a real tenant, which CI can't do: a Teams message to the agentic user round-tripped and its `invoke_agent` span exported (`rejectedSpans: 0`, one identity group), and the proactive job delivered through both `app.send` and the conversation activity API with its span exported the same way.

That live run caught two bugs unit tests didn't (both fixed in the SDK PR): `App.getAgenticUser` read `options.tenantId` rather than resolved credentials, so any app configured via `TENANT_ID` threw; and the proactive job exited before the batch processor flushed, dropping the span it had just created.

What it doesn't prove: the export response says the service accepted the spans, not that they surface in a customer-queryable table. Portal-side visibility is Purview DSPM, on its own ingestion delay.


examples/agent365/README.md
```diff
@@ -1,6 +1,18 @@
 # AgenticBlueprint and Agentic User example
 
-Demonstrates using an AgenticBlueprint to send and receive messages as an Agentic User in the Agentic User program.
+Demonstrates using an AgenticBlueprint to send and receive messages as an Agentic User, and how to wire the resulting telemetry into Agent365 observability.
```
corinagum: Maybe link to OTel official docs too?

examples/agent365/src/proactive.ts
```diff
@@ -29,26 +40,53 @@ async function main() {
...
+
+        scope.recordOutputMessages('Proactive digest sent.');
+      });
+    } finally {
+      scope.dispose();
+    }
+  });
+
+  // Spans are batched, so a job that exits here would take its span with it.
+  await flushAgent365Spans();
```
corinagum: This should probably be in a `finally` block so it's run on failure too

MehakBindra (approved):

corinagum (approved):

PR #682, opened 2026-07-31
(Re)Introduce AgenticIdentity operation scope

heyitsaamir:
## Summary

Use `AgenticIdentity` as the TypeScript Agent 365 operation scope for proactive sends, scoped API clients, and request baggage.

## Why

Agent 365 starts from a blueprint, which can create apps, which can optionally have users. So `agenticAppBlueprintId` is required, while `agenticAppId` and `agenticUserId` stay optional until a token flow actually needs them.

## Interesting bits

- No `kind` discriminator; field presence is enough for now.
- Token helpers stay specific: `getAgenticUserToken` and `getAgenticAppToken`.
- Lifecycle stays Agentic User-specific because the service events are about agentic users.
- Service wire names stay as-is, including lifecycle `agenticAppInstanceId` and `agentIdentityBlueprintId`.
- Auth diagnostics now report `app_only`, `agentic_user`, or `agentic_app`.

## Reviewer tips

Start with the `AgenticIdentity` model, API client token dispatch, token manager/provider, and lifecycle event types. The main thing to watch is the intentional split between SDK-facing names and service-owned wire names.


packages/apps/src/token-manager.ts
```diff
@@ -85,54 +85,64 @@ export class TokenManager {
...
+      return this.toProviderToken(await getAgenticIdentityToken(scope, agenticIdentity, tenantId));
     }
 
-    this.requireClientCredentials(this.credentials, 'Agentic User');
+    this.requireClientCredentials(this.credentials, 'AgenticIdentity');
 
-    const agenticAppInstanceId = agenticUser.agenticAppInstanceId;
-    const blueprintAssertion = this.blueprintAssertionFor(agenticAppInstanceId, tenantId);
+    if (!isUserBackedAgenticIdentity(agenticIdentity)) {
+      const { token } = await this.acquireAgenticAppToken(
```
corinagum: Same as PY: suggest validating against empty or whitespace user id's

corinagum (approved):
1 nit, otherwise lgtm!

MehakBindra (approved):

PR #683, opened 2026-07-31
feat: enforce 7-day package-age gate for Dependabot PRs

corinagum:
## Summary

Enforces a 7-day minimum package-age policy on Dependabot PRs so we never merge an npm version that is still inside the Microsoft **Central Feed Services (CFS)** quarantine window. CFS defers newly-published npm versions for 7 days (measured from the public npm publish time); during quarantine the version is not served by the Azure Artifacts feed, so merging a fresh package breaks `npm ci` (E404) for the whole repo.

## What changed

- **`.github/dependabot.yml`** — enables npm version updates with a 7-day `cooldown` (major/minor/patch), delaying version-update PRs until the target release is ≥7 days old.
- **`.github/workflows/dependabot-package-age-gate.yml`** — publishes a `package-age/7-day` commit status on every Dependabot PR. Runs on `pull_request_target` (opened/synchronize/reopened), a daily schedule (so PRs that were too new auto-clear once packages age out), and `workflow_dispatch`.
- **`.github/scripts/dependabot-pr-age-check.mjs`** — computes eligibility from the `package-lock.json` diff (every added version, including transitive) and sets the status: `success` when all changed versions are ≥7 days old, `pending` when a version is <7 days old (description shows the eligible date), and `failure` when age can't be confirmed (fail-safe: blocks merge).

## Why `cooldown` alone isn't enough

`dependabot.yml` `cooldown` only applies to version updates. Dependabot security updates ignore cooldown and open immediately with the fixed version — which may be <7 days old and break `npm ci`. The gate covers all Dependabot PRs (security + version), closing that hole.

## ⚠️ Required follow-up

Enforcement is inert until `package-age/7-day` is added as a required status check in branch protection for `main`. Until then the status posts but does not block merges.

## Testing

Validated the script end-to-end against mocked GitHub + npm responses: all-old→success, young top-level→pending (+eligible date), young transitive dep→pending, missing npm timestamp→failure, registry 404→failure.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: "/"
+    schedule:
+      interval: daily
+    cooldown:
+      default-days: 7
+      semver-major-days: 7
+      semver-minor-days: 7
+      semver-patch-days: 7
--- a/.github/scripts/dependabot-pr-age-check.mjs
+++ b/.github/scripts/dependabot-pr-age-check.mjs
@@ -38,11 +38,18 @@ async function main() {
   const pullRequests = await getTargetPullRequests(context);
 
   for (const pullRequest of pullRequests) {
-    if (!isDependabotPullRequest(pullRequest)) {
-      continue;
-    }
-
     try {
+      if (!isDependabotPullRequest(pullRequest)) {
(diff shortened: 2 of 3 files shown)


rajan-chari (approved):
7 day wait for dependabot PRs.  Makes sense.  Thanks Corina.

Issue #351, opened 2025-09-16
SingIn event triggered more that once


rido-min:
After a successful OAuth flow, the event `signin` is triggered multiple times.

https://github.com/microsoft/teams.ts/blob/9b8507d2f2d6b6c1db619a8a0cccbea12a85f291/tests/graph/src/index.ts#L41-L44

based on the number of clients that are already receiving the OAuth card we will see more duplicated events.

PR #686, opened 2026-08-01
fix: deduplicate token exchange invokes to prevent multiple signin events

jibin7jose:
**Fixes:** #351 

**Description:**
When multiple clients (Desktop, Web, Mobile) receive the OAuth card simultaneously, they each independently attempt a silent token exchange by sending a `signin/tokenExchange` invoke activity. 

Previously, `app.oauth.ts` blindly emitted the `signin` event for each successful exchange without verifying if the token exchange `id` had already been processed. This resulted in the `signin` event firing multiple times for a single login flow.

This PR implements standard deduplication by caching processed `activity.value.id` strings in a short-lived, rolling array (max length 1000). If subsequent invoke activities carry an already processed `id`, the framework now gracefully ignores them and responds with an HTTP `200 OK` status, preventing duplicate event emissions while remaining memory safe.

Co-authored-by: Jibin Jose <jibinjose884@gmail.com>


packages/apps/src/app.oauth.ts
```diff
@@ -41,7 +44,19 @@ export class OauthHandlers<TPlugin extends IPlugin = IPlugin> {
...
+    const exchangeId = activity.value.id;
+    if (exchangeId) {
+      if (this.processedExchangeIds.includes(exchangeId)) {
+        return { status: 200 };
+      }
+      const existingLock = this.tokenExchangeLocks.get(exchangeId);
+      if (existingLock) {
+        await existingLock.catch(() => {});
+        return { status: 200 };
+      }
```
corinagum: This will always return HTTP 200. If the original exchange returns 412, other callers should also receive that failure. Otherwise fallback authentication might be skipped.

packages/apps/src/app.oauth.ts
```diff
@@ -22,6 +22,9 @@ import { PluginAdditionalContext } from './types/app-routing';
  * class fields so they can be passed directly as route callbacks.
  */
 export class OauthHandlers<TPlugin extends IPlugin = IPlugin> {
+  private readonly processedExchangeIds: string[] = [];
+  private readonly tokenExchangeLocks = new Map<string, Promise<void>>();
```
corinagum: This does not prevent duplicates across replicas or restarts. However, distributed state is needed in order to implement. Could you please add the following comment?
``// TODO: Implement production de-duplication with state to prevent duplicates across replicas or restarts

corinagum (changes requested):
Thanks for the PR, @jibin7jose. Could you please see comments and add some unit tests? TIA!

jibin7jose:
Ok

corinagum:
@jibin7jose thanks for the updates! Could you do a pass and resolve the addressed/stale comments? Looks like there's a few cleanup opportunities. :) Thanks in advance!

jibin7jose:
I've resolved all the addressed comments.

packages/apps/src/app.oauth.ts
```diff
@@ -72,54 +76,91 @@ export class OauthHandlers<TPlugin extends IPlugin = IPlugin> {
...
-              recordUnexpectedOAuthError(span, telemetry, error, APP_OAUTH_ERROR_TYPE.httpError);
-              this.events.emit('error', { error, activity });
-              return { status: error.status || 500 };
-            }
-          } else {
-            recordUnexpectedOAuthError(span, telemetry, error, APP_OAUTH_ERROR_TYPE.exception);
-          }
+            this.events.emit('signin', { ...ctx, token, isSignedIn: true });
+            telemetry.callbackInvoked = true;
+            next(ctx);
```
corinagum:
```suggestion
            await next(ctx);
```

packages/apps/src/app.oauth.spec.ts
```diff
@@ -22,6 +23,170 @@ import {
...
+        next
+      };
+      
+      const results = await Promise.all([
+        handlers.onTokenExchange(ctx),
+        handlers.onTokenExchange(ctx)
+      ]);
+      
+      expect(results).toEqual([{ status: 200 }, { status: 200 }]);
+      expect(mockApi.users.exchangeToken).toHaveBeenCalledTimes(1);
```
corinagum, line 151: Would be good to verify `signin` and `next(ctx)` also run exactly once.

packages/apps/src/app.oauth.ts
jibin7jose: I've updated this to await next(ctx)
 here and in the other OAuth handlers as well.

packages/apps/src/app.oauth.spec.ts
jibin7jose: added assertions to verify that both the signin event and next(ctx) run exactly once for successful concurrent requests, and zero times for the failure case

corinagum (approved):
Thank you for the PR!

PR #688, opened 2026-08-02
fix(api): the final streamed message carries no streamSequence

### LINKED ISSUE #687: [Bug]: the final streamed message carries streamSequence, which the Teams contract forbids there
### Bug Description

`MessageActivityInput.addStreamFinal()` always attaches a `streamSequence` to the `streaminfo` entity, and `HttpStream.close()` builds the final message with `MessageActivityInput` — so every final the streamer sends carries it.

The two `addStreamFinal()` implementations in `message.ts` disagree, and the one the streamer uses is the non-conforming one:

```js
// MessageActivityInput.addStreamFinal()   <-- used by HttpStream.close()
const streamSequence = this.channelData.streamSequence ?? 1;
this.channelData.streamSequence = streamSequence;
this.addEntity({ type: 'streaminfo', streamId, streamType: 'final', streamSequence });

// MessageActivity.addStreamFinal()        <-- already correct
return this.addEntity({ type: 'streaminfo', streamId: this.id, streamType: 'final' });
```

The Teams streaming contract says the opposite. From [Stream agent messages](https://learn.microsoft.com/en-us/microsoftteams/platform/bots/streaming-ux#stream-message-through-rest-api), under the `entities.streamSequence` property:

> Here are the requirements for using `streamSequence` for REST APIs:
> - First one must be number '1'.
> - Subsequent numbers (except final) must be a monotonic increasing integer (for example, 1->2->3).
> - **For the final message, `streamSequence` must not be set.**

The documented "Final Streaming" payload carries only `streamId` and `streamType: "final"` in the entity — no `streamSequence`.

The service does not reject it: the final is accepted with `202 {}`, so nothing surfaces in logs. What we observed on a real tenant is that a final carrying `streamSequence` does not consistently terminate the stream in the client — the bubble keeps its Stop button while the message is delivered. We could not reproduce that deterministically enough to claim it as *the* symptom, and a stale web client muddied our own investigation, so the concrete, checkable part of this report is simply that the SDK sends a field the contract forbids there.

### Steps to Reproduce

```ts
app.on('message', async ({ stream }) => {
  stream.update('Thinking...');
  stream.emit('Hello');
  await stream.close();
});
```

Inspect the final activity sent by `close()`:

```json
"entities": [{ "type": "streaminfo", "streamId": "1785...", "streamType": "final", "streamSequence": 1 }]
```

### Expected Behavior

The final message omits `streamSequence` entirely, matching the documented payload:

```json
"entities": [{ "type": "streaminfo", "streamId": "1785...", "streamType": "final" }]
```

### Actual Behavior

`streamSequence` is always present on the final, defaulting to `1` when no chunks were counted.

### SDK Version

@microsoft/teams.apps 2.0.14, @microsoft/teams.api 2.0.14

### Node.js Version

v24.18.0

### Additional Context

We work around it by posting the final ourselves — same shape as `close()` builds (`id`, `channelData`, the `streaminfo` entity, `replyToId` since #646), minus `streamSequence` — and leaving the SDK to send the chunks. If `addStreamFinal()` drops the field we would delete that and go back to `close()`.

If the fix is taken, `MessageActivity.addStreamFinal()` already has the shape to copy, so the change is local to `MessageActivityInput`, and a test asserting the final's `streaminfo` entity has no `streamSequence` key would pin it.

maelga:
Fixes #687.

`MessageActivityInput.addStreamFinal()` puts `streamSequence: this.channelData.streamSequence ?? 1` on the final's `streaminfo` entity and on its `channelData`. Teams documents the sequence as required on every streaming request *except* this one:

> Here are the requirements for using `streamSequence` for REST APIs:
> - First one must be number '1'.
> - Subsequent numbers (except final) must be a monotonic increasing integer (for example, 1->2->3).
> - **For the final message, `streamSequence` must not be set.**

— [Stream agent messages](https://learn.microsoft.com/en-us/microsoftteams/platform/bots/streaming-ux#stream-message-through-rest-api)

`HttpStream.close()` builds its final with `MessageActivityInput`, so every final the streamer sends carries the field. `MessageActivity.addStreamFinal()` — the sibling implementation in the same file — already omits it, so this brings the two in line rather than inventing a shape.

The sequence is also deleted from `channelData`, so one carried over from the chunks doesn't reach the wire either.

### Tests
Two added in `message.spec.ts`: the final's entity and channelData have no `streamSequence`, and a sequence carried over from the chunks is dropped while `streamType` still becomes `final`.

- `packages/api` — 205 passed, 21 suites
- `packages/apps` — `src/http` 63 passed, `http-stream.spec.ts` included

### Note on impact
The service accepts a final carrying `streamSequence` with `202 {}`, so nothing surfaces in logs. On a real tenant we saw finals not consistently terminating the stream in the client — the bubble keeping its Stop button while the message was delivered — but we could not make that deterministic, and a stale web client muddied our own investigation. So this PR is offered as "the SDK sends a field the contract forbids there", not as a proven fix for that symptom.

### DIFF
--- a/packages/api/src/activities/message/message.spec.ts
+++ b/packages/api/src/activities/message/message.spec.ts
@@ -2,7 +2,7 @@ import { AdaptiveCard, TextBlock } from '@microsoft/teams.cards';
 
 import { Account, cardAttachment } from '../../models';
 
-import { MessageActivity } from './message';
+import { MessageActivity, MessageActivityInput } from './message';
 
 describe('MessageActivity', () => {
   const user: Account = {
@@ -467,6 +467,30 @@ describe('MessageActivity', () => {
         .addQuote('msg-1')
         .addText(' manual text');
       expect(activity.text).toEqual('<quoted messageId="msg-1"/> manual text');
+    });
+  });
+
+  describe('addStreamFinal', () => {
+    // Teams documents `streamSequence` as required on every streaming request except this one:
+    // "For the final message, `streamSequence` must not be set."
+    it('should not put a streamSequence on the final', () => {
+      const activity = new MessageActivityInput('done').withId('stream-1').addStreamFinal();
+
+      const streamInfo = (activity.entities ?? []).find((e) => e.type === 'streaminfo');
+      expect(streamInfo).toMatchObject({ streamId: 'stream-1', streamType: 'final' });
+      expect(streamInfo).not.toHaveProperty('streamSequence');
+      expect(activity.channelData).not.toHaveProperty('streamSequence');
+    });
+
+    it('should drop a streamSequence carried over from the chunks', () => {
+      const activity = new MessageActivityInput('done')
+        .withChannelData({ streamId: 'stream-1', streamType: 'streaming', streamSequence: 4 })
+        .addStreamFinal();
+
+      const streamInfo = (activity.entities ?? []).find((e) => e.type === 'streaminfo');
+      expect(streamInfo).not.toHaveProperty('streamSequence');
+      expect(activity.channelData).not.toHaveProperty('streamSequence');
+      expect(activity.channelData?.streamType).toEqual('final');
     });
   });
 });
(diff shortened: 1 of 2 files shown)


maelga:

heyitsaamir (approved):
Makes sense!

corinagum (approved):

corinagum:
Thank you @maelga!

PR #704, opened 2026-08-05
fix: use canonical npm registry URLs in package-lock

corinagum:
## Summary

- replace Azure Artifacts mirror URLs in `package-lock.json` with canonical `registry.npmjs.org` URLs
- preserve the exact dependency graph, versions, and integrity hashes
- add a CI guard that rejects future lockfile entries resolved from non-npmjs registry hosts

The repository does not set a registry. External contributors can use npmjs, internal developers can use their configured proxy, and Azure pipelines can continue configuring and authenticating their internal registry. npm's default `replace-registry-host=npmjs` behavior redirects canonical npmjs lock URLs through the active registry.

## Local validation

- clean npm 11.12.1 install from npmjs with an empty cache
- clean npm 12 install from npmjs with an empty cache
- clean npm 12 install through `https://packagefeedproxy.microsoft.io/npm/` with an empty cache
- clean npm 12 install using a temporary public `--registry` override while the saved registry remained internal
- simulated internal registry routing confirmed npm redirected the canonical tarball URL through the configured registry path
- `npm run check:lockfile-registry`
- `npm run lint`
- `npm run build`
- `npm run test`

## Remaining validation

The internal Azure build pipeline should run its authenticated `npm ci` stage before merge to verify feed authentication, mirror completeness, and 1ES network policy.

### DIFF
--- a/.github/workflows/build-test-lint.yml
+++ b/.github/workflows/build-test-lint.yml
@@ -25,6 +25,8 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
+      - name: Check package-lock registry
+        run: npm run check:lockfile-registry
       - name: Install Dependencies
         run: npm ci
       - name: Check unused dependencies
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+omit-lockfile-registry-resolved=true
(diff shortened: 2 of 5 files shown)


singhk97 (approved):
lgtm

PR #712, opened 2026-08-06
Feature: Receive files in personal scope

corinagum:
Adds API for receiving files to the app in personal scope. A user uploads a file, which arrives on the activity, and `ctx.files.list()` provides metadata of the file(s) and lazy handlers to download it. The url is `tempauth` and will expire.

- **`ctx.files`**: `list()` / `first()` map inbound `file.download.info` attachments to `IIncomingFile`. Never throws: non-message activities and empty/malformed attachments yield an empty list; bad entries are skipped.
- **`IIncomingFile`**: file metadata with lazy `download()`, `stream()`, `text(encoding?)`, `arrayBuffer()`, `saveAs(path)`. Each read re-fetches (bounded by the short-lived URL).
- **`IDownloadedFile`**: buffered snapshot you own (`bytes`, `contentType`, `filename`, `sourceUrl`), with synchronous `text()`/`arrayBuffer()`/`saveAs()`.
- **Errors**: Differentiates between first fetch error (url expired before first fetch), reread error (url expired after first fetch), and scope not supported error (today, only 1:1).
- **Types**: new `FileDownloadInfo` model + `file.download.info` content-type constant; `ConversationType` extended with `channel` (forward-compatible).

## Testing

- Unit tests: `files-accessor.spec.ts`, `incoming-file.spec.ts`
- Manual e2e

## Note

- After investigating thumbnail url behavior, I adjusted `Attachments.thumbnailUrl` docstrings to indicate it will not be populated on incoming files.
- Similarly for `etag`, this field may be set by Graph, which is not in scope for this PR.

### DIFF
--- a/packages/apps/src/files/download.ts
+++ b/packages/apps/src/files/download.ts
@@ -59,6 +59,10 @@ async function openPersonalFileStream(
     throw new Error('cannot download personal file: no download URL is available');
   }
 
+  if (!/^https:\/\//i.test(url)) {
+    throw new Error('cannot download file: download URL must use https');
+  }
+
   const doFetch = options?.fetch ?? defaultFetch;
 
   // Plain GET with no bearer token: the download URL embeds its own `tempauth` credential, and attaching a credential can get the request rejected.
--- a/packages/apps/src/files/download.ts
+++ b/packages/apps/src/files/download.ts
@@ -1,0 +1,113 @@
+import { ConversationType } from '@microsoft/teams.api';
+
+import { FileScopeNotSupportedError, FileUrlExpiredError } from './errors';
+
+/**
+ * Pluggable fetch used to retrieve file bytes. Defaults to the global `fetch`; injectable so tests can supply a real `Response` without hitting the network.
+ */
+export type FileFetch = (url: string, init?: { signal?: AbortSignal }) => Promise<Response>;
+
+/**
+ * The minimal file description the download dispatcher needs to open a byte stream.
+ */
+export type FileFetchTarget = {
+  /** Conversation scope; the dispatcher is keyed on this. */
+  scope: ConversationType;
+  /** Short-lived, pre-authorized download URL (personal scope). */
+  downloadUrl?: string;
+  /** MIME type reported by the incoming file, used as a fallback when the response omits one. */
+  contentType?: string;
+};
+
+/**
+ * A freshly opened, single-consumption byte stream plus the metadata resolved while opening it.
+ */
+export type OpenedFileStream = {
+  /** The raw response body stream. Uncapped; the caller bounds it. */
+  stream: ReadableStream<Uint8Array>;
+  /** The URL the bytes were actually fetched from. */
+  sourceUrl: string;
+  /** MIME type resolved from the response, falling back to the incoming file's. */
+  contentType: string;
+};
+
+const defaultFetch: FileFetch = (url, init) => fetch(url, { method: 'GET', signal: init?.signal });
+
+/**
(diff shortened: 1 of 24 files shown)


packages/apps/src/files/download.ts
```diff
@@ -0,0 +1,117 @@
...
+export type OpenedFileStream = {
+  /** The raw response body stream. Uncapped; the caller bounds it. */
+  stream: ReadableStream<Uint8Array>;
+  /** The URL the bytes were actually fetched from. */
+  sourceUrl: string;
+  /** MIME type resolved from the response, falling back to the incoming file's. */
+  contentType: string;
+};
+
+const defaultFetch: FileFetch = (url, init) => fetch(url, { method: 'GET', signal: init?.signal });
```
heyitsaamir: we should be using http-client for this, right?

packages/apps/src/files/download.ts
```diff
@@ -0,0 +1,117 @@
...
+): Promise<OpenedFileStream> {
+  const url = target.downloadUrl;
+
+  if (!url) {
+    throw new Error('cannot download personal file: no download URL is available');
+  }
+
+  if (!/^https:\/\//i.test(url)) {
+    throw new Error('cannot download file: download URL must use https');
+  }
```
heyitsaamir, line 90: NIT: should we be making this assertion?

heyitsaamir (commented):
Looks good! Can you include a sample as well?

packages/apps/src/files/download.ts
corinagum: Honestly I added it in response to Copilot's SSRF flag. It doesn't buy us SSRF protection and we're explicitly trusting the service to do JWT-authentication for us, but the AI security reviews will continue to flag these. This shows as intentional as opposed to a missed gap.

packages/apps/src/files/download.ts
corinagum: Done. Downloads now go through `HttpClient`. One issue: we must not send `Authorization`, since the download URL carries its own short-lived `tempauth` and a bearer token on top of that gets rejected. I'm using the per-request override which `withConfig` resolves. First I tried `clone({ token: undefined})`, but a refactor to `options.token ?? this.token` would have silently re-attached auth. The per-request way avoids building a client per download. Added a test on `client.spec.ts`. 

Downloads now inherit the client's `timeout`, where global `fetch` had none. TS now diverges from PY/C#, which use plain transports. Did we have a plan to update that in the future?

corinagum:
@heyitsaamir yep as discussed on Friday, samples are in progress and will be in a PR immediately following this one. :)

packages/apps/src/files/download.ts
```diff
@@ -0,0 +1,195 @@
...
+async function requestViaHttpClient(
+  url: string,
+  client: HttpClient,
+  signal?: AbortSignal
+): Promise<TransportResponse> {
+  const response = await client.get<Readable>(url, {
+    responseType: 'stream',
+    signal,
+    // The URL carries its own `tempauth` credential; a bearer token on top of it can get the request rejected.
+    token: () => undefined,
```
lilyydu, line 131: this is an edge case but we may need to also remove the default `Authorization` header from the shared client. `Client.withConfig()` merges `options.headers` into the request so file downloads can still send auth to the pre-authorized URL

should we explicitly strip/override `Authorization` here?

corinagum: Thanks! Fixed

lilyydu (approved):

PR #716, opened 2026-08-08
feat(apps): inbound Socket Mode transport

singhk97:
## Socket Mode

Adds an experimental inbound Socket Mode transport using Teams-negotiated SignalR WebSockets.

- Socket Mode is exclusive: it starts no HTTP listener, routes, fallback, or public endpoint.
- Activities use the shared App pipeline with an internally synthesized connection-authenticated token.
- Supports independent multi-geo connections, startup retries, reconnects, readiness fencing, and proactive token refresh.
- Invoke responses and activity acknowledgements return over SignalR.
- Browser-based HTTP features are unavailable.
- Supports regular production clouds only; Canary may be selected through `negotiateBaseUrl`.

WebSocket is only recommended for use when developing agents.

NOTE:
We open 3 geo-specific sockets (AMER, APAC, and EMEA) at startup. With the regular client, Teams initiates the request and provides the correct region through the activity or token `serviceUrl`. In socket mode, the bot must establish connections before any activity arrives, so it cannot know which region the backend service will route traffic through. Opening all three sockets ensures a regional connection is already available and keeps EU traffic within its required data boundary.

PR #720, opened 2026-08-11
refactor(apps): introduce transport seam at the IHttpServerAdapter layer

singhk97:
## Summary
Introduces a transport-agnostic inbound seam at `IHttpServerAdapter` so non-HTTP transports (e.g. Socket Mode) can live **inside a single `HttpServer`**, rather than as a sibling of a parent `IServer` abstraction with a public multi-server type.

This is the **base layer of a stack**; the upper layer (#716) adds an inbound Socket Mode transport as an `IHttpServerAdapter`.

## Changes
- **`IHttpServerRequest.token` (optional)** — transports that authenticate at the connection level (Socket Mode authenticates once at negotiate and carries no per-frame JWT) deliver a pre-resolved identity. `HttpServer.handleRequest` trusts it and skips per-request JWT validation. HTTP adapters never set it, so **HTTP behavior is byte-for-byte unchanged**.
- **`IHttpServerAdapter.initialize(deps)` (optional)** — `HttpServer.initialize` forwards app-level deps (`credentials`/`cloud`) to the adapter, so connection-authenticated transports initialize purely through the seam.
- **`IHttpServerAdapter` documented as the transport seam** — a non-HTTP transport implements it and may no-op `serveStatic` and non-messaging `registerRoute` paths.
- Minor tidy-ups carried along: `PluginDependencyOptions.name` typing (`Omit<string, …>` can't subtract string literals, so it read as enforcing something it doesn't) and `APX` → `Teams backend service` in `thread.ts` comments.

## Compatibility
`App.server` remains the concrete `HttpServer` — no `IServer` type, no `CompositeServer`, no transport type-guards. Pure refactor: the apps test suite is green, including red-green guards for the pre-authenticated-token path and adapter-init forwarding.


packages/apps/src/types/plugin/decorators/dependency.ts
```diff
@@ -77,6 +77,13 @@ export function HttpServer() {
   return Dependency({ name: 'IHttpServer' });
 }
 
+/**
+ * inject the `App` `IServer` instance (the active inbound transport)
+ */
+export function Server() {
+  return Dependency({ name: 'IServer' });
+}
```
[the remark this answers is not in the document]
corinagum: Confirming this one, it's real. Add `Server` to the barrel export list in `decorators/index.ts`, which currently exports only `Dependency`, `DependencyOptions`, `Logger`, `Storage`, and `HttpServer`.

`Server()` is defined here and `App` does register `'IServer'` in the container, but the decorator never leaves the module. Nobody outside the package can use it today.

packages/apps/src/types/plugin/decorators/dependency.ts
```diff
@@ -127,13 +135,26 @@ export type HttpServerDependencyOptions = {
...
+   * the name used to resolve the dependency
+   */
+  readonly name: 'IServer';
+
+  /**
+   * if optional, the app will not throw
+   * if the dependency is not found
+   */
+  readonly optional?: false;
+};
```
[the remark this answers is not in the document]
corinagum: Agreed. Either make `optional` a real `boolean` here or drop the field and its doc comment.

As written, `@Dependency({ name: 'IServer', optional: true })` is a type error, so the documented behavior can't be reached. `IServer` in particular seems like one you'd want optional, since a custom runtime may not have one.

packages/apps/src/types/plugin/decorators/dependency.ts
```diff
@@ -127,13 +135,26 @@ export type HttpServerDependencyOptions = {
...
+
 export type PluginDependencyOptions = {
   /**
    * the name used to resolve the dependency
    */
   readonly name?: Omit<
     string,
-    'ILogger' | 'IStorage' | 'IHttpServer'
+    'ILogger' | 'IStorage' | 'IHttpServer' | 'IServer'
   >;
```
corinagum: Not blocking, and pre-existing, but `Omit<string, ...>` doesn't do what it reads like.

`Omit` on a primitive doesn't exclude string literals, so `PluginDependencyOptions` still accepts `'IServer'` and the union stays ambiguous. Since you're touching this line to add `'IServer'`, worth a note or a follow-up issue so the next person doesn't assume it's enforced.

packages/apps/src/types/plugin/decorators/dependency.ts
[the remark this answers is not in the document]
singhk97: Resolved by removing `Server()` entirely (commit 797e08bf) rather than exporting it. The plugin DI model is being superseded by the transport abstraction this PR introduces, and its sibling `HttpServer()` is already `@deprecated` and unused by any plugin — so adding a new experimental plugin decorator sent the wrong signal. The `App` still registers `'IServer'` in the container for internal resolution.

packages/apps/src/types/plugin/decorators/dependency.ts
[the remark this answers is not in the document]
singhk97: Resolved by removing `ServerDependencyOptions` along with the `Server()` decorator (commit 797e08bf), so the unreachable `optional: false` typing is gone. If a transport-injection decorator is reintroduced later it should type `optional` as a real `boolean`.

packages/apps/src/types/plugin/decorators/dependency.ts
singhk97: Good catch. Replaced the misleading `Omit<string, ...>` with plain `string` plus a comment explaining that the reserved DI names (`ILogger`, `IStorage`, `IHttpServer`, `IServer`) aren't excluded at the type level — TS can't subtract string literals from `string` — so reusing one resolves the built-in instead of a plugin (commit 797e08bf). Removes the false "it's enforced" signal without pretending to fix the underlying limitation.

corinagum (approved):

lilyydu (approved):

packages/apps/src/app.ts
```diff
@@ -269,7 +274,7 @@ export class App<TPlugin extends IPlugin = IPlugin> {
   readonly cloud: CloudEnvironment;
   readonly graph: GraphClient;
   readonly log: ILogger;
-  readonly server: HttpServer;
+  readonly server: IServer;
```
puneetgulia-rubberducky: Changing this public property from `HttpServer` to `IServer` is source-breaking for existing consumers: `adapter`, `registerRoute`, `serveStatic`, `handleRequest`, and `messagingEndpoint` disappear from the inferred type even though every construction path in this PR still creates `HttpServer`. The casts added to this PR's own tests demonstrate the downstream compilation break. Please preserve an HTTP-typed compatibility path or treat and document this as a breaking release.

packages/apps/src/http/http-server.ts
```diff
@@ -40,8 +41,12 @@ export type HttpServerOptions = {
 /**
  * Public interface for HttpServer, exposed via DI for plugins
  */
-export interface IHttpServer {
+export interface IHttpServer extends IServer {
```
puneetgulia-rubberducky: Making the existing public `IHttpServer` extend `IServer` breaks previously valid structural implementations and test doubles. They now need the transport tag, lifecycle methods, `onRequest`, `registerRoute`, and `serveStatic`. Marking the newly introduced `IServer` experimental does not make changes to the existing stable interface non-breaking. Please preserve the old contract or provide an explicitly versioned migration path and compile-contract tests.

packages/apps/src/server/server.ts
```diff
@@ -0,0 +1,40 @@
...
+  readonly transport: string;
+
+  /** Callback the App sets to process an inbound activity and return its response. */
+  onRequest?: (event: IActivityEvent) => Promise<InvokeResponse>;
+
+  /** Initialize the server with app-level dependencies. Called by App.initialize(). */
+  initialize(deps: IServerInitializeDeps): Promise<void>;
+
+  /** Start the server. Called by App.start(). `port` is transport-specific and may be ignored. */
+  start(port?: number | string): Promise<void>;
```
puneetgulia-rubberducky: This interface makes `server.start()` without arguments type-correct, but the shipped `HttpServer.start()` immediately throws when `port` is undefined. That converts a previous compile-time error into a deterministic runtime failure. The shared interface should not accept an invocation that its default implementation rejects.

packages/apps/src/app.ts
```diff
@@ -529,6 +534,7 @@ export class App<TPlugin extends IPlugin = IPlugin> {
 
     // Register HTTP server for plugins that need HTTP capabilities
     this.container.register('IHttpServer', { useValue: server });
+    this.container.register('IServer', { useValue: server });
```
puneetgulia-rubberducky: This newly reserves the previously legal plugin metadata/class name `IServer`. `PluginManager.add()` pushes a plugin before container registration, so a dynamic installation collision throws while leaving the supposedly rejected plugin registered and eligible for `onInit()`. Validate reserved names before mutation or namespace the token, and make plugin registration transactional.

packages/apps/src/app.ts
```diff
@@ -62,6 +63,10 @@ function isAppSendOptions(value: ActivityLike | DeprecatedInputActivity | AppSen
   return !('type' in value);
 }
 
+function isHttpServer(server: IServer): server is IHttpServer {
+  return server.transport === 'http';
```
puneetgulia-rubberducky: The public `IServer.transport` is typed as plain `string`, so consumers cannot use this tag to narrow `IServer` to `IHttpServer`; this working predicate is private. It is also tag-only, so an `IServer` claiming `transport: 'http'` without HTTP capabilities passes and later fails with a `TypeError`. Please expose a sound guard or model the server surface as a closed discriminated/capability union.

puneetgulia-rubberducky (changes requested):
This looks source-breaking rather than behavior-preserving. `@microsoft/teams.apps` is already a public stable 2.x package with substantial download volume, so we should assume these APIs have external consumers even though private customer usage is not publicly searchable.

**Blocking concerns:**

1. Changing `App.server` from `HttpServer` to `IServer` removes HTTP capabilities from the public TypeScript surface even though every construction path in this PR still creates `HttpServer`.
2. Extending the existing public `IHttpServer` with the new `IServer` contract breaks previously valid structural implementations and mocks.
3. `IServer.start(port?)` accepts a no-argument call that the default `HttpServer` deterministically rejects.
4. The new string DI token `IServer` creates a plugin-name/class-alias collision, and dynamic registration can leave the rejected plugin active because mutation occurs before container registration throws.
5. `transport: string` does not form a usable public discriminated union, while the private guard trusts the tag without checking HTTP capabilities.
6. There are no behavior or compile-contract tests for the new non-HTTP guards, no-port startup, `IServer` resolution, token collisions, or legacy/custom `IHttpServer` compatibility.

There are also two pre-existing lifecycle issues that should be fixed before landing the 720 -> 716 -> 743 stack: `App.start()` swallows transport startup failures, contradicting PR 716's stated startup contract, and a rejecting plugin `onStop()` prevents `server.stop()` from running.

Finally, this branch currently conflicts with `main`, including substantial concurrent changes to `app.ts`. Please rebase and rerun the full build, tests, and lint after resolving those conflicts.

packages/apps/src/http/http-server.ts
```diff
@@ -148,13 +153,16 @@ export class HttpServer implements IHttpServer {
    * Start the HTTP server
    * Called by App.start()
    */
-  async start(port: number | string) {
+  async start(port?: number | string) {
```
premkaretimsft: This contract makes it look like 'port' is optional whereas we're throwing immediately if it's not set so, this should be a required one based on your implementation

packages/apps/src/app.ts
lilyydu: if we register both `IHttpServer` and `IServer` - does that mean we could end up with 2 HTTP servers?

packages/apps/src/app.ts
singhk97: Addressed by the rework: `App.server` stays typed as `HttpServer` — the `IServer` change was reverted. The transport seam moved down to `IHttpServerAdapter` (`new App({ httpServerAdapter })`), so `adapter`, `handleRequest`, and `messagingEndpoint` remain on the public type and there's no source break for existing consumers. Resolving as outdated.

packages/apps/src/http/http-server.ts
singhk97: Addressed: `IHttpServer` is a standalone interface again (http-server.ts:43) and no longer extends any `IServer`; its members are unchanged. Existing structural implementations and test doubles keep the old contract — no transport tag, lifecycle, `onRequest`, `registerRoute`, or `serveStatic` are required of them. Resolving as outdated.

packages/apps/src/server/server.ts
singhk97: Addressed: there is no `server/server.ts` / `IServer` anymore. The only `start` is on the adapter — `start?(port: number | string)` with `port` required — so the interface no longer accepts a no-arg invocation that the implementation rejects at runtime. Resolving as outdated.

packages/apps/src/app.ts
singhk97: Addressed: no plugin-name reservation of `IServer` remains. The only DI tokens registered are the built-ins (`ILogger`/`IStorage`/`IHttpServer`) via `useValue`, done once outside the plugin loop — there's no `PluginManager.add()` ordering hazard around a reserved server token. Resolving as outdated.

packages/apps/src/app.ts
singhk97: Addressed: the `transport` tag and the `IServer` union were dropped entirely. A single `HttpServer` now wraps pluggable `IHttpServerAdapter` transports, so there's no `string`-typed discriminant to narrow and no way for an `IServer` to claim `transport: 'http'` without HTTP capabilities. Resolving as outdated.

packages/apps/src/http/http-server.ts
singhk97: Addressed: `port` is now required on the adapter's `start?(port: number | string)`, matching the implementation that throws when it's missing. Resolving as outdated.

packages/apps/src/app.ts
singhk97: Good question — addressed by the rework: we no longer register both. Only `IHttpServer` is registered (once, via `useValue`), and a single `HttpServer` instance wraps the chosen `IHttpServerAdapter`, so there's no path to two HTTP servers. Resolving as outdated.

corinagum (approved):

teddyam (approved):

PR #726, opened 2026-08-13
feat: add Anthropic support to AI MCP sample

umangsehgal:
## Summary
- add `AI_PROVIDER` selection for Azure OpenAI and Anthropic Claude
- implement Anthropic streaming and client tool execution while preserving the Teams handler boundary
- share MCP execution, citations, clarification cards, prompts, and follow-up generation across providers
- document both provider configurations

## Validation
- TypeScript compilation and ESLint
- local server startup and Microsoft Learn MCP discovery
- mocked Anthropic tool loop
- live Claude request using Microsoft Learn MCP, streaming text, citations, and follow-up prompts

## Related PRs
- C#: https://github.com/microsoft/teams.net/pull/650
- Python: https://github.com/microsoft/teams.py/pull/559
- Documentation: https://github.com/microsoft/teams-sdk/pull/2957

### DIFF
--- a/examples/ai-mcp/README.md
+++ b/examples/ai-mcp/README.md
@@ -1,111 +1,122 @@
-# ai-mcp — Azure OpenAI + MCP sample
+# ai-mcp — multi-provider AI + MCP sample
 
-A Teams bot powered by **Azure OpenAI** (via the `openai` SDK's `AzureOpenAI` client) and the **Model Context Protocol** SDK. Demonstrates streaming responses, per-conversation memory, a local clarification-card tool, remote MCP server tools, inline citations, follow-up suggestions, and custom feedback.
+A Teams bot powered by either **Azure OpenAI** or **Anthropic Claude**, plus the **Model Context Protocol** SDK. It demonstrates streaming responses, per-conversation memory, a local clarification-card tool, remote MCP tools, inline citations, follow-up suggestions, and custom feedback.
 
-This is the TypeScript counterpart to the .NET [`ExtAIBot`](https://github.com/microsoft/teams.net/pull/486) sample. The structural shape matches: a single bot process owns the AI plumbing, runs the tool-call loop in-process, and connects to MCP servers directly. The auto-loop comes from `openai.chat.completions.runTools()` — the OpenAI SDK's helper that auto-executes each tool's `function` callback and feeds the result back to the model until it produces final text.
-
-> **Provider scope.** This sample is bound to the OpenAI chat-completions wire protocol — Azure OpenAI works; vanilla OpenAI works; non-OpenAI providers do not. (See the .NET sample for an `IChatClient` abstraction that's provider-agnostic — TS has no equivalent today.)
+The Teams layer is provider-neutral: activity handlers call `agent.run(conversationId, text, stream)`. Each provider implementation owns its message history and tool loop.
 
 ## Features
 
-- **Streaming** — `runner.on('content', delta => stream.emit(delta))` forwards text token-by-token
-- **Conversation memory** — each conversation keeps its own `Map<conv, ChatCompletionMessageParam[]>` in process
-- **Local tool** — `request_clarification`: a `RunnableToolFunction` whose `function` callback pushes an Adaptive Card into a per-turn bucket and returns a placeholder string. The agent discards its wrap-up text and sends only the card.
-- **MCP client** — connects to the [Microsoft Learn docs MCP server](https://learn.microsoft.com/api/mcp) at startup. Each MCP tool is wrapped as a `RunnableToolFunction` whose `function` callback invokes the server and feeds the raw result into the citation collector before returning it to the model.
-- **Inline citations** — `CitationCollector` parses MCP tool results for `{ contentUrl, title, content }` records; `[N]` markers in the final reply become clickable Teams citation entities.
-- **Follow-up suggestions** — a separate non-streaming `chat.completions.create({ response_format: { type: 'json_schema', ... }})` call produces two short prompts shown as suggested-action chips.
-- **Custom feedback** — every text reply enables `addFeedback('custom')`; clicking thumbs up/down opens a bot-rendered task module, and submissions hit `message.submit.feedback`.
+- **Provider selection** — set `AI_PROVIDER=azure-openai` or `AI_PROVIDER=anthropic`.
+- **Streaming** — forwards model text into the Teams `IStreamer`.
+- **Conversation memory** — keeps provider-native history per Teams conversation.
+- **Local tool** — shows an Adaptive Card when the request needs clarification.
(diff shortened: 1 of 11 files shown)


umangsehgal:
### Deletion audit

I reviewed every removed line against `main`.

- No files are deleted.
- The existing Azure OpenAI `Agent`, `runTools()` loop, history, streaming, citations, cards, feedback, and follow-up path remain.
- Most removals are code moves: prompts moved to `prompts.ts`; clarification execution and MCP execution were extracted so Azure OpenAI and Anthropic share them; handlers changed only from concrete `Agent` to `IAgentRunner`.
- The PR adds 543 lines and removes 137; 62 removed lines are README rewriting, while the code removals are replaced by shared helpers or provider-neutral types.

Validation covers compilation, ESLint, the existing Azure path structurally, mocked Anthropic tools, and a live Claude + MCP run.

examples/ai-mcp/appPackage/manifest.json
```diff
@@ -15,7 +15,7 @@
     "termsOfUseUrl": "https://www.microsoft.com/legal/terms-of-use"
   },
   "description": {
-    "short": "Teams bot powered by Azure OpenAI + MS Learn MCP",
+    "short": "Teams bot powered by AI + MS Learn MCP",
```
rajan-chari, line 18: nit: AI is broad.  Consider narrowing it to LLM.  For example, this won't run with Supervised learning or  Boosted Trees or Reinforcement Learning ... etc.

examples/ai-mcp/src/anthropic-agent.ts
```diff
@@ -0,0 +1,246 @@
...
+    return history;
+  }
+
+  private async _generateFollowUps(history: Anthropic.MessageParam[]): Promise<string[]> {
+    try {
+      const response = await this._client.messages.create({
+        model: this._model,
+        max_tokens: 200,
+        system: `${SYSTEM_PROMPT}\n\nReturn valid JSON only with string properties "prompt1" and "prompt2".`,
+        messages: [...history, { role: 'user', content: FOLLOW_UPS_PROMPT }],
```
corinagum, line 199: This call omits `tools`, but `history` contains `tool_use` and `tool_result` blocks as soon as any tool has run, which in this sample is essentially every turn. Anthropic rejects that combination:
```
400 invalid_request_error: Requests which include `tool_use` or `tool_result` blocks must define tools.
```
The catch on line 208 swallows it and returns `[]`, so follow-up chips silently never render on the Anthropic path. It degrades quietly, which is why it isn't obvious in manual testing: the only symptom is a `Follow-up generation failed` warning in the bot console.

Fix is either passing `tools: this._buildTools()` here, or stripping tool blocks from the history copy sent for follow-ups. The first is less code; the second saves input tokens.

examples/ai-mcp/src/anthropic-agent.ts
```diff
@@ -0,0 +1,246 @@
...
+      const modelStream = this._client.messages
+        .stream({
+          model: this._model,
+          max_tokens: this._maxTokens,
+          system: SYSTEM_PROMPT,
+          messages: history,
+          tools,
+        })
+        .on('text', (text) => {
+          fullText += text;
```
corinagum, line 109: `fullText` accumulates text from every round, including the text Claude emits before a `tool_use` block. Claude narrates before tool calls far more consistently than GPT does, so the final Teams message routinely leads with the narration:

`"Let me search Microsoft Learn for that. You stream responses in teams.ts by..."`

The Azure OpenAI path has the same structure. Resetting `fullText = ''` when a round ends in `tool_use` would keep only the final round's prose, which is what the user should actually see.

examples/ai-mcp/src/anthropic-agent.ts
```diff
@@ -0,0 +1,246 @@
...
+        return fullText;
+      }
+
+      const toolResults = await Promise.all(
+        toolUses.map((toolUse) => this._executeTool(toolUse, pendingCards, citations))
+      );
+      history.push({ role: 'user', content: toolResults });
+    }
+
+    throw new Error(`Anthropic tool loop exceeded ${MAX_TOOL_ROUNDS} rounds.`);
```
corinagum, line 130: This throw propagates out of `agent.run` into `app.on('message')` in handlers.ts, where nothing catches it. The stream is never closed, so the user sits on "Thinking..." indefinitely with no error.

examples/ai-mcp/src/anthropic-agent.ts
```diff
@@ -0,0 +1,246 @@
...
+    return turn;
+  }
+
+  private async _runTurn(
+    teamsConvId: string,
+    userText: string,
+    stream: IStreamer
+  ): Promise<AgentRunResult> {
+    const history = this._getOrCreateHistory(teamsConvId);
+    history.push({ role: 'user', content: userText });
```
corinagum, line 75: The user message is committed to the stored history before the request is made, so a failed turn leaves it orphaned there permanently. I confirmed the next turn still succeeds (injected a 529 and the following turn recovered), so this is context pollution rather than breakage.

Building the turn against a local copy and committing back to `this._histories` only on success would avoid it.

examples/ai-mcp/src/local-tools.ts
```diff
@@ -1,3 +1,4 @@
+import type { JSONSchema } from 'openai/lib/jsonschema';
```
corinagum, line 1: This types the shared clarification schema with OpenAI's `JSONSchema`, and anthropic-agent.ts then consumes it via `CLARIFICATION_TOOL_SCHEMA`. It's type-only so there's no runtime cost, but it means the provider-neutral module is expressed in the OpenAI package's types. A plain inline type would keep the shared surface neutral.

examples/ai-mcp/README.md
```diff
@@ -1,111 +1,122 @@
...
+src/anthropic-agent.ts    Anthropic streaming and tool-use loop
+src/prompts.ts            Shared system and follow-up prompts
+src/local-tools.ts        Provider-neutral clarification behavior plus OpenAI adapter
+src/mcp-tools.ts          Provider-neutral MCP execution plus OpenAI adapter
+src/citation-collector.ts MCP citation parsing and Teams citation attachment
+src/handlers.ts           Provider-neutral Teams activity handlers
 ```
 
-### Tool loop
+## Anthropic tool loop
```
corinagum, line 101: The rewrite drops the tool-loop walkthrough, the clarification-flow steps, and the "how the pieces fit" section, and replaces them with Anthropic-specific equivalents. Azure OpenAI is still the default provider, so the default path loses its explanation while the opt-in path gains one.

Could the Azure OpenAI walkthrough come back alongside the Anthropic one? The provider-neutral framing reads well otherwise.

PR #729, opened 2026-08-13
Add per-turn conversation and user state

MehakBindra:
## Summary

- add dirty-tracked, sealable per-turn conversation and user state scopes
- expose state through `AppOptions.state` and `ctx.state`, loading before dispatch and saving in `finally`
- keep keyed reads clean and require `set()` for persisted mutations
- persist each scope as a raw-dictionary JSON string through an explicitly typed `IStorage<string, string>`
- give `state: true` dedicated process-local string storage instead of casting or reusing legacy `AppOptions.storage`; production apps can provide `state.storage`
- percent-encode conversation and user ID key segments to avoid delimiter collisions
- treat conversation and user scopes as independent persistence units: each dirty scope is serialized and written separately, successful scopes become clean, and no cross-scope atomicity is promised
- keep `IStorage.set` as a minimal key/value contract; expiration and other policies remain owned by the configured provider
- treat malformed persisted values as absent and emit a warning without logging state contents
- route state load/save failures through the turn error pipeline and return 500; when dispatch and save both fail, report dispatch first and save separately so neither masks the other
- preserve state mutations on handler failures and always seal loaded state
- emit the normalized `InvokeResponse` through the public `activity.response` event, matching its declared contract
- deprecate direct `AppOptions.storage`, `app.storage`, and `ctx.storage` access in favor of `state.storage` or application-owned persistence
- deprecate, but retain, `IListStorage` and `ListLocalStorage`; Python still exposes list storage, while the .NET state/storage design does not include this generic abstraction
- add a TypeScript state example modeled on the .NET state bot

Ports the combined state foundation from microsoft/teams.py#553 and microsoft/teams.py#554, aligned with microsoft/teams.net while keeping TypeScript keyed state explicit and JSON-native. Standardized TTL and other per-write options are intentionally deferred until the SDK has a concrete cross-provider requirement.

## Validation

- focused state and activity-processing tests pass
- changed-file lint passes
- apps package build passes


packages/apps/src/app.process.ts
```diff
@@ -313,6 +327,18 @@ export class ActivityProcessor<TPlugin extends IPlugin = IPlugin> {
...
         });
+      } finally {
+        if (context.state && this.options.stateLoader && conversationId) {
+          try {
+            await this.options.stateLoader.save(
+              context.state,
+              conversationId,
+              userId
+            );
+          } finally {
```
corinagum: `catch`?

packages/apps/src/state/loader.ts
```diff
@@ -0,0 +1,177 @@
...
+    return `${this.keyPrefix}:conv:${conversationId}`;
+  }
+
+  /**
+   * Returns the persisted user-within-conversation key.
+   * @param conversationId Conversation ID.
+   * @param userId User ID.
+   */
+  userKey(conversationId: string, userId: string): string {
+    return `${this.keyPrefix}:user:${conversationId}:${userId}`;
```
corinagum: These should be escaped

packages/apps/src/app.process.ts
lilyydu: ^ we should do for load & save - load failures bypass onError right now

packages/apps/src/state/loader.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  /**
+   * Persists dirty scopes from a loaded container.
+   *
+   * Dirty empty scopes delete their backing keys; clean scopes perform no I/O.
+   * @param container State container to persist.
+   * @param conversationId Conversation ID associated with the state.
+   * @param userId Optional sender ID associated with the user scope.
+   */
+  async save(
```
lilyydu: Corina left this for PY but conversation state is serialized/written before user state. if we encounter an invalid user state, itll leave the conversation state partially persisted. We should pepare both serialized values first, then perform ops

packages/apps/src/state/loader.ts
```diff
@@ -0,0 +1,177 @@
...
+  fallbackStorage: IStorage<string, unknown>,
+  logger: ILogger
+): TurnStateLoader | undefined {
+  if (!state) {
+    return undefined;
+  }
+
+  const options = state === true ? {} : state;
+  const storage = options.storage ??
+    fallbackStorage as IStorage<string, string>;
```
lilyydu: should we support providers returning decoded objects too?

packages/apps/src/state/loader.ts
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  private async loadScope(key: string): Promise<TurnState> {
+    const value = await this.storage.get(key);
+    if (!value) {
+      return new TurnState();
+    }
+
+    const data = deserializeState(value);
+    if (!data) {
```
lilyydu, line 146: I think we should delete malformed vals here - otherwise parsing fails every turn

packages/apps/src/state/loader.ts
```diff
@@ -0,0 +1,177 @@
...
+      return new TurnState();
+    }
+    return new TurnState(data);
+  }
+
+  private async saveScope(key: string, state: TurnState): Promise<void> {
```
lilyydu, line 155: should successful delete/set paths call `state.markClean()`? Otherwise the exported `TurnStateLoader.save()` leaves the scope dirty and a second save repeats the storage op

packages/apps/src/state/loader.ts
MehakBindra: state’s contract is JSON strings, and accepting objects would weaken type guarantees and create inconsistent provider behavior

packages/apps/src/state/loader.ts
MehakBindra: Conversation and user state are independent scopes with separate keys and dirty tracking, so successfully persisting conversation state despite invalid user state seems fine

packages/apps/src/state/loader.ts
MehakBindra: Loading should be read-only; automatic deletion can fail the turn, erase provider data unexpectedly, or race with a concurrent correction. cleanup should be an explicit operation

lilyydu: I think we should log at least if its malformed data

packages/apps/src/state/loader.ts
lilyydu: They are independent but it leaves it in a weird spot wtih partial persistence - I don't have a strong pref but tagging @corinagum here as she initiated the discussion

MehakBindra: also note provider failures may still cause partial writes, so successful scopes remaining clean and failed scopes dirty is a good way to handle this

corinagum: What does the dirty flag do for us on a failed scope? `save()` only runs in the turn's finally and `seal()` runs right after, so I don't see anything picking it up.

Maybe narrow things down: provider failures are transient and need a transaction. But for serialization, a circular ref or BigInt in user state will make `JSON.stringify` throw every turn, and conversation write already landed. That's not a one-off partial write, it's a permanently wedged app, and moving both `stringify` calls above the first storage call costs nothing and removes the whole class.

MehakBindra: But for serialization, a circular ref or BigInt in user state will make JSON.stringify throw every turn, and conversation write already landed -- > that's the intended behavior, consistent with .net

corinagum: I'd rather match PY than diverge on the one point we're debating. If TS should intentionally differ here, let's say why in the PR body so the next person doesn't have to rediscover it.

packages/apps/src/state/loader.ts
corinagum: `createStateLoader` does `fallbackStorage as IStorage<string, string>`, and `AppOptions.storage` is `IStorage`, i.e. `<any, any>`. So every app that already passes a storage provider gets silently retyped as string storage with nothing enforcing it. A provider that decodes JSON on read returns an object that `JSON.parse` would stringify to `object Object` and throw, and scope will silently reset on every turn. In PY, `_deserialize` accepts a `dict` directly. I think objects should be accepted like in PY, or drop the cast and make state storage an explicit, separately-typed thing so the contract is real.

packages/apps/src/app.process.ts
```diff
@@ -280,45 +288,97 @@ export class ActivityProcessor<TPlugin extends IPlugin = IPlugin> {
...
+
+    if (turnError) {
+      recordTeamsBotApplicationException(activityProcessSpan, turnError);
+      await this.options.eventManager.onError({
+        error: turnError instanceof Error
+          ? turnError
+          : new Error(String(turnError)),
+        activity,
+      });
+      return { status: 500 };
```
corinagum, line 372: This makes storage a hard availability dependency for every activity, in both directions. If `dispatch()` returned 200 and we already replied to the user, a transient `save` failure rewrites that turn into a 500 to the channel. And if `load()` throws at line 339, `dispatch()` never runs at all, so the handler is skipped and the user gets nothing back. I'd expect a persistence failure to surface through `onError` and telemetry without downgrading a turn that succeeded.

packages/apps/src/app.process.ts
```diff
@@ -280,45 +288,97 @@ export class ActivityProcessor<TPlugin extends IPlugin = IPlugin> {
...
         }
+      );
 
-        this.options.eventManager.onActivityResponse({
-          ...ref,
-          activity,
-          response: response,
-        });
-      }
+      this.options.eventManager.onActivityResponse({
```
corinagum, line 316: This does bring the runtime value in-line with the declared `IActivityResponseEvent.response` type, so I think it's the right direction, but it's a public plugin event, so it should be called out in the PR description and adding a test.

packages/apps/src/state/loader.ts
```diff
@@ -0,0 +1,199 @@
...
+      throw new Error('A conversation ID is required to save turn state.');
+    }
+
+    await this.saveScope(
+      this.conversationKey(conversationId),
+      container.conversation
+    );
+
+    if (container.user) {
+      if (!userId) {
```
corinagum: This throws after the conversation scope has already been written, so it's another partial-persistence path. And more generally, taking `conversationId`/`userId` as parameters means a caller can save a container under keys it wasn't loaded from. In PY, this is avoided by reading identity off the container: `TurnStateContainer` carries `conversation_id`/`user_id` and the docstring calls out that "a save can never be told to persist under a different key than it was loaded from." `TurnStateLoader` is exported public API here, so I think we should match that shape.

packages/apps/src/state/loader.ts
```diff
@@ -0,0 +1,199 @@
...
+  async delete(conversationId: string, userId?: string): Promise<void> {
+    await this.storage.delete(this.conversationKey(conversationId));
+    if (userId) {
+      await this.storage.delete(this.userKey(conversationId, userId));
+    }
+  }
+
+  private async loadScope(key: string): Promise<TurnState> {
+    const value = await this.storage.get(key);
+    if (value === undefined) {
```
corinagum, line 141: Providers commonly return `null` for a miss rather than `undefined`, and `IStorage.get` is typed `TValue | undefined` with `TValue = any`, so `null` passes. It then falls into `deserializeState`, fails `isStateRecord`, and logs "malformed persisted turn state" on every cold key. Maybe `value == null` so a plain miss stays quiet and the warning remains clear.

packages/apps/src/app.process.ts
MehakBindra: Hmm, if state is enabled, it IS part of the turn, consistent with PY and .NET. But worth thinking about a bit more

packages/apps/src/app.ts
```diff
@@ -167,9 +168,21 @@ export type AppOptions<TPlugin extends IPlugin> = {
...
+   * @deprecated Configure `state.storage` for turn state. Applications that
+   * need general persistence should own and use their storage provider directly.
    */
   readonly storage?: IStorage;
 
+  /**
+   * Enables per-turn conversation and user state.
+   *
+   * Pass `true` to use the app storage with default keys, or provide options
+   * to configure dedicated storage, key prefix, or expiration. State is
```
corinagum: `StateOptions` only has `storage` and `keyPrefix` now, no expiration.

examples/state/src/index.ts
```diff
@@ -0,0 +1,95 @@
...
+    },
+  },
+});
+
+const markdown = (text: string) =>
+  new MessageActivityInput(text).withTextFormat('markdown');
+
+app.on('message', async ({ activity, log, reply, state }) => {
+  if (!state) {
+    throw new Error('Turn state is not enabled.');
```
corinagum, line 18: Could we make `state` non-optional when it's enabled, or give the context a narrowing helper so handlers in a state-enabled app don't pay for needing to check state and `state.user` (e.g. line 51)?

packages/apps/src/state/loader.ts
MehakBindra: Not quite. Here storage is explicitly `IStorage<string, string>`, so `null` violates the contract; only `undefined` means a miss.

packages/apps/src/state/loader.ts
MehakBindra: Thsi is intentional, persistence identity is supplied to `SaveAsync` by the activity processor rather than stored on the handler-facing container. This keeps the container focused on state values and avoids exposing infrastructure identity through `ctx.state`. Since conversation and user scopes are independent, persisting conversation state before a user-scope validation failure is expected behavior.

packages/apps/src/state/loader.ts
MehakBindra: Good point !

MehakBindra: Removed the unsafe fall back cast

packages/apps/src/app.process.ts
MehakBindra: good callout, updated !

examples/state/src/index.ts
MehakBindra: Good point. The callback type cannot currently infer that state was enabled through runtime `AppOptions`, and user state must remain optional because some activities have no sender identity.

packages/apps/src/app.process.ts
MehakBindra: Finally doing this : When state is enabled, persistence is part of successful turn completion; returning success would falsely imply the state update was committed.
The tradeoff is possible retry/duplicate side effects, but that is preferable to silently acknowledging a turn whose required state was lost. Keep dispatch failure primary and report save failure separately when both occur.

lilyydu (approved):
LGTM but looks liek there's some merge conflicts!

packages/apps/src/state/loader.ts
lilyydu: I think there's two issues Corina's bringing up here:

1) `throw` after the write for conversation,- I don't see a harm in validating both conversationId and userId at the begininning of save
2) `save` should be locked to the keys it loaded
`delete` already gets this right — `load()` binds it to a closure, so `container.delete()` can only touch the keys it loaded. `save` doesn't; it takes the IDs as params, so `load(convA, userA)` then `save(container, convB, userB)` writes under a different key with nothing stopping it.

let's consider binding `save` like `delete` — capture identity at load time and call `container.save()` with no ID params. Handlers still don't see the IDs, and saving under the wrong key isn't possible

MehakBindra: Hmm, Idm changing , but then if we decide that persistence identity should be bound at load time, I think we should apply that consistently to both `save` and `delete` in TypeScript and Python.

lilyydu: yep in Python for save, we check the container for the fields

corinagum (approved):

lilyydu (approved):

PR #732, opened 2026-08-18
Resolve npm audit vulnerabilities

heyitsaamir:
## Summary

This clears the workspace npm audit report while keeping the repo on current dependency lines instead of taking `npm audit --force` downgrades. The security fix patches the BotBuilder line and constrains vulnerable transitive `uuid`/`esbuild` paths that upstream packages have not yet released with audit-safe ranges.

## Decisions

- **Avoided audit-force downgrades.** npm proposed semver-major downgrades for BotBuilder and Restify to satisfy transitive `uuid` advisories. Keeping current packages and overriding only the vulnerable transitive packages avoids regressing supported package and example versions.
- **Kept broader outdated upgrades out of scope.** `npm outdated` still reports non-security upgrades, including several majors; this PR limits movement to the install/audit fix path so reviewers can assess security risk separately from larger dependency migrations.

## Validation

- `npm audit --audit-level=low`
- `npm run check:lockfile-registry`
- `npx turbo build --filter=@microsoft/teams.botbuilder --filter=@examples/botbuilder`
- `npx turbo test --filter=@microsoft/teams.botbuilder`

### DIFF
--- a/examples/botbuilder/package.json
+++ b/examples/botbuilder/package.json
@@ -21,7 +21,7 @@
     "@microsoft/teams.apps": "*",
     "@microsoft/teams.botbuilder": "*",
     "@microsoft/teams.cards": "*",
-    "botbuilder": "4.23.1"
+    "botbuilder": "4.23.3"
   },
   "devDependencies": {
     "@microsoft/teams.config": "*",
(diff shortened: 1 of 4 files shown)


examples/botbuilder/package.json
```diff
@@ -21,7 +21,7 @@
     "@microsoft/teams.apps": "*",
     "@microsoft/teams.botbuilder": "*",
     "@microsoft/teams.cards": "*",
-    "botbuilder": "4.23.1"
+    "botbuilder": "4.23.3"
```
corinagum, line 24: Didn't we want to intentionally pin to 4.23.1? Or am I mis-remembering something?

PR #739, opened 2026-08-19
Fix targeted message updates by dropping recipient from wire payload (#738)

### LINKED ISSUE #738: [Bug]: Targeted message update via ctx.send sends recipient and fails
### Bug Description

Updating a targeted message through `ctx.send`/handler `send` with an existing activity id fails because the SDK includes `recipient` in the update request body. Teams rejects targeted-message updates that attempt to edit the recipient.

### Steps to Reproduce

1. In a channel or group chat that supports targeted messages, send a targeted message:
   ```ts
   const result = await send(
     new MessageActivityInput('initial')
       .withRecipient(activity.from, true)
   );
   ```
2. Attempt to update it through `ctx.send`/handler `send` by setting the original activity id and preserving the targeted recipient:
   ```ts
   await send(
     new MessageActivityInput('updated')
       .withId(result.id)
       .withRecipient(activity.from, true)
   );
   ```
3. Observe the update request fail.

### Expected Behavior

`ctx.send` should use `recipient.isTargeted === true` to route the update through `updateTargetedActivity`, but the outbound update body should not include `recipient`, because the recipient of an existing targeted message cannot be edited.

### Actual Behavior

The SDK routes to the targeted update endpoint, but sends `recipient` in the payload. Teams returns:

```text
400 Bad Request
BadArgument: Cannot edit Recipient of Targeted Message
ms-cv: 6xi2E/GC6k6hdZMSUVzMhA.0
```

### SDK Version

2.1.0

### Node.js Version

v23.7.0

### Additional Context

Confirmed manually in `examples/targeted-messages`: stripping `recipient` from the wire payload after using it to choose the targeted update endpoint makes the update succeed.

jibin7jose:
Summary:
- Route targeted `ctx.send` updates through `updateTargetedActivity` without sending `recipient`.
- Add a regression test that verifies the targeted update payload omits `recipient`.

Fixes #738

### DIFF
--- a/packages/apps/src/activity-sender.spec.ts
+++ b/packages/apps/src/activity-sender.spec.ts
@@ -134,7 +134,11 @@ describe('ActivitySender', () => {
       const conversations = (mockClient as any).conversations;
       expect(conversations.createTargetedActivity).toHaveBeenCalledWith(
         'conv-123',
-        expect.objectContaining({ type: 'message', text: 'targeted' })
+        expect.objectContaining({
+          type: 'message',
+          text: 'targeted',
+          recipient: expect.objectContaining({ isTargeted: true }),
+        })
       );
     });
 
@@ -150,7 +154,7 @@ describe('ActivitySender', () => {
         recipient: { id: 'user-1', name: 'User', role: 'user', isTargeted: true },
       } as ActivityParams;
 
-      await sender.send(activity, groupRef);
+      const result = await sender.send(activity, groupRef);
 
       const conversations = (mockClient as any).conversations;
       expect(conversations.updateTargetedActivity).toHaveBeenCalledWith(
@@ -164,6 +168,7 @@ describe('ActivitySender', () => {
       );
       expect(conversations.updateTargetedActivity.mock.calls[0][2]).not.toHaveProperty('recipient');
       expect(conversations.createActivity).not.toHaveBeenCalled();
+      expect(result.recipient).toEqual(expect.objectContaining({ isTargeted: true }));
     });
 
     it('should merge bot and conversation from ref into activity', async () => {
@@ -262,6 +267,7 @@ describe('ActivitySender', () => {
 
       const result = await sender.send(activity, groupRef);
       expect(result).toEqual(expect.objectContaining({ id: 'activity-1' }));
+      expect(result.recipient).toEqual(expect.objectContaining({ isTargeted: true }));
     });
   });
 
--- a/packages/apps/src/activity-sender.spec.ts
+++ b/packages/apps/src/activity-sender.spec.ts
@@ -156,8 +156,13 @@ describe('ActivitySender', () => {
       expect(conversations.updateTargetedActivity).toHaveBeenCalledWith(
         'conv-123',
         'existing-id',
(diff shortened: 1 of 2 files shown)


packages/apps/src/activity-sender.spec.ts
```diff
@@ -156,8 +156,13 @@ describe('ActivitySender', () => {
       expect(conversations.updateTargetedActivity).toHaveBeenCalledWith(
         'conv-123',
         'existing-id',
-        expect.objectContaining({ recipient: expect.objectContaining({ isTargeted: true }) })
+        expect.objectContaining({
```
corinagum, line 163: The `createTargetedActivity` assertion at lines `135-138` only checks `type` and `text`. If the strip ever moved above the create/update branch, targeted sends would break and that test would stay green. Would you mind adding `recipient: expect.objectContaining({ isTargeted: true })` there?

packages/apps/src/activity-sender.ts
```diff
@@ -87,4 +87,9 @@ export class ActivitySender implements IActivitySender {
     return { ...payload, ...res };
   }
 
+  private stripRecipient<T extends { recipient?: unknown }>(activity: T): Omit<T, 'recipient'> {
+    const { recipient: _recipient, ...updatePayload } = activity;
+    return updatePayload;
```
corinagum: `Omit<T, 'recipient'>` collapses `payload` into the intersection of keys. At runtime the object will still have `textFormat`, `attachments`, etc., but the type no longer does: `stripped.textFormat` fails. May be better to strip payload directly.


```suggestion
if (payload.id) {
  // `recipient` selects the targeted endpoint, but Teams rejects it in the update body.
  const { recipient: _recipient, ...targetedUpdate } = payload;
  const res = isTargeted
    ? await api.conversations.updateTargetedActivity(ref.conversation.id, payload.id, targetedUpdate)
    : await api.conversations.updateActivity(ref.conversation.id, payload.id, payload);
  return { ...payload, ...res };
}
```

packages/apps/src/activity-sender.spec.ts
```diff
@@ -156,8 +156,13 @@ describe('ActivitySender', () => {
...
+          type: 'message',
+          text: 'targeted update',
+          id: 'existing-id',
+        })
       );
+      expect(conversations.updateTargetedActivity.mock.calls[0][2]).not.toHaveProperty('recipient');
```
corinagum, line 169: Could you also assert that `SentActivity` still has `recipient`? 

`expect(result.recipient).toEqual(expect.objectContaining({ isTargeted: true }));`

corinagum (commented):
Hey @jibin7jose, thanks for the PR! I've tested this e2e to validate. I've pinged @ShanmathiMayuramKrithivasan and asked her to also take a look in case there are updates we need on the wire. 

teams.py has the same bug at `packages/apps/src/microsoft_teams/apps/activity_send.py:43`, which passes the whole activity to `update_targeted_activity`. teams.net is unaffected because `Context.SendAsync` has no update path. Would you be interested in helping out with the PY fix?

corinagum (approved):

PR #743, opened 2026-08-20
test(apps): add Socket Mode e2e matrix

singhk97:
## Summary

Adds a categorized App-level end-to-end matrix for inbound **Socket Mode**, restores make-before-break token rotation, and completes the deterministic Jest coverage for the Jest and Hybrid matrix criteria.

## Automated coverage

Exercises the real `App` / `SocketModeAdapter` / `GeoSocket` pipeline through contract-faithful controlled connections:

- core message, Adaptive Card, dialog, and reaction routing;
- file, authentication, message-extension, and meeting activity routing;
- exclusive socket transport and direct `App.onActivity(...)` dispatch;
- multi-geo readiness, startup retries, reconnect, restart, and five-second make-before-break handoff;
- concurrent invokes and mixed message, reaction, invoke, and task activity routing;
- non-draining in-flight shutdown behavior;
- outbound reply and proactive-send independence from the inbound transport;
- envelope casing, payload aliases, protocol versions, correlation, acknowledgements, errors, lifecycle events, and credential redaction.

Hybrid scenarios retain their focused live Teams-client validation boundary for behavior owned by Teams, APX routing, Azure SignalR, or service-generated activity shapes.

## Validation

- Socket Mode Jest suite: **9 suites passed, 117 tests passed**
- apps build ✅
- ESLint ✅


packages/apps/src/ws-connect/app.e2e.spec.ts
```diff
@@ -0,0 +1,338 @@
...
+  } = {
+    startCalls: 0,
+    stopCalls: 0,
+  };
+
+  const factory: SocketConnectionFactory = (context, handlers): ISocketConnection => {
+    captured.context = context;
+    captured.handlers = handlers;
+    return {
+      start: async () => {
```
corinagum: `start()` should stay pending until readiness is signalled. Right now it resolves either way, and `autoReady: false` only skips the `onReady` call.

packages/apps/src/ws-connect/app.e2e.spec.ts
```diff
@@ -0,0 +1,338 @@
...
+        if (options.startError) {
+          throw options.startError;
+        }
+        if (options.autoReady !== false) {
+          handlers.onReady({ botKey: 'bot1', connectionId: 'connection-1' });
+        }
+      },
+      stop: async () => {
+        captured.stopCalls++;
+      },
```
corinagum: `stop()` should call `handlers.onDisconnected()` before resolving, the way the real connection does.

packages/apps/src/socket-mode/app.e2e.spec.ts
```diff
@@ -0,0 +1,338 @@
...
+        protocolVersion: WS_CONNECT_PROTOCOL_VERSION,
+        envelopeId: 'failed-message',
+        status: 500,
+      });
+      expect(reply?.body).toBeUndefined();
+    });
+
+    it('forwards default and custom connection options to the connection factory', async () => {
+      const defaults = makeFakeFactory();
+      const defaultApp = createTestApp({
```
corinagum, line 308: This covers the defaults, but only through the options-object form. Nothing in the matrix constructs an app with `wsConnect: true`.

packages/apps/src/ws-connect/app.e2e.spec.ts
```diff
@@ -0,0 +1,338 @@
...
+      const { captured, factory } = makeFakeFactory({ autoReady: false });
+      const app = createTestApp({
+        clientId: 'bot1',
+        logger: quiet(),
+        wsConnect: { connectionFactory: factory, fallbackToHttp: false },
+      });
+      const ready = jest.fn();
+      const disconnected = jest.fn();
+      const reconnected = jest.fn();
+      app.wsConnect!.events.on('ready', ready);
```
corinagum: `EventEmitter.emit` in `teams.common` runs handlers in a plain loop with no try/catch, and `SocketServer` emits `ready` from inside the connection's `onReady` callback. Maybe add a case where one of the listeners throws?

corinagum (commented):
Are you planning on adding integration tests? maybe scoped to .net? Please verify if TS and PY need any language-specific additions.

packages/apps/src/ws-connect/app.e2e.spec.ts
singhk97: Fixed. `start()` now stays pending until readiness is signalled — the fake blocks on an internal promise resolved by `fireReady()`, so `autoReady: false` genuinely models "connected but not yet ready" and the lifecycle test drives readiness explicitly.

packages/apps/src/ws-connect/app.e2e.spec.ts
```diff
@@ -0,0 +1,338 @@
...
+        }
+        if (options.autoReady !== false) {
+          handlers.onReady({ botKey: 'bot1', connectionId: 'connection-1' });
+        }
+      },
+      stop: async () => {
+        captured.stopCalls++;
+      },
+    };
+  };
```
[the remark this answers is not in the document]
singhk97: Fixed. The fake now models the real readiness contract: `start()` stays pending when `autoReady` is false and only resolves once `SocketReady` is fired (via the new `fireReady()` helper), matching the real `ISocketConnection` which doesn't resolve `start()` until readiness.

packages/apps/src/socket-mode/app.e2e.spec.ts
singhk97: Added. There's now a dedicated case constructing an app with the boolean `socketMode: true` shorthand, asserting it runs the composite transport, exposes `app.socketMode`, and forwards the default connection options (negotiate URL, readiness timeout, keep-alive/server-timeout) to the connection.

packages/apps/src/ws-connect/app.e2e.spec.ts
singhk97: Addressed with the new connection contract. The real connection now reports termination via `onClosed` (not `onDisconnected`), and the fake models a drop through `drop()` → `handlers.onClosed()`, which is what drives the supervisor's disconnect→reconnect path in the lifecycle test. The server translates that into the public `disconnected` event.

packages/apps/src/ws-connect/app.e2e.spec.ts
singhk97: Added, and the underlying issue is fixed on #716. There's now a case where a `ready` listener throws: `app.start()` still resolves and status reaches `ready`. On the impl side, `SignalRSocketConnection` completes the internal readiness transition before invoking observers and wraps `onReady` in a try/catch, and `SocketServer.safeEmit` isolates event-listener throws — so a throwing listener can't wedge readiness even though `teams.common` `EventEmitter.emit` has no try/catch.

lilyydu (dismissed):

puneetgulia-rubberducky (dismissed):
Requesting changes for the following validated issues:

1. **Startup failure is tested at the wrong layer and the documented behavior is contradictory.** The test says `App.start()` fails and tears down, but calls `CompositeServer.start()` directly. `App.start()` catches the rejection, calls `stop()` (closing the already-started HTTP fallback), reports the error, and resolves. The linked manual matrix separately says a socket failure is non-fatal and HTTP continues serving. Please define one contract and test it through `App.start()`: either rethrow after cleanup, or explicitly assert the error event and stopped state.

2. **`ackRequired` is ignored.** The envelope contract says it controls whether a one-way delivery ACK is expected, but `SocketServer` returns an ACK for every non-invoke activity. Add a false/missing case and either suppress the reply or clarify/change the contract.

3. **Closed/stale generations are not synchronously fenced.** `onClosed` only settles the supervisor promise; `readyGen` is cleared later. An activity queued in that interval can still pass `canDispatch`, and generation callbacks do not verify that they belong to the current generation. Invalidate the generation synchronously and add a test dispatching through the dropped connection before reconnect completes.

4. **The lifecycle stop assertion is a false positive.** Reconnect already stops the old generation, so `stopCalls >= 1` is true before `app.stop()`. The test would pass if final shutdown leaked the active replacement connection. Track per-connection stop counts and assert both generations separately.

5. **The fake cannot model stop during readiness.** Its `start()` ignores `AbortSignal`; with `autoReady = false`, neither abort nor `stop()` settles the pending readiness promise. This contradicts `ISocketConnection` and leaves the critical start/stop race untested.

6. **Readiness observability and multi-geo lifecycle claims are not pinned.** `ready` is emitted before that geo's status changes to `ready`, so listeners observe `connecting`. The lifecycle case is single-geo, while the three-geo case verifies only URLs/count—not geo-tagged disconnect/reconnect independence.

7. **Positive fixtures are not contract-valid.** Account roles and `conversationType` are omitted, the Adaptive Card action lacks required `type`/`data`, and broad casts hide these defects. Use typed fixtures with `satisfies`; reserve unsafe casts for explicit malformed-input tests.

Existing readiness, boolean-shorthand, and throwing-listener comments appear addressed. The earlier `stop()` comment is outdated after the `onClosed` contract change, but final-shutdown coverage remains insufficient as noted above.

packages/apps/src/socket-mode/socket-mode-adapter.resilience.spec.ts
```diff
@@ -270,18 +299,161 @@ describe('SocketModeAdapter resilience', () => {
...
+        expect(connState.connections).toHaveLength(3);
+        expect(server.status).toBe('ready');
+
+        await server.stop();
+      } finally {
+        jest.useRealTimers();
+      }
+    });
+
+    it('reports an outage if the old socket dies before its replacement is ready', async () => {
```
lilyydu, line 429: nit: can we move the make before break tests to `app.e2e.transport-recovery.spec.ts`

lilyydu (approved):

PR #744, opened 2026-08-21
feat(apps): add multi-flow OAuth lifecycle

MehakBindra:
## Summary

Adds a first-class, per-connection OAuth lifecycle to `@microsoft/teams.apps`, modeled on `microsoft/teams.net` and using the turn-state support from #729.

- adds `OAuthFlow` for silent token lookup, interactive sign-in, sign-out, and completion/failure callbacks
- supports `AppOptions.oauthFlows`, `app.addOAuthFlow(...)`, and `app.getOAuthFlow(...)`
- routes token exchange, verify-state, and sign-in failure invokes across multiple connections
- persists pending attribution and exchange deduplication in turn state, with bounded process-local fallback when state is disabled
- adds `ctx.getConnectionStatus()`, including direct token checks that correct stale silent-SSO status results
- handles channel/group-chat OAuth cards consistently with Teams SSO behavior
- replaces the Graph-only example with a Graph and GitHub OAuth example

## Public API

```ts
const app = new App({ oauthFlows: ['graph', 'github'] });
const graph = app.getOAuthFlow('graph');

graph
  .onSignInComplete(async (ctx, token) => { /* use token */ })
  .onSignInFailure(async (ctx, failure) => { /* handle failure */ });

const token = await graph.signIn(ctx);
await graph.signOut(ctx);
const statuses = await ctx.getConnectionStatus();
```

## Compatibility and behavior

- retains existing OAuth settings, context helpers, events, routes, response shapes, explicit legacy connection overrides, and the TypeScript sign-in activity override
- registered flows do not enable eager per-turn token lookup
- omitted state is enabled internally for registered flows; explicit `state: false` keeps bounded process-local tracking
- expected token misses (`400`, `404`, `412`) trigger interactive fallback; unexpected service or transport failures propagate
- verify-state continues only for connection misses; unexpected HTTP failures preserve their status
- successful exchanges are marked processed before callbacks, so retries remain deduplicated if a callback throws

## Breaking changes

- `ctx.signin()` previously treated every token lookup failure as a missing token and sent an OAuth card. It now sends a card only for expected token-miss responses (`400`, `404`, `412`). Unexpected service and transport failures propagate and may produce a `500` for the turn instead of silently switching to interactive sign-in.
- OAuth invoke spans are unified under `microsoft.teams.oauth`, with `oauth.operation` distinguishing token exchange, verification, and failure. Existing `oauth.result` values also move to lifecycle-specific values. Dashboards and alerts using the previous span names or result values must migrate; this must be called out in the release changelog. The non-HTTP error value remains `exception`.


packages/apps/src/oauth/registry.ts
```diff
@@ -0,0 +1,152 @@
...
+   * Validates a connection before a deprecated context OAuth helper runs.
+   */
+  validate(connectionName: string, connectionNameProvided: boolean): void {
+    if (this.usesRegisteredFlows && !connectionNameProvided) {
+      throw new Error(
+        'OAuth connection name is required when OAuth flows are registered.'
+      );
+    }
+
+    this.get(connectionName);
```
lilyydu, line 170: we should consider checking registry only if `usesRegisteredFlows` is true. Previously,
`ctx.signin({ connectionName: "github" })` and `ctx.signout("github")` forwarded explicit connection names even when the app default was different. adding this would mean these existing calls would throw & would be a breaking change

packages/apps/src/oauth/handlers.ts
```diff
@@ -0,0 +1,442 @@
...
+    try {
+      token = await ctx.api.users.getToken({
+        channelId: ctx.activity.channelId,
+        userId: ctx.activity.from.id,
+        connectionName: flow.connectionName,
+        code: state,
+      });
+    } catch (error) {
+      if (error instanceof AxiosError) {
+        this.clearPending(ctx, flow);
```
lilyydu: should these be deferred until routing is resolved? in `verifyState`, 400/404/412 normally means “this code does not belong to this channel; try the next flow,” not that the flow failed

if graph returns 404 and github subsequently succeeds, graph's failure callback currently fires and its pending attribution is cleared. I think probe misses should remain silent

packages/apps/src/app.ts
```diff
@@ -348,18 +144,37 @@ export class App<TPlugin extends IPlugin = IPlugin> {
...
+    if (
+      hasConfiguredOAuthFlows &&
+      this.options.oauth?.defaultConnectionName !== undefined
+    ) {
+      throw new Error(
+        'oauth.defaultConnectionName cannot be combined with registered OAuth flows. ' +
+        'Remove defaultConnectionName and name the connection when calling OAuth helpers.'
+      );
+    }
+    if (hasConfiguredOAuthFlows && this.options.state === false) {
```
lilyydu: hmm I don't think we need to require turn state for all flows, for Python we allow a process-local oauth fallback when state is disabled

packages/apps/src/oauth/handlers.ts
```diff
@@ -0,0 +1,442 @@
...
+          log.warn(
+            `auth state not found for conversation "${activity.conversation.id}" and user "${activity.from.id}"`
+          );
+          telemetry.result = APP_OAUTH_RESULT.noToken;
+          telemetry.responseStatus = 404;
+          return { status: 404 };
+        }
+
+        for (const flow of flows) {
+          const response = await this.verifyFlow(ctx, flow, activity.value.state, span, telemetry);
```
lilyydu: token service can return a 5xx here but it get swallowed as a 404

packages/apps/src/oauth/index.ts
```diff
@@ -0,0 +1,491 @@
...
+    return (await this.getToken(context)) !== undefined;
+  }
+
+  /**
+   * Gets the token service status for the current user's configured connections.
+   *
+   * The Bot Framework endpoint returns all connection statuses; callers can
+   * select this flow's entry by matching {@link connectionName}.
+   */
+  getConnectionStatus(context: IActivityContext): Promise<TokenStatus[]> {
```
lilyydu: should this filter to only return for a specific connection name, or rename to pluralize?

packages/apps/src/oauth/handlers.ts
```diff
@@ -0,0 +1,442 @@
...
+  onSignInFailure = async (
+    ctx: contexts.IActivityContext<ISignInFailureInvokeActivity, PluginAdditionalContext<TPlugin>>
+  ) => {
+    const { log, activity, next } = ctx;
+    const failure: SignInFailure = activity.value ?? {};
+    const { code, message } = failure;
+    const pendingFlows = this.getOrderedFlows(ctx, true).filter(
+      flow => flow.getPending(ctx, true) !== undefined
+    );
+    const flows = pendingFlows.length > 0 ? [pendingFlows[0]] : this.getFlows();
```
corinagum, line 129: if TTL is expired or state is unavailable, this falls back to every registered flow. With the example's handlers both calling `ctx.send`, a user who leaves a GitHub card sitting sees both "Microsoft Graph sign-in failed." and "GitHub sign-in failed.". Should the no-pending case notify nothing, or only fall back when a single flow is registered? The tests only cover the has-pending path.

packages/apps/src/oauth/handlers.ts
```diff
@@ -0,0 +1,442 @@
...
+ * Previously a set of `this: App`-bound free functions; now a collaborator the
+ * {@link App} constructs and registers as system routes. The handlers are arrow
+ * class fields so they can be passed directly as route callbacks.
+ */
+export class OauthHandlers<TPlugin extends IPlugin = IPlugin> {
+  private static readonly STATE_TTL_MS = 5 * 60 * 1000;
+  private static readonly MAX_EXCHANGE_ENTRIES = 1_000;
+  private static readonly EXCHANGE_STATE_KEY = '__oauth:exchanges';
+
+  private readonly tokenExchangeLocks = new Map<string, Promise<{ status: number, body?: TokenExchangeInvokeResponse }>>();
```
corinagum: `tokenExchangeLocks` is keyed on the bare `value.id`, but the durable dedupe in `markExchangeProcessed` is conversation-scoped. Should key both the same way. Maybe also add a comment that this map is process-local, so concurrent duplicates only collapse on a single instance.

packages/apps/src/oauth/index.ts
```diff
@@ -0,0 +1,491 @@
...
+
+  try {
+    const response = await context.api.users.getToken({
+      channelId: context.activity.channelId,
+      userId: context.activity.from.id,
+      connectionName,
+    });
+    return response.token;
+  } catch (error) {
+    if (!isMissingTokenError(error)) {
```
corinagum: This changes `ctx.signin()` behavior. Previously everything fell through to sending a card; now anything outside 400/404/412 propagates, so a transient token service blip turns a sign-in card into a 500. I think the new behavior is correct, but the description lists `ctx.signin()` under retained compatibility and the semantics note only mentions `OAuthFlow.signIn()`. Can we call this out as a breaking change?

packages/apps/src/diagnostics/constants.ts
```diff
@@ -67,17 +65,26 @@ export const APP_OAUTH_OPERATION = {
...
-  notified: 'notified',
+  cached: 'token_cached',
+  cardSent: 'signin_card_sent',
+  hit: 'token_found',
+  miss: 'token_not_found',
+  success: 'operation_succeeded',
+  failure: 'operation_failed',
+  duplicate: 'request_deduplicated',
+  noToken: 'connection_not_matched',
+  notified: 'failure_callback_notified',
```
corinagum, line 90: Every value here changes (`success` to `operation_succeeded`, `no_token` to `connection_not_matched`, and so on), and the three oauth span names collapse into one. These are `@internal`, but anything already dashboarding or alerting on them breaks with no compile error to catch it. Should be noted in PR summary & changelog at release time. 

Separately, `exception: 'invalid_op'` on line 96 seems odd for "a non-HTTP exception was thrown".

packages/apps/src/app.ts
```diff
@@ -985,4 +839,14 @@ export class App<TPlugin extends IPlugin = IPlugin> {
...
     return this.options.oauth?.defaultConnectionName !== undefined;
   }
+
+  private enableOAuthState(): void {
+    if (this.options.state === false) {
+      throw new Error(
+        'OAuth flows require turn state. Remove state: false or configure state options.'
+      );
+    }
+    this.stateLoader ??= createStateLoader(true, this.storage, this.log);
```
corinagum: Registering a flow silently enables state using the app's default storage, which is `LocalStorage` unless configured. OAuth correctness now depends on that state for pending attribution and exchange dedupe, so across more than one instance sign-in can simply never complete. The existing `LocalStorage` warning is generic. Can it name the OAuth consequence when state is enabled from here?

packages/apps/src/oauth/registry.ts
MehakBindra: yup, can do

packages/apps/src/oauth/handlers.ts
MehakBindra: the exchange ID contract defines it as unique per token-exchange request, while conversation state merely provides the persistence namespace. This map is process-local, so it only collapses concurrent duplicates that reach the same app instance, conversation state handles persisted retries across instances

packages/apps/src/oauth/index.ts
MehakBindra: Yupp

packages/apps/src/oauth/index.ts
MehakBindra: haha yeah this one is weird. I'm just moving it to ctx

packages/apps/src/app.ts
MehakBindra: yes agreed, I don't know why I did this

packages/apps/src/oauth/handlers.ts
MehakBindra: yes this is a bug, fixing , thanks !

MehakBindra: If the code doesn't belong to the flow, I don't think we get an error code, we just get no token back. verified this behavior via testing

packages/apps/src/oauth/handlers.ts
MehakBindra: Hmm, this needs to be fixed in .net too

packages/apps/src/contexts/activity.ts
```diff
@@ -221,12 +212,16 @@ export interface IBaseActivityContext<T extends Activity = Activity, TExtraCtx e
   /**
    * trigger user signin flow for the activity sender
    * @param options options for the signin flow
+   * @deprecated Register a connection with `app.addOAuthFlow(...)` and call
+   * `flow.signIn(ctx, options)` instead.
    */
-  signin: (options?: Partial<SignInOptions>) => Promise<string | undefined>;
+  signin: (options?: OAuthSignInOptions) => Promise<string | undefined>;
```
lilyydu, line 227: for signin and signout, bc these now call `validateOAuthConnection`, it throws for legacy apps if the connection name isn't the configured default. before, any name was passed through. Legacy apps using more than one connection through these helpers will now throw

MehakBindra: No , not really. `validate()` returns immediately while the app is in legacy mode, so arbitrary connection overrides still pass through for both helpers.

It becomes strict only after using `oauthFlows` or `addOAuthFlow()`. Tests explicitly cover preserving legacy non-default `signin` and `signout` connections.

packages/apps/src/oauth/handlers.ts
MehakBindra: For now, we’ll have to notify all registered flows when pending attribution is unavailable. Unfortunately, the `signin/failure` payload does not identify the connection, and preserving the failure notification is preferable to silently dropping it. This is an intentional tradeoff; callbacks may produce duplicate user-facing messages in this fallback case.

README.md
```diff
@@ -100,7 +100,7 @@ $: npm run test
 - [`@examples/echo`](./examples/echo/README.md)
 - [`@examples/auth`](./examples/auth/README.md)
 - [`@examples/botbuilder`](./examples/botbuilder/README.md)
-- [`@examples/graph`](./examples/graph/README.md)
+- [`@examples/oauth`](./examples/oauth/README.md)
```
corinagum: I think we need to keep graph because of templates, right?

packages/apps/src/oauth/handlers.ts
MehakBindra: aligning for 400/404/412

README.md
lilyydu: I ended up just updating the oauth sample in PR #5 to showcase multi-oauth, I left the graph sample alone

corinagum (approved):

examples/oauth/src/index.ts
```diff
@@ -0,0 +1,105 @@
...
+  logger: new ConsoleLogger('@tests/auth', { level: 'debug' }),
+  oauthFlows: ['graph', 'github'],
+  // This is an example of overriding the token URL for a specific region (e.g., Europe).
+  // Uncomment this block if needed.
+  // apiClientSettings: {
+  //   oauthUrl: 'https://europe.token.botframework.com',
+  // }
+});
+const graph = app.getOAuthFlow('graph');
+const github = app.getOAuthFlow('github');
```
lilyydu, line 18: for PY, I did two graph calls instead - one for profile/email, cause github is more difficult to configure/test

packages/apps/src/oauth.ts
```diff
@@ -1,23 +0,0 @@
-export type OAuthSettings = {
```
lilyydu, line 1: should this file be marked deprecated instead?

examples/oauth/src/index.ts
MehakBindra: I think it's okay for examples to differ. and this aligns wtih .net , 2 different connections is good to show

packages/apps/src/oauth.ts
MehakBindra: the individual methods/fields are marked deprecated where needed

lilyydu (approved):

PR #761, opened 2026-08-25
Remove internal naming from comments and docs

corinagum:
Clean up comments and docs


packages/api/src/activities/activity.ts
```diff
@@ -761,7 +761,7 @@ export class Activity<T extends string = string> implements IActivity<T> {
    * Add a targeted message info entity for prompt preview.
    * Skips if already present. In reactive flows, `ctx.send()` and `ctx.reply()`
    * populate this automatically — use this helper for proactive or deferred sends.
-   * An invalid or expired messageId causes APX to silently drop the preview
+   * An invalid or expired messageId causes the service to silently drop the preview
```
heyitsaamir: NIT: let's use "Teams Service" instead because this code is running in a service too

lilyydu (approved):

PR #779, opened 2026-08-28
refactor(dev): replace uuid with native randomUUID

MehakBindra:
## Summary

- replace `uuid.v4()` calls in `@microsoft/teams.dev` with Node's built-in `crypto.randomUUID()`
- remove the `uuid` dependency instead of adopting its ESM-only v14 release
- preserve CommonJS and Jest compatibility while keeping UUID v4 behavior

Supersedes #777, whose UUID 14 update fails because the release is ESM-only.

## Validation

- `npm test --workspace @microsoft/teams.dev -- --coverage=false --runInBand`
- `npm run build --workspace @microsoft/teams.dev`
- `npm run lint --workspace @microsoft/teams.dev`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20679,7 +20679,6 @@
         "@microsoft/teams.common": "*",
         "express": "^5.0.0",
         "jsonwebtoken": "^9.0.2",
-        "uuid": "^11.0.5",
         "ws": "^8.18.1"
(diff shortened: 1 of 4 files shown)


MehakBindra:
Closing in favor of the original #777 as requested.

PR #784, opened 2026-08-31
fix(apps): source informative textFormat from the update + add textFormat to update()

singhk97:
## Summary

Follow-up to #762. Two fixes for `textFormat` on streamed **informative updates**:

1. **Source informative `textFormat` from the update itself.** In `HttpStream.flush()`, informative typing chunks previously took their format from `this.finalActivity?.textFormat`. But `finalActivity` is only set once a *message* is drained from the queue — which happens *after* informative updates are collected — so an informative update's own `textFormat` was ignored. It now reads `informativeUpdate.textFormat` directly. The streamed-text chunk keeps its correct last-emitted-message-wins behavior.

2. **`update()` can now carry a `textFormat`.** `IStreamer.update(text)` / `HttpStream.update(text)` gain an optional, nullable `textFormat`:
   ```ts
   update(text: string, textFormat?: TextFormat | null): void;
   ```
   `undefined`/`null` → Teams default (`markdown`); an explicit value (e.g. `extendedmarkdown`) → that format. Previously the only way to format an informative update was to hand-build a `{ type: 'typing', ..., textFormat }` object and call `emit()`.

## Sample

`examples/stream` demonstrates the new `update()` overload and cleans up the `extended-markdown` scenario's deltas to match the existing sample style while showcasing checkboxes and ~~strikethrough~~.

## Tests

Added to `http-stream.spec.ts`:
- informative update carries its own `textFormat`, independent of `finalActivity` (red/green — fails against the old `finalActivity`-sourced logic)
- `update(text, 'extendedmarkdown')` sends an informative chunk with that format
- `update(text)` and `update(text, null)` omit `textFormat`

All apps tests pass (648/648), build + eslint clean.

## Cross-language

Counterpart PRs (not yet merged) will fold in the equivalent changes:
- Python microsoft/teams.py#581
- .NET microsoft/teams.net#661 (design differs — coordinated separately)


packages/apps/src/types/streamer.ts
```diff
@@ -97,8 +98,10 @@ export interface IStreamer {
   /**
    * send status updates before emitting (ex. "Thinking...")
    * @param text the status text to send
+   * @param textFormat format of `text` (ex. `'extendedmarkdown'`); omit or pass `null`
+   * to use the Teams default (`'markdown'`)
    */
-  update(text: string): void;
```
singhk97, line 101: the other option is to introduce another overload to `update` that strictly enforces textFormat. What's ur preference @corinagum ?

lilyydu (approved):

PR #789, opened 2026-09-01
Route threaded replies through reply endpoint

MehakBindra:
## Summary
- route threaded activities through the Bot Framework reply endpoint while keeping root sends on the create endpoint
- expose separate regular and targeted reply APIs; targeted threaded replies use `POST /v3/conversations/{conversationId}/activities/{threadRootId}?isTargetedActivity=true`
- preserve proactive callers by translating valid legacy `;messageid=` conversation IDs to base IDs plus endpoint placement
- apply reactive placement defaults for personal chats, group chats, and channels using typed `channelData.thread.id` metadata with legacy fallback
- infer targeted threaded delivery from outbound recipient metadata for proactive and reactive sends
- keep quoting independent from placement and targeted delivery, including explicit quotes on targeted threaded replies
- deprecate auto-quoting context helpers and threaded-ID construction, and update the threading and targeted-message samples

## Compatibility
Existing `App.reply` signatures and valid legacy threaded IDs continue to work. Legacy suffixes are translated internally and are not sent to APX. Existing targeted create/update/delete behavior is unchanged; targeted replies now retain their thread placement and explicit quote metadata.

## Validation
Focused API and apps suites were invoked, but the local dependency installation is incomplete: `npx` cannot resolve the repository Jest/TypeScript toolchain and fails before test execution on the `ts-jest` preset. `git diff --check` passes, and the complete targeted-thread diff received a static correctness review with no findings.


packages/apps/README.md
```diff
@@ -49,6 +49,12 @@ By default, the app receives Teams activities at `/api/messages`.
 For local-only testing without Teams service token validation, set `dangerouslyAllowUnauthenticatedRequests: true` or
 `DANGEROUSLY_ALLOW_UNAUTHENTICATED_REQUESTS=true`.
 
+## Reactive and proactive sends
+
+Use `ActivityContext.send(activity)` while handling an inbound activity to send to its current conversation. To send to
+a different conversation, use proactive `App.send()`. The legacy `ActivityContext.send(activity, conversationRef)`
+overload remains runtime-compatible but is deprecated.
```
corinagum, line 56:
```suggestion
is deprecated.
```

packages/apps/src/app.ts
```diff
@@ -556,7 +565,34 @@ export class App<TPlugin extends IPlugin = IPlugin> {
...
-      return this.send(toThreadedConversationId(conversationId, messageId), activity, options);
+      if (!this.id) {
+        throw new Error('App has no credentials set up');
+      }
+      if (!conversationId) {
+        throw new Error('conversationId must be a non-empty string');
+      }
+      if (!messageId || !/^\d+$/.test(messageId) || messageId === '0') {
+        throw new Error(
+          `Invalid messageId "${messageId}": must be a non-zero numeric value`
```
corinagum, line 576: Nit: maybe this is a tiny bit misleading, activity id's are always a string, but with number values, right?

```suggestion
          `Invalid messageId "${messageId}": must be a string type of non-zero numeric value`
```
Just to be explicit when viewing the error out of context.

examples/interacting-with-messages/src/quoting.ts
```diff
@@ -59,25 +59,6 @@ export async function handleQuoteMessage(
...
-  context: MessageContext,
-  text: string
-): Promise<boolean> {
-  if (text !== 'quote add') {
-    return false;
-  }
-
-  const sent = await context.send('Please review the latest PR before end of day.');
-  const message = new MessageActivityInput()
-    .addQuote(sent.id, 'Done! Left my comments on the PR.');
```
corinagum, line 76: Is `addQuote` still being showcased? This is removing both instances (`handleQuoteManual`).

corinagum:
~@MehakBindra (All three langs): aspects that include channel data will need preview markers until this is fully GA.~

PR #798, opened 2026-09-03
chore(deps-dev): bump jest to 30.5.1 and @types/jest to 30.0.0

lilyydu:
Bumps jest from 29.7.0 to 30.5.1 and @types/jest from 29.5.x to 30.0.0 across all workspaces. ts-jest stays at 29.4.12, which declares `jest: ^29.0.0 || ^30.0.0` as a peer and is compatible with Jest 30.


heyitsaamir (approved):
Stamp

PR #810, opened 2026-09-09
fix(apps): handle Socket Mode auth errors

teddyam:
Socket Mode negotiate failures currently use generic messages and retry every response, which makes invalid credentials difficult to diagnose and causes authorization failures to retry indefinitely.

This change carries the negotiate HTTP status on `NegotiateError` and provides actionable guidance for authentication and authorization failures. HTTP 401 remains retryable so corrected or refreshed credentials can recover, while HTTP 403 is treated as terminal during startup and reconnect because retrying cannot change the bot's authorization.

The package README and changelog document the behavior. Unit coverage verifies the 401 and 403 messages, startup retry decisions, terminal reconnect behavior, and error logging.

## Testing

- `npm test --workspace @microsoft/teams.apps -- --runInBand src/socket-mode/negotiate.spec.ts src/socket-mode/socket-server.resilience.spec.ts`
- `npm run build --workspace @microsoft/teams.apps`
- `npm run lint --workspace @microsoft/teams.apps -- --quiet`


packages/apps/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # @microsoft/teams.apps
 
+## Unreleased
```
lilyydu: don't need this part!

packages/apps/README.md
```diff
@@ -49,6 +49,18 @@ By default, the app receives Teams activities at `/api/messages`.
 For local-only testing without Teams service token validation, set `dangerouslyAllowUnauthenticatedRequests: true` or
 `DANGEROUSLY_ALLOW_UNAUTHENTICATED_REQUESTS=true`.
 
+## Socket Mode connection errors
+
+Socket Mode reports actionable errors when its negotiate request is rejected:
```
lilyydu: can also remove from README - we'll add these details eventually to official docs

packages/apps/src/socket-mode/socket-server.ts
```diff
@@ -672,6 +677,13 @@ class GeoSocket {
       } catch (err: any) {
         lastError = err;
         if (!this.server.accepting) break;
+        if (isTerminalConnectionError(err)) {
```
lilyydu, line 680: is it possible for this error to not be a `NegotiateError`? wondering if its enough to just check for 403

packages/apps/src/socket-mode/negotiate.ts
```diff
@@ -48,17 +48,42 @@ export function assertSecureNegotiateUrl(negotiateUrl: string): void {
...
     super(message);
     this.name = 'NegotiateError';
   }
 }
 
+/** Build an actionable error message for a failed negotiate request. */
+function negotiateErrorMessage(status: number, body: string): string {
+  switch (status) {
+    case 401:
+      return (
```
lilyydu, line 74: for 401/403, I think we should keep the original status/body with the message.

e.g., still keep "verify the bot credentials and restart the app" 

 (in case there's backend changes in the future, itll still get propagated up)

packages/apps/src/socket-mode/socket-server.ts
```diff
@@ -672,6 +677,13 @@ class GeoSocket {
...
+          this.log.error(
+            `socket-mode[${this.geo}]: initial connection failed with a non-retryable error`,
+            err
+          );
+          break;
+        }
```
lilyydu, line 686: do we also need to do:

`this.server.emit('disconnected', { geo: this.geo, error });` here and maybe in L870?

packages/apps/src/socket-mode/socket-server.ts
teddyam: connectCycle() initiates the SignalR handshake which could return a plain `Error`. We could simplify that line by only checking `error.statusCode == 403` and it'd still work but I think we should prob check the obj type

packages/apps/src/socket-mode/socket-server.ts
teddyam: I feel like this would be good to test but Copilot says no need in both cases.
- **Initial startup 403:** No `disconnected` event should fire because the geo was never connected. `start()` rejects with the `NegotiateError`, which is the correct signal.

- **Reconnect 403 after an unexpected drop:** `disconnected` was already emitted when the original socket dropped, before reconnect began. Emitting it again would duplicate the event.

teddyam: Actually there is an edge case where we should fire `disconnected` at L870 -- when the reconnect originated from planned token rotation. Let me know if the logic there makes sense cause it needed a small tweak

packages/apps/src/socket-mode/negotiate.ts
teddyam: as in we don't touch it at all or we keep the original body and append details? i'm assuming don't touch at all

lilyydu: keep original body and append details

packages/apps/src/socket-mode/socket-server.ts
lilyydu: Yeah I think we need to do the emit for the refresh case

lilyydu (approved):

corinagum (approved):

lilyydu (approved):

PR #815, opened 2026-09-10
fix: remove sovereign escape hatch & and implement make-before-break socket reconnects

lilyydu:
**GOAL**: improve the resilience and correctness of the socket handoff process during proactive token refreshes in Socket Mode, and tighten cloud environment restrictions. 

- when a socket connection is replaced (e.g., due to token rotation), the old connection is kept alive for a short period to handle backend cache overlap, preventing dropped messages
- sovereign cloud support is now explicitly disallowed regardless of endpoint overrides

**Socket handoff and connection lifecycle improvements:**
* Introduced a `CONNECTION_HANDOFF_MS` constant (5 seconds) to keep the old socket connection alive after its replacement is ready, ensuring continued delivery while backend caches expire (`socket-server.ts`).
* Refactored `GeoSocket` to track both the active and retiring connections, allowing both to receive messages during the handoff window, and updated connection management and cleanup logic accordingly 

**Cloud environment restrictions:**
* Changed logic to reject Socket Mode in sovereign clouds regardless of whether an explicit `negotiateBaseUrl` is provided, and updated related tests to expect rejection in all sovereign scenarios


packages/apps/src/socket-mode/socket-server.ts
```diff
@@ -38,6 +38,11 @@ const RECONNECT_MAX_DELAY_MS = 15_000;
 
 /** Renegotiate this long before the negotiate token is due to expire. */
 const TOKEN_REFRESH_MARGIN_MS = 60_000;
+/**
+ * Backend caches connection IDs for roughly five seconds. Keep the superseded
+ * socket alive for that window after its replacement reaches `SocketReady`.
+ */
+const CONNECTION_HANDOFF_MS = 5_000;
```
teddyam, line 45: is there a reason for 5 seconds here

lilyydu: This is APX's restriction - they have 5s cache

packages/apps/src/socket-mode/composite-adapter.ts
```diff
@@ -12,8 +12,10 @@ import { HttpMethod, HttpRouteHandler, IHttpServerAdapter, IHttpServerInitialize
  * which transport delivers each activity, so the SDK just needs to receive on
  * both. There's no fallback logic or health state — route registrations and
  * static-file mounts fan out to every child, and each delivers what it receives
- * into the shared {@link HttpServer} pipeline. The Teams backend service delivers
- * each activity over exactly one transport, so there is no double-processing.
+ * into the shared {@link HttpServer} pipeline. Initial delivery uses one
+ * transport, but the Teams backend service may retry over HTTP after an
+ * ambiguous socket failure, so application handlers can observe the same
```
teddyam, line 17: good catch on the retry verbage

packages/apps/src/socket-mode/socket-server.ts
```diff
@@ -705,14 +706,15 @@ class GeoSocket {
...
-  /** Drain-agnostic close of this geo: abort refresh, close the connection. */
+  /** Close every connection generation still owned by this geo. */
   async stop(): Promise<void> {
     this.clearRefreshTimer();
-    const connection = this.connection;
-    this.connection = undefined;
-    if (connection) {
-      await connection.stop().catch(() => undefined);
-    }
+    const connections = [this.active?.connection, ...this.retiring.values()].filter(
```
teddyam, line 712: if shutdown occurs during an in-progress replacement handshake, could there be a generation inside `connectCycle()` that isn't accounted for in this snapshot?

teddyam (approved):

packages/apps/src/socket-mode/socket-server.ts
```diff
@@ -845,8 +872,11 @@ class GeoSocket {
   }
 
   private async reconnect(
-    prevError: Error | undefined,
-    outageReported: boolean
+    prevError?: Error,
+    options: { keepServing: boolean; outageReported: boolean } = {
```
corinagum, line 876: Do we want to drop the default value, since `reconnect()` has one call site and it always passes both options explicitly? Default is currently unreachable.

packages/apps/src/socket-mode/socket-connection.ts
```diff
@@ -115,14 +115,21 @@ export class SignalRSocketConnection implements ISocketConnection {
...
+    // until it reaches readiness.
+    const onAbort = () => {
+      if (connected) {
+        failReady(new Error('Socket Mode connect aborted'));
+      }
+      void this.stop().catch((err) => {
+        this.log?.warn('socket-mode: failed to stop an aborted connection', err);
+      });
+    };
     signal?.addEventListener('abort', onAbort, { once: true });
```
corinagum, line 129: A failed SignalR handshake will still have the listener attached until `finally`, and could accumulate.

examples/socket/README.md
```diff
@@ -39,8 +39,9 @@ You should see per-geo `ready` logs as each connection comes up, then `you said
 
 ## Notes
 
-- **Single-transport delivery** — with the HTTP fallback enabled, each activity
-  is delivered over exactly one transport (socket or HTTP, never both), so no
-  dedupe is required.
+- **HTTP fallback is opt-in** — `fallbackToHttp: true` is permitted only for
+  production bots and is generally not recommended. It starts a local HTTP
+  receiver, but you must separately enable and configure the matching public
+  messaging endpoint in Teams Developer Portal (TDP).
```
teddyam, line 45: the endpoint can be configured for Azure bots in Azure too right

lilyydu: technically yes, but I believe all config regarding socket mode is preferably to be set in TDP

teddyam (approved):

teddyam (approved):

teddyam (approved):

PR #817, opened 2026-09-10
feat(apps): resolve Agentic User file bytes through Graph /shares

corinagum:
### What this does

Agentic User file attachments arrive with a `contentUrl` and **no** `downloadUrl`, so the pre-authorized ("tempauth") URL traditional bots rely on is absent. This resolves those bytes through Microsoft Graph, `GET /shares/{u!encoded}/driveItem/content`, using the Agentic User's own credential.

The route is chosen from the **shape of the attachment**, not from the identity: a `contentUrl` with no `downloadUrl`, at personal scope, resolves via `graphShare`. Everything else keeps the existing `preauthUrl` path unchanged.

### What it deliberately does not do

**An expired pre-authorized URL is terminal.** It raises `FileUrlExpiredError` and makes no Graph call.

A developer can still call Graph with their own client, and keeps the locator because `contentUrl` is public API. The constraint is on what the SDK does automatically.

Auditing trap: the agentic token is itself delegated-shaped, carrying `scp` rather than `roles`. Follow the identity, not the word "delegated".

### Why `.default` and not a named scope

A by-name request hard-fails on exact-string consent mismatch: a blueprint consented to `Files.ReadWrite.All` returns `AADSTS65001` and **no token at all** for a `Files.Read.All` request. Graph's own `/shares` docs list `Files.ReadWrite.All` as least-privileged and list no read-only permission, so admins following Microsoft's guidance land exactly where a by-name request breaks.

The cost is a late, ambiguous 403 when a file permission is missing. We pay that in the SDK instead: the pre-flight rejects a token carrying permissions but **none file-capable**, which restores an early, legible failure. A `Files.Read.All`-then-`.default` ladder was rejected, because `AADSTS65001` is also what "no file permission consented at all" produces, so the ladder makes the most common deployment error harder to diagnose, not easier.

**For security review:** under `.default` the token carries the blueprint's full consented set, including unrelated scopes. The mitigation is the grant, not the request string: do not consent unrelated scopes on a file-handling blueprint.

### Testing

Unit and integration coverage is in the diff. All three scenarios were also run **live against real Teams**, tenant asserted twice per run (browser session state, and the SharePoint host in the bot's own log).

| Scenario | Result |
| --- | --- |
| Traditional bot, fresh file | `preauthUrl`, byte-exact |
| Agentic User via Graph | `/shares/{u!}/driveItem/content` 200, byte-exact, driving the shipped helpers |
| Expired URL | `FileUrlExpiredError`, **zero requests to `graph.microsoft.com` in the entire process log** |

The third row was checked by grepping the whole bot process log for an *absence*, not by reading the rendered result, and the expiry is real rather than simulated: the same `UniqueId` returns 200 early and a hard 401 later. And the pre-authorized URL lifetime is **not a constant**: three runs bounded the flip at (62.7, 72.4], (62.5, 68.4] and (63.6, 66.4] minutes. Poll rather than picking a wait time.

---

## This language

22 commits, 28 files, +2004 / -121.

**One commit is not about files.** `Resolve the activity tenant from channel data as well as the conversation, and stop passing hardcoded 'common'` fixes an auth bug this work exposed. TypeScript builds an authority URL, so with no tenant it substituted `'common'`, which is wrong for a single-tenant bot.

**Python has the same fix; .NET correctly does not.** TS and PY *construct* an authority URL and so need a placeholder when they have no tenant. .NET *configures* one through named MSAL options and leaves `Tenant` unset, so there is no placeholder to remove.

`Test that the Graph bearer is not forwarded across a redirect` pins that the bearer does not reach a storage host across a 302. The comparison is origin-based and `.host` includes the port.


packages/apps/src/files/files-accessor.ts
```diff
@@ -95,13 +98,26 @@ export class FilesAccessor implements IFilesAccessor {
 
     const content = asFileDownloadInfo(attachment.content);
     const downloadUrl = content?.downloadUrl;
+    // Narrowed for the same reason `content` is shape-checked above: the wire is untrusted, and a truthy non-string here would pass the locator test and then throw inside the sharing-url encoder rather than being skipped.
+    const contentUrl = typeof attachment.contentUrl === 'string' ? attachment.contentUrl : undefined;
```
teddyam, line 103: `asFileDownloadInfo()` can return undefined if `content` is malformed in addition to when `downloadUrl` doesn't exist (expected for Agentic User). Could this would trigger the Graph flow for cases where we don't want it?

packages/apps/src/files/download.ts
```diff
@@ -145,6 +360,31 @@ async function requestViaHttpClient(
...
+        return undefined;
+      }
+
+      try {
+        const chunks: Buffer[] = [];
+        let size = 0;
+
+        for await (const chunk of body) {
+          chunks.push(chunk as Buffer);
+          size += (chunk as Buffer).length;
```
teddyam: here the chunk is appended before checking the limit, so a chunk could theoretically exceed the bound significantly. The fetch adapter also buffers the complete body by `response.text()`. Should we keep only the remaining bytes up until we hit the `ERROR_BODY_LIMIT`?

teddyam (approved):

PR #840, opened 2026-09-16
refactor(ci): read issue event data directly from the event payload

corinagum:
The `Issue Analysis → Teams` workflow had a `Resolve issue details` step that copied fields off the issue event into `$GITHUB_OUTPUT` so the two steps that actually use them could read them back as `steps.issue.outputs.*`.

That indirection isn't needed. Both consuming steps can read `github.event.issue.*` directly in their own `env:` block, so this drops the intermediate step and wires the values straight through.

## Changes

- Remove the `Resolve issue details` step.
- `Analyze issue with Copilot CLI` takes `ISSUE_NUMBER` from `github.event.issue.number` in `env:` rather than assigning it inside the script.
- `Triage and notify Teams` reads its six `ISSUE_*` values from `github.event.issue.*`.
- Label joining moves from a `jq` call to the `join()` expression.

## Notes

The analysis script's interface is unchanged. It still reads `ISSUE_NUMBER`, `ISSUE_TITLE`, `ISSUE_AUTHOR`, `ISSUE_HTML_URL`, `ISSUE_LABELS` and `ISSUE_BODY` from the environment and still splits labels on commas, and `join(github.event.issue.labels.*.name, ',')` produces the same comma-joined string `jq` did. Verified that the workflow still parses and that no `steps.issue` references remain.


lilyydu (approved):

teddyam (approved):

corinagum:
FYI work is showing as merged to `main` though the queue stalled: https://github.com/microsoft/teams.ts/commit/317175ce8019f300df29164c0eb327bb906460c7
Will close this PR as merged if it doesn't resolve.

PR #847, opened 2026-09-18
Guard npm `latest` dist-tag to the current stable release branch

lilyydu:
`2.1.0` is GA, but `.azdo/publish.yml` picked the npm tag from the version string alone — any stable build got `latest`. Since npm dist-tags are functional, a `2.0.17` hotfix would have silently downgraded every consumer off 2.1.0. Latent only; live tags are healthy (`latest=2.1.0`).

- Gate `latest` on a new `currentStableBranch` variable; a stable build from any other branch now fails with a message naming the variable to bump.
- `RELEASE.md`: one stable line at a time, `release/v2.0` closed at `2.0.16`, fixes ship forward (matches microsoft/teams.py#599).

Separately, 11 packages carry a stray `$(npmtag)` → `2.0.6-preview.1` dist-tag from an old run — all except `teams.apps`, `teams.devtools`, and `teams.m365extensions`. The bad reference is already gone from the repo, so this is cosmetic registry cruft needing a manual `npm dist-tag rm`; it does not affect `latest` or `next` resolution.


corinagum (approved):

corinagum:
nice catch!

PR #848, opened 2026-09-18
feat(apps): option to strip only the bot's own leading mention

### LINKED ISSUE #833: Provide a built-in option to strip only the bot’s own mention from inbound activity text
The current `activity.mentions.stripText: true` option removes all structured mentions from inbound activity text. Bot command handlers commonly need different behavior: remove the bot’s leading mention while preserving mentions of other users.

For example:

```text
Input:  <at>GitHub</at> help <at>Alice</at>
Output: help <at>Alice</at>
```

Today this requires application code to identify the mention whose account ID matches `activity.recipient.id` and remove it manually:

```ts
for (const entity of activity.entities ?? []) {
  if (
    entity.type === 'mention' &&
    entity.mentioned?.id === activity.recipient.id &&
    entity.text
  ) {
    activity.text = activity.text.replace(entity.text, '').trim();
    break;
  }
}
```

The lower-level `stripMentionsText` utility supports filtering by `accountId`, but applications must know to call it and pass the recipient ID for every activity. The app-level mention middleware could provide this common behavior directly because it already has access to the activity and its recipient.

Would you consider an option such as:

```ts
const app = new App({
  activity: {
    mentions: {
      stripText: {
        recipient: true,
        leadingOnly: true,
      },
    },
  },
});
```

or a dedicated utility:

```ts
stripRecipientMentionText(activity, { leadingOnly: true });
```

Expected behavior:

- Remove the bot/recipient mention from the beginning of the message.
- Preserve mentions of other users.
- Preserve a bot mention appearing later in the message when `leadingOnly` is enabled.
- Handle mention entities that provide either `text` or only the mentioned account’s name.
- Trim whitespace left by removing the leading mention.

This would cover the common Teams bot command-parsing scenario without requiring each application to implement its own mention matching and text mutation.

Montana:
Hi all, 

This will close #833.

In essence, `activity.mentions.stripText: true` currently removes **every** structured mention from inbound text. That's wrong for the most common bot scenario, command parsing, where the bot needs to drop *its own* leading `@mention` and keep any other mentions as meaningful arguments:

```
Input:  <at>GitHub</at> help <at>Alice</at>
Before: help                       // Alice is lost
After:  help <at>Alice</at>        // with { recipient: true, leadingOnly: true }
```

This PR adds two composable options:

| Option | Layer | Meaning |
|---|---|---|
| `leadingOnly?: boolean` | `@microsoft/teams.api` → `StripMentionsTextOptions` | Remove matching mentions only while they appear at the start of the text. |
| `recipient?: boolean` | `@microsoft/teams.apps` → new `AppStripMentionsTextOptions` | Resolve `accountId` from each inbound activity's `recipient.id` at middleware time. |

Usage:

```ts
const app = new App({
  activity: {
    mentions: {
      stripText: { recipient: true, leadingOnly: true },
    },
  },
});
```

## Motivation

Without this, every app has to write something like the following, usually with subtle bugs around name-only mention entities and whitespace:

```ts
for (const entity of activity.entities ?? []) {
  if (entity.type === 'mention' && entity.mentioned?.id === activity.recipient.id && entity.text) {
    activity.text = activity.text.replace(entity.text, '').trim();
    break;
  }
}
```

The lower-level `stripMentionsText` utility already supports `accountId`. But the app-level option has no way to express "the recipient of *this* activity," because the bot's ID is only known per activity. The middleware already has the activity in scope, so it's the natural place to bind it.

### `leadingOnly` (api)

1. Candidate mentions are the `mention` entities, filtered by `accountId` when one is given.
2. Each candidate resolves to a tag: `mention.text` if present, otherwise `<at>${mention.mentioned.name}</at>`. This is the same fallback the existing code uses.
3. The text is `trimStart()`-ed, then, in a loop, if it starts with any candidate tag, that tag is consumed and the remainder is `trimStart()`-ed again.
4. The loop stops at the first non-matching token. Nothing after that point is touched.
5. With `tagOnly: true`, a consumed tag is replaced by its inner name, and the original whitespace after it is preserved.
6. The result is `trim()`-ed, which is consistent with the existing return contract.

Consequences:

| Input | Options | Output |
|---|---|---|
| `<at>Bot</at> help <at>Alice</at>` | `{ accountId: bot, leadingOnly }` | `help <at>Alice</at>` |
| `<at>Bot</at> tell <at>Bot</at> hi` | `{ accountId: bot, leadingOnly }` | `tell <at>Bot</at> hi` |
| `hey <at>Bot</at> help` | `{ accountId: bot, leadingOnly }` | `hey <at>Bot</at> help` (unchanged) |
| `<at>Alice</at> <at>Bot</at> help` | `{ accountId: bot, leadingOnly }` | unchanged: a non-matching leading mention blocks stripping |
| `<at>Bot</at> <at>Alice</at> help <at>Alice</at>` | `{ leadingOnly }` | `help <at>Alice</at>`: consecutive leading mentions of any account |
| `<at>Bot</at> help <at>Alice</at>` | `{ accountId: bot, leadingOnly, tagOnly }` | `Bot help <at>Alice</at>` |
| `<at>Bot</at>` | `{ accountId: bot, leadingOnly }` | `''` |

### `recipient` (apps middleware)

- Applies to the same activity types as before: `message`, `messageUpdate`, `typing`.
- For each activity, `accountId` is set to `activity.recipient.id`. When `recipient` is set, it **takes precedence** over a static `accountId`, which is documented on the type.
- If `recipient: true` but the activity has no `recipient.id`, stripping is **skipped** and the middleware calls `next()`. The alternative, falling through to "strip all mentions," would silently remove user-authored content.

## Backward compatibility

- Both options are optional and default to off. `stripText: true` and existing `StripMentionsTextOptions` objects behave exactly as before.
- The non-`leadingOnly` path was refactored to share the tag-resolution helpers, but its behavior is unchanged. It still replaces the first occurrence of each mention's tag, and every pre-existing test passes unmodified.
- The public type surface is additive only. `StripMentionsTextOptions` gains one optional field. `AppStripMentionsTextOptions` is new and exported from `@microsoft/teams.apps`. I gave it a distinct name to avoid shadowing the api package's `StripMentionsTextOptions` for consumers who import from both.

## Design notes and alternatives considered

- **Standalone `stripRecipientMentionText(activity, opts)` utility** (the issue's second proposal). This is still possible and cheap to add on top. I didn't make it the primary API because it leaves every app to remember to call it per activity, which is the problem the issue describes. It can be added in a follow-up if maintainers want both.
- **Stripping only a single leading mention** instead of consecutive ones. I rejected this because `<at>Bot</at> <at>Bot</at> …` would leave a dangling mention, and without `accountId`, "leading" most naturally means the whole leading run.
- **Regex-based matching.** I avoided it because mention text is user-controlled display text, so exact prefix matching on resolved tags avoids escaping pitfalls.
- **No integration test.** Per CONTRIBUTING, this is inbound text normalization and doesn't change anything on the wire.

## Tests

**`packages/api/src/activities/utils/strip-mentions-text.spec.ts`**: 9 new cases under `leadingOnly`:
- strips the leading recipient mention and preserves others
- preserves a later recipient mention
- does not strip a non-leading recipient mention
- does not strip when another account's mention leads
- handles leading whitespace
- falls back to `mentioned.name` when `text` is absent
- strips consecutive leading mentions without `accountId`
- keeps the name with `tagOnly`
- returns `''` when the text is only the mention

**`packages/apps/src/middleware/strip-mentions-text.spec.ts`** (new; the middleware had no tests before): 6 cases:
- default strips all mentions
- `recipient` strips only the recipient's mentions
- `recipient` + `leadingOnly` preserves later bot mentions
- the recipient is resolved per activity, not captured once
- `recipient` overrides `accountId`
- non-text activities pass through untouched and `next()` is always called

Results: the full `@microsoft/teams.api` suite passes (25 suites, 255 tests), the new apps middleware suite passes (6/6), and ESLint is clean on all touched files.

Cheers, 
Michael


Montana:
> @microsoft-github-policy-service agree
> @microsoft-github-policy-service agree company="self"

@microsoft-github-policy-service agree company="self"

